PR #6, opened 2017-10-16
feat: Implement full minidump support

jan-auer:
This PR intends to implement full minidump support for Linux and MacOS,
so that it can be tested on iOS and Android. While previously most of
the functionality resided in the breakpad crate, it is now moved directly
into symbolic. This especially allows us to use `ByteView` and `Arch`.

Note that the first implementation will require the users to generate
Breakpad symbol files instead of just uploading Dwarf debug infos. At
a later stage, this can be replaced by an improved symbol file generation.

 - [x] Remove reference to `breakpad` crate
 - [x] Add submodules for breakpad and lss
 - [x] Move source code from breakpad over
 - [x] Refactor to use ByteView and Arch
 - [x] Move or write tests for processing
 - [x] Export processor to python
 - [x] Export CFI cache generation to python


jan-auer:
@mitsuhiko This should be ready now. 

Binary caches (should reduce the CFI cache sizes and make loading faster) and building with the `gcc` crate will follow in future PRs.

Also, there are still two potential bugs that I'll have to investigate soon-ish:
 - Linux has an unusually high number of frames in the crashed thread, will have to symbolicate and look at them
 - The CFI cache generator yields slightly different outputs from breakpad, most notably empty rows. While this doesn't break stackwalking, it yields a warning and should be prevented in future.

PR #14, opened 2017-12-05
fix: Correct demangling for ambiguous C++/Rust symbols

jan-auer:
The heuristic to detect `_ZN...E` as Rust symbols generates false positives for some symbols encountered in Electron stack traces, e.g.:

```
_ZN4base8internal7InvokerINS0_9BindStateIMN4mate19TrackableObjectBaseEFvvEJNS_7WeakPtrIS4_EEEEEFvvEE7RunImplIRKS6_RKNSt3__15tupleIJS8_EEEJLm0EEEEvOT_OT0_NS_13IndexSequenceIJXspT1_EEEE
```

Demangled:

```
void base::internal::Invoker<base::internal::BindState<void (mate::TrackableObjectBase::*)(), base::WeakPtr<mate::TrackableObjectBase> >, void ()>::RunImpl<void (mate::TrackableObjectBase::* const&)(), std::__1::tuple<base::WeakPtr<mate::TrackableObjectBase> > const&, 0ul>(void (mate::TrackableObjectBase::* const&&&)(), std::__1::tuple<base::WeakPtr<mate::TrackableObjectBase> > const&&&, base::IndexSequence<0ul>)
```

Also, the `try_demangle_cpp` splits incorrectly at argument lists of the template parameter:

```
"void base::internal::Invoker<base::internal::BindState<void "
```

### DIFF
--- a/demangle/src/lib.rs
+++ b/demangle/src/lib.rs
@@ -174,10 +174,8 @@ impl<'a> Symbol<'a> {
             return Some(Language::ObjC);
         }
 
-        // rust
-        if (self.mangled.starts_with("_ZN") || self.mangled.starts_with("__ZN")) &&
-            self.mangled.ends_with("E")
-        {
+        // rust?
+        if rustc_demangle::try_demangle(self.mangled).is_ok() {
             return Some(Language::Rust);
         }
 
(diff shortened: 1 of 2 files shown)


jan-auer:
Let us please merge this, as it is certainly more correct than the previous code to auto-detect the language. False positives should actually be quite rare, as Rust symbols always have trailing numbers that C++ doesn't have. So if a symbol is a parseable Rust symbol, I'm pretty certain it cannot be a C++ symbol...

jan-auer:
Replacing the C++ demangler and adding the language flag should be separate

PR #15, opened 2017-12-06
feat(symcache): Use language from debug symbols to demangle

jan-auer:
This uses the language obtained from debug symbols to determine which demangler
should be used. If the language cannot be determined, the same heuristics are
employed as before.


jan-auer:
👍  Updated & rebased to `master`

PR #18, opened 2017-12-12
feat(minidump): Implement symcache support for Breakpad symbols

### LINKED ISSUE #10: Support Breakpad symbols
Currently, only dwarf data can be processed by the stackwalker and symbolicator. However, our customers **might** wish to Breakpad symbols instead, as they have them already. In that case we should consider supporting those as well.

jan-auer:
This PR adds support to creating symcaches from Breakpad ASCII symbol files. Documentation on the format can be found [here](https://github.com/google/breakpad/blob/master/docs/symbol_files.md#public-records). Generation should work out-of-the-box with the high-level methods used before. 

**The way it works:**

It adds a new kind of `Object` file and implements respective methods in the SymCache. Python can just interface with the same functions as before and it will just work.

**Changes:**
 - Add architectures (`Arch`) for PowerPC and PowerPC64
 - Allow to convert Breakpad architecture identifiers to `Arch`
 - Introduce a concept of `DebugKind` for object files. This determines the strategy to extract debugging information regardless of the container. Can either be `Dwarf` or `Breakpad` currently. ELF and Mach-O are currently hard-coded to return `Dwarf`.
 - Add a new `Object`/`FatObject` subtype `BreakpadObject`
 - Separate DWARF-reading from symcache generation and move it to a `DwarfInfo` (i.e. file).
 - Implement a primitive breakpad symbol file parser in `BreakpadInfo`
 - Create an abstraction over `DwarfInfo` and `BreakpadInfo` with `DebugInfo` and use that when generating symcaches.

**NOTE**

This PR also bumps `scroll` to version 0.8 to fix a version mismatch with the one used in `goblin` 0.0.13. Also it removes the pinned GitHub version of `goblin` from `symbolic-cabi`. 

Closes #10


py/tests/test_symcache.py
```diff
@@ -18,6 +18,54 @@ def test_basic(res_path):
...
+    cache = SymCache.from_bytes(stream.read())
+
+    # Verify a known symbol
+    symbol = cache.lookup(0x107bb9f25 - 0x107bb9000)[0]
+    assert symbol.symbol == 'main'
+    assert symbol.lang == 'unknown'
+    assert symbol.line == 186
+    # Comprises "comp_dir" and "base_dir"
+    assert symbol.filename == '/Users/electron/workspace/electron-osx-x64/out/R/../../atom/app/atom_main.cc'
+    # "lang", "comp_dir" and "base_dir" are not available in .sym files
```
jan-auer: This could be an interesting detail. Breakpad symbols do not contain the compilation directory. If we strictly need it, some sort of heuristics would be necessary to "guess" it.

PR #30, opened 2017-12-21
ci: Silence output from docker pull

jan-auer:
Hopefully fixes flaky travis builds


jan-auer:
nvm, it's the combination of docker, make and pip.

PR #36, opened 2018-01-03
fix(debuginfo): Parse INFO records in breakpad symbols

jan-auer:
Adds support to parse `INFO` records in Breakpad ASCII symbols. Also, breakpad to symcache conversion silently swallowed errors.

### DIFF
--- a/symcache/src/breakpad.rs
+++ b/symcache/src/breakpad.rs
@@ -37,8 +37,8 @@ impl<'input> BreakpadInfo<'input> {
 
     fn parse(&mut self, object: &'input Object) -> Result<()> {
         let mut records = object.breakpad_records();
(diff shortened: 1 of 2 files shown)


symcache/src/breakpad.rs
```diff
@@ -37,8 +37,8 @@ impl<'input> BreakpadInfo<'input> {
 
     fn parse(&mut self, object: &'input Object) -> Result<()> {
         let mut records = object.breakpad_records();
-        while let Some(Ok(record)) = records.next() {
-            match record {
+        while let Some(record) = records.next() {
```
jan-auer: Having `if let Some(Ok(foo)) = ...` should probably be a red flag in future code reviews.

PR #39, opened 2018-01-09
ref(demangle): Improve C++ demangling

jan-auer:
Switches to `cpp_demangle` in favor of `__cxa_demangle`  and adds some more C++ tests


jan-auer:
@mitsuhiko updated as discussed

PR #42, opened 2018-01-10
ci: Fix random Travis failures related to blocking stdout

jan-auer:
Besides, we're not really interested in `yum` output anyway


jan-auer:
Seems to work  👀

PR #46, opened 2018-01-11
fix(minidump): Use original return addresses

jan-auer:
This PR tries to expose the original return addresses from stack frames without Breakpad's instruction heuristics. It allows us to apply our own heuristics defined in `symcache/heuristics` from Sentry. On the bottom line, Sentry no longer has to distinguish between minidump stacktraces and KSCrash stacktraces to perform correct symbolication.

### Getting the correct return address

The first issue was to obtain the actual return address from the process state. Breakpad defines the following interface for stack frames:

```
struct StackFrame {
  // Return the actual return address, as saved on the stack or in a
  // register. See the comments for 'instruction', below, for details.
  virtual uint64_t ReturnAddress() const { return instruction; }

  // The program counter location as an absolute virtual address.
  // - [Except for the first frame], this address is within the instruction that
  //   caused execution to branch to this frame's callee [...]
  uint64_t instruction;
}
```

This suggests that Breakpad tries to infer the caller address and saves it in `instruction` while returning the original return address with `ReturnAddress`. However, this default implementation simply returns `instruction` which already looks like a source for errors. In fact, it is not overridden for `ARM` and `ARM64`: The `instruction` field is offset by `-2` and `-4` respectively, so `ReturnAddress()` does not return the original value. 

For this reason, our implementation of `StackFrame::return_address` in Rust compensates this difference. 

### Instruction heuristics

The heuristics inside `symcache/src/heuristics.rs` have been refactored to be more clear about what they do and why. Semantics should not have changed. Notable changes are:

 - `Arch::instruction_alignment` now centralizes all alignment-related operations
 - All logic whether a frame's instruction should be adjusted now lives in `should_adjust_caller`
 - Public helper methods `aligned_address` and `previous_address` use the above to navigate instructions
 - Renamed `find_best_instruction` to `caller_address` (only in Rust) to be more specific about what the method actually does
 - Added detailed comments about the reasoning behind the manipulations, as this is a fragile yet important piece of code.

Based on [this document](https://opensource.plausible.coop/wiki/display/PLCR/Automated+Crash+Report+Analysis), however, I am still not quite clear about two parts. Also, we do not seem to apply the [two Link Register Heuristics](https://opensource.plausible.coop/wiki/display/PLCR/Automated+Crash+Report+Analysis#AutomatedCrashReportAnalysis-FixingtheIssue.1) for ARM.

### Final remarks

 - Sentry must must use `return_address` over `instruction` to get frames' addresses when building events from minidumps. This can be implemented after the next release of `symbolic`.
 - With CFI (which was never present on KSCrash), some of these heuristics should probably be turned off. Sentry is not using the `FrameTrust` field yet, but it should probably pass it to the heuristics. This might require a change to the frame interface in Sentry.
 - We do not yet have any test dumps to verify these heuristics or the quality of stack traces
 - We still need a way to read register values from the process state and add them to events

### DIFF
--- a/symcache/src/heuristics.rs
+++ b/symcache/src/heuristics.rs
@@ -1,4 +1,4 @@
-use symbolic_common::{Arch, CpuFamily};
+use symbolic_common::Arch;
 
 const SIGILL: u32 = 4;
 const SIGBUS: u32 = 10;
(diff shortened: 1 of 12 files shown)


symcache/src/heuristics.rs
```diff
@@ -4,65 +4,113 @@ const SIGILL: u32 = 4;
...
+            return true;
+        }
+
+        // The crashing frame usually contains the actual register contents,
+        // which points to the exact instruction that crashed and must not be
+        // adjusted. A notable exception to this is if we crashed with one of
+        // the crash signals.
+        // TODO: Document reason of this
+        if let Some(ip) = self.ip_reg {
             if ip != self.addr && self.is_crash_signal() {
```
jan-auer: This is the first condition, I'm not entirely clear about.

symcache/src/heuristics.rs
```diff
@@ -4,65 +4,113 @@ const SIGILL: u32 = 4;
...
+        } else {
+            self.aligned_address()
+        };
+
+        // For some crashes on ARM, the aligned address resolved to the wrong
+        // symbol.
+        // TODO: Document details for this
+        match self.arch.cpu_family() {
+            CpuFamily::Arm32 => addr + 1,
+            CpuFamily::Arm64 => addr + 3,
```
jan-auer: This is the second one. Do we have an example where this shows effect? Judging by how our symbol lookup works, this **should** make no difference, so I suspect a bug in a different place.

jan-auer:
@mitsuhiko this is ready now

PR #47, opened 2018-01-15
fix(debuginfo): Compute correct UUID for ELF binaries

jan-auer:
This PR fixes some issues with ELF processing and refactors some format-specific code into its own helper modules.

### Build ID Algorithm

This algorithm computes a UUID for the ELF binary. The previous way to use `e_ident` in the ELF header is simply wrong, as `e_ident` is no identifier by any means but rather a collection of magic and flags. For obvious reasons, we try to closely resemble what Breakpad is doing:

1. Determine the build identifier:
    1. First, try to locate a `NT_GNU_BUILD_ID` in one of the `PT_NOTE` program header entries
    2. Otherwise, locate the above note in the contents of the `".note.gnu.build-id"` section
    3. If the above does not work, take the entire `".text"` section (the program code) and XOR it into a 16 byte buffer
2. Strip or extend the buffer obtained above to exactly 16 bytes
3. Interpret the above as 4-field UUID: `(u32, u16, u16, [u8; 8])`
4. If the target architecture is little endian, flip the first three fields into big endian

### Load Address

We previously assumed the load address is always `0x0`, which is completely wrong for executables. In that case, the load address is the start address of the first `PT_LOAD` segment.

Note: We refer to the load address as `vmaddr` (virtual memory address) internally, as we prefer Mach speak.

### Refactoring

To structure the above code better, all ELF-related utility functions have been moved to an internal `elf` module which exports functions like `find_elf_section` or `get_elf_uuid`. In the same way, all Mach-O related helpers have been moved to the `mach` module.

The `find_*_section` now always return a structure with reference to the section header and the data. For now, I've decided not to expose these functions publicly, but they are for sure candidates for the public API. There might be a follow-up PR with further refactoring of the code structure here.

### Dependencies

This PR makes use of `goblin::note::NoteIterator` to parse the contents of note sections. The easiest way to create an instance was to expose the internal `Elf.ctx` field for now. However, I don't really like this solution and might change this before creating a PR with `goblin`.

### DIFF
--- a/debuginfo/src/elf.rs
+++ b/debuginfo/src/elf.rs
@@ -3,6 +3,8 @@ use std::cmp;
 use goblin::elf;
 use uuid::Uuid;
 
+use symbolic_common::Result;
+
 const UUID_SIZE: usize = 16;
 const PAGE_SIZE: usize = 4096;
 
@@ -182,4 +184,20 @@ pub fn get_elf_uuid(elf: &elf::Elf, data: &[u8]) -> Option<Uuid> {
     }
 
     None
+}
+
+/// Gets the virtual memory address of this object's .text (code) section
+pub fn get_elf_vmaddr(elf: &elf::Elf) -> Result<u64> {
+    // For non-PIC executables (e_type == ET_EXEC), the load address is
+    // the start address of the first PT_LOAD segment.  (ELF requires
+    // the segments to be sorted by load address.)  For PIC executables
+    // and dynamic libraries (e_type == ET_DYN), this address will
+    // normally be zero.
+    for phdr in &elf.program_headers {
+        if phdr.p_type == elf::program_header::PT_LOAD {
+            return Ok(phdr.p_vaddr);
+        }
+    }
+
+    Ok(0)
 }
--- a/debuginfo/src/elf.rs
+++ b/debuginfo/src/elf.rs
@@ -1,0 +1,185 @@
+use std::cmp;
+
+use goblin::elf;
+use uuid::Uuid;
+
+const UUID_SIZE: usize = 16;
+const PAGE_SIZE: usize = 4096;
+
+/// A section inside an ELF binary
+pub struct ElfSection<'elf, 'data> {
+    pub header: &'elf elf::SectionHeader,
+    pub data: &'data [u8],
+}
+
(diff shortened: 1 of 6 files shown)


debuginfo/src/elf.rs
```diff
@@ -0,0 +1,185 @@
...
+/// As opposed to Mach-O, ELF does not specify a unique ID for object files in
+/// its header. Compilers and linkers usually add either `SHT_NOTE` sections or
+/// `PT_NOTE` program header elements for this purpose.
+///
+/// If neither of the above are present, this function will hash the first page
+/// of the `.text` section (program code) to synthesize a unique ID. This is
+/// likely not a valid UUID since was generated off a hash value.
+///
+/// If all of the above fails, the UUID will be `None`.
+pub fn get_elf_uuid(elf: &elf::Elf, data: &[u8]) -> Option<Uuid> {
```
jan-auer: This is the entrypoint to calculating an ELF uuid. See `find_build_id` for the logic of resolving the original build ID and `create_elf_uuid` for the endianness normalization.

debuginfo/src/elf.rs
```diff
@@ -0,0 +1,203 @@
...
+/// Gets the virtual memory address of this object's .text (code) section
+pub fn get_elf_vmaddr(elf: &elf::Elf) -> Result<u64> {
+    // For non-PIC executables (e_type == ET_EXEC), the load address is
+    // the start address of the first PT_LOAD segment.  (ELF requires
+    // the segments to be sorted by load address.)  For PIC executables
+    // and dynamic libraries (e_type == ET_DYN), this address will
+    // normally be zero.
+    for phdr in &elf.program_headers {
+        if phdr.p_type == elf::program_header::PT_LOAD {
+            return Ok(phdr.p_vaddr);
```
jan-auer: This is the new `vmaddr` logic.

jan-auer:
Updated to new interface as proposed in m4b/goblin#71

jan-auer:
This should be ready and stable now. Still not sure how we can provide tests for this reasonably, though. We could run some blackbox tests with the symbols in `py/tests/res` but I would rather solve that properly once we have got everything else stable.

PR #48, opened 2018-01-18
build: Fix makefile, improve CI and enforce style

jan-auer:
This PR fixes a couple of related issues:

 - CI runs rust tests, python tests and the python wheel separately now. Also it leverages build caches to speed up the tests. Unfortunately, since we are building the linux wheels inside a docker container, we can't leverage caches there.
 - The makefile was bogus and travis configuration confusing. The new ones should be more straight forward to read
 - We check the rustfmt style now with every build before even running tests. This means, all code is now reformatted - but the differences are actually not too bad in most cases. The only place where it really got worse is `common::Arch`, but that needs a refactor anyway.


jan-auer:
Note: I still have to verify if all artifacts arrive correctly in Zeus before we merge this.

jan-auer:
Note that `rustfmt` currently needs to run in its own job (it is platform independent anyway) because (a) it requires nightly to compile and (b) Travis has an old version preinstalled, so we have to install it with `--force`.

PR #55, opened 2018-02-09
fix(debuginfo): Match Apple's hidden symbols correctly

jan-auer:
This was definitely broken in previous releases and also in `sentry-cli` since we migrated to `symbolic`. Since our symbol iterator strips the first `"_"` in the symbol name, the regular expression never matched.

Also adding a trailing `"_"`, as hidden symbols seem to have this. @mitsuhiko can you confirm?


jan-auer:
That's because in all other cases we did strip it before. What I missed was that we didn't do that when checking for hidden symbols in sentry-cli.

PR #63, opened 2018-03-22
fix(debuginfo): Fix ID/UUID backward compatibility

jan-auer:
- [x] Add a utility to normalize IDs
- [x] Additionally support legacy UUIDs
- [x] Parse debug ids without dashes
- [x] Fix tests


jan-auer:
@mitsuhiko This is complete now.

PR #64, opened 2018-03-23
test(minidump): Add test fixtures and tests for minidumps and CFI

jan-auer:
Needs #63 for tests to pass.


jan-auer:
@mitsuhiko This is ready for review now.

PR #68, opened 2018-04-20
ref: Migrate from error_chain to failure

jan-auer:


jan-auer:
CABI still doesn't work because `cbindgen` is not able to parse the crate.

PR #69, opened 2018-04-23
feat: Create features for every top-level crate

jan-auer:
Introduces features for every top-level crate. Since `common` is always needed, there is no feature for that. In additon, each feature might have dependencies on other features:

- `debuginfo`: none
- `demangle`: none
- `minidump`: `debuginfo`
- `proguard`: none
- `sourcemap`: none
- `symcache`: `debuginfo`, `demangle`
- `with_serde`: activates inside `common`, `debuginfo` and `minidump`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ exclude = [
 [workspace]
 
 [features]
-default = []
+default = ["debuginfo"]
 debuginfo = ["symbolic-debuginfo"]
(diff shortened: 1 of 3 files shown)


jan-auer:
@mitsuhiko Was thinking of adding `default-features = ["debuginfo"]`, do you have an opinion?

mitsuhiko:
Wfm

PR #73, opened 2018-05-29
fix(debuginfo): Properly advance object iterators on error

jan-auer:
Fixes an endless loop in sentry-cli. I am testing with react native's `libjsinspector.a`, which for some reason cannot be parsed by goblin.


jan-auer:
I want to investigate what's happening with the archive file, since `otool` does not have problems parsing it.

PR #78, opened 2018-07-27
feat: Add tests for react native ios and android source maps

HazAT:
~Do not merge yet, the android test still fails.~
~This branch is used for a repro case for facebook.~


HazAT:
@jan-auer @mitsuhiko This should be good to merge now. 
I found a way to fix it.

PR #79, opened 2018-07-27
feat(minidump): Expose frame register values

jan-auer:
screen shot 2018-07-27 at 18 09 18


jan-auer:
@mitsuhiko This is ready for review now.

jan-auer:
Register names are without `$` now.

PR #92, opened 2018-10-09
feat(debuginfo): Add human output for object classes

jan-auer:


common/src/types.rs
```diff
@@ -613,6 +613,18 @@ impl ObjectClass {
...
 
+    pub fn human_name(self) -> &'static str {
+        match self {
+            ObjectClass::None => "file",
+            ObjectClass::Relocatable => "object",
+            ObjectClass::Executable => "executable",
+            ObjectClass::Library => "library",
+            ObjectClass::Dump => "dump",
+            ObjectClass::Debug => "debug companion",
+            ObjectClass::Other => "file",
```
jan-auer: Not sure about this, but I guess this is a reasonable human readable file name. See https://github.com/getsentry/sentry-cli/pull/412 for a potential use case.

PR #94, opened 2018-10-17
feat: Support ELF and Breakpad symbols

jan-auer:
This PR adds support for reading symbol tables from ELF (symtab) and Breakpad objects (PUBLIC records).

Work in progress.


jan-auer:
Superseeded by https://github.com/getsentry/symbolic/pull/127

PR #97, opened 2018-10-24
fix(common): Add missing variants to Arch::from_str

jan-auer:


jan-auer:
FTR, I liked `arm?` slightly more, but see the point :)

PR #105, opened 2018-11-09
feat: expose unreal as C-ABI

bruno-garcia:


py/symbolic/unreal.py
```diff
@@ -0,0 +1,21 @@
+from symbolic._lowlevel import lib, ffi
+from symbolic.utils import RustObject, rustcall
+
+__all__ = ['Unreal4Crash']
+
+class Unreal4Crash(RustObject):
```
jan-auer: This needs a

```py
    __dealloc_func__ = lib.semaphore_unreal4crash_free
```

py/symbolic/unreal.py
```diff
@@ -0,0 +1,21 @@
...
+
+    @classmethod
+    def from_bytes(cls, buffer):
+        """Parses an Unreal Engine 4 crash"""
+        buffer = bytes(buffer)
+        return cls._from_objptr(rustcall(lib.symbolic_unreal4_crash_from_bytes,
+                                buffer, len(buffer)))
```
jan-auer: In other cases we do `ffi.from_buffer(buffer)` to pass in the buffer. But if it works, it works.

py/symbolic/unreal.py
bruno-garcia: good catch. Except u've been doing too much `semaphore` hehe
`symbolic_unreal4_crash_free`

py/symbolic/unreal.py
bruno-garcia: Not sure what's the difference here. maybe @mitsuhiko has some insight?

jan-auer: https://cffi.readthedocs.io/en/latest/ref.html#ffi-buffer-ffi-from-buffer
It basically ensures to give you the raw pointer, no matter what you pass in as long as it conforms to the buffer interface.

PR #110, opened 2018-11-29
feat(debuginfo): Support zlib compressed ELF sections

jan-auer:
The changes in `symcache/dwarf.rs` got a little out of hand since gimli's structs only operate on slices but we need to hold on to the buffers at the same time.

The symbols still don't fully process, but this might be unrelated.


jan-auer:
It seems to work, but I want to run some more tests before merging this.

PR #117, opened 2018-12-19
feat(unreal): Expose logs via C-ABI

bruno-garcia:


cabi/src/unreal.rs
```diff
@@ -29,14 +29,23 @@ ffi_fn! {
...
         let context = unreal.get_context()?;
         Ok(SymbolicStr::from_string(serde_json::to_string(&context)?))
     }
 }
 
+ffi_fn! {
+    unsafe fn symbolic_unreal4_get_logs(unreal: *const SymbolicUnreal4Crash) -> Result<SymbolicStr> {
+        let unreal = &*(unreal as *const Unreal4Crash);
+
+        let context = unreal.get_logs(100)?;
```
jan-auer: Discussed offline. Since there is no reason to change this value anytime soon, we can leave it hard coded for now. We will re-evaluate this with the next major release of symbolic.

PR #129, opened 2019-03-19
feat(debuginfo): Implement PDB line information

jan-auer:
- [x] Create a PDB debugging session
- [x] Expose procedure symbols as functions
- [x] Load line information and append to functions
- [x] Prefer mangled names over demangled names

Inlining information will be exposed in a follow-up PR.


jan-auer:
Merging this once the next release of `pdb` is out.

PR #132, opened 2019-03-26
feat(minidump): Implement CFI extraction from PDB and PE

jan-auer:
This adds functionality to directly extract CFI from PDB and PE files without the need to convert to Breakpad symbols first. Based on the architecture (machine type), Microsoft compilers store frame data either in the PDB or the PE file.

- [x] From PDBs for 32-bit programs
- [x] From PEs for 64-bit programs

Based on https://github.com/getsentry/symbolic/pull/129
Requires https://github.com/willglynn/pdb/pull/43


jan-auer:
Going to merge this once `pdb` and `goblin` are released.

PR #133, opened 2019-04-01
ref: Switch to updated CodeId signature

jan-auer:
Updates to the latest version of `debugid` which includes the updated `CodeId` type.

### DIFF
--- a/debuginfo/src/pe.rs
+++ b/debuginfo/src/pe.rs
@@ -65,11 +65,15 @@ impl<'d> PeObject<'d> {
         let header = &self.pe.header;
         let optional_header = header.optional_header.as_ref()?;
 
-        let timestamp = header.coff_header.time_date_stamp;
-        let size_of_image = optional_header.windows_fields.size_of_image;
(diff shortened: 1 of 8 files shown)


debuginfo/src/pe.rs
```diff
@@ -65,11 +65,15 @@ impl<'d> PeObject<'d> {
...
-        let size_of_image = optional_header.windows_fields.size_of_image;
-        let string = format!("{:08X}{:X}", timestamp, size_of_image);
+        let timestamp = header.coff_header.time_date_stamp.to_be_bytes();
+        let size_of_image = optional_header.windows_fields.size_of_image.to_be_bytes();
+        let first_size_byte = size_of_image.iter().position(|x| *x != 0).unwrap_or(3);
 
-        Some(CodeId::new(string))
+        let mut code_id = Vec::with_capacity(8);
+        code_id.extend(&timestamp);
+        code_id.extend(&size_of_image[first_size_byte..]);
```
jan-auer: This is maybe weird, since this effectively formatting as it was previously

PR #134, opened 2019-04-03
feat: Ability to open ByteView from open filehandle

untitaker:
This is useful for opening byteviews from unnamed tempfiles (files that have no hardlink).

### DIFF
--- a/common/src/byteview.rs
+++ b/common/src/byteview.rs
@@ -78,20 +78,8 @@ impl<'a> ByteView<'a> {
         ByteView::from_cow(Cow::Owned(buffer))
     }
 
-    /// Constructs a `ByteView` from any `std::io::Reader`.
-    ///
-    /// This currently consumes the entire reader and stores its data in an internal buffer. Prefer
-    /// `ByteView::open` when reading from the file system or `ByteView::from_slice` /
-    /// `ByteView::from_vec` for in-memory operations. This behavior might change in the future.
-    pub fn read<R: io::Read>(mut reader: R) -> Result<Self, io::Error> {
-        let mut buffer = vec![];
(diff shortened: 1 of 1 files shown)


common/src/byteview.rs
```diff
@@ -78,20 +78,8 @@ impl<'a> ByteView<'a> {
...
-        let mut buffer = vec![];
-        reader.read_to_end(&mut buffer)?;
-        Ok(ByteView::from_vec(buffer))
-    }
-
-    /// Constructs a `ByteView` from a file path by memory mapping the file.
-    pub fn open<P: AsRef<Path>>(path: P) -> Result<Self, io::Error> {
-        let file = File::open(path)?;
+    /// Constructs a `ByteView` from an open file handle by memory mapping the file.
+    pub fn from_file(file: File) -> Result<Self, io::Error> {
```
jan-auer:
```suggestion
    pub fn map_file(file: File) -> Result<Self, io::Error> {
```

common/src/byteview.rs
```diff
@@ -111,6 +99,23 @@ impl<'a> ByteView<'a> {
...
+    pub fn read<R: io::Read>(mut reader: R) -> Result<Self, io::Error> {
+        let mut buffer = vec![];
+        reader.read_to_end(&mut buffer)?;
+        Ok(ByteView::from_vec(buffer))
+    }
+
+    /// Constructs a `ByteView` from a file path by memory mapping the file.
+    pub fn open<P: AsRef<Path>>(path: P) -> Result<Self, io::Error> {
+        let file = File::open(path)?;
+        Ok(Self::from_file(file))
```
jan-auer:
```suggestion
        Self::from_file(file)
```

PR #139, opened 2019-05-07
ref(common): Revert to string-based CodeId

jan-auer:
`debugid` 0.5.2 is released already.


jan-auer:
Let's take the easy route now by reverting to the previous debugid build and discuss this in more detail afterwards.

PR #148, opened 2019-05-31
fix(debuginfo): Consider all program headers for ELF

jan-auer:
This reverts #120. As it turns out, the first `PT_LOAD` segment is always the correct one to use as load address.

This goes along the lines of this change: https://github.com/google/breakpad/commit/3444ed7cf1118f3f288bc6941c8e726f732ad8c6


untitaker:
Just for context, how do we deal with the cases that #120 was supposed to fix?

jan-auer:
@untitaker #120 was speculative and based on a wrong assumption, as it turns out.

PR #152, opened 2019-06-17
ref(unreal): Update unreal parsing

jan-auer:
This PR refactors the Unreal API and modernizes it a little. 

**Most important changes**

- AppleCrashReports are now a separate file type that is detected in the unreal code. Previously, they were classified as `Minidump`.
- Parsing invalid unreal crash report files no longer panics, by switching from `bytes::Buf` to `scroll`.

**Other changes**
- `Unreal4File` now owns a handle to its data. Since the offsets are already verified when parsing the entire crash, one doesn't need to hold on to the entire crash.
- zlib implementation is now switched to `flate2`, which is also used in `symbolic_debuginfo` and in `symbolicator`. One less dependency.
- Removed some shortcut-methods that are superfluous. E.g. instead of providing a function `Unreal4Crash.process_minidump`, one can get the minidump file and then process it using `ProcessState.from_minidump_buffer`. This creates cleaner separation of concerns. 
- Removed obsolete errors (one was dead before, one became dead by switching to `scroll`).

When updating `symbolic` prior to the 6.0 release, this was missed. Since this is a breaking change, we will likely have to bump symbolic to 7.0 after this.


jan-auer:
This is final now. Going to merge once this has been implemented (and thus tried out) in Sentry.

PR #153, opened 2019-06-18
Who's dumb? That was me

bruno-garcia:


jan-auer:
Thanks, but will be superseeded by https://github.com/getsentry/symbolic/pull/152

PR #155, opened 2019-07-01
fix(symcache): Allow functions with more than 65k lines

jan-auer:
This fixes three conditions:

 1. It allows functions with more than 65k line records by splitting the address range and allowing multiple function records.
 2. It splits functions with a code length of more than 65k bytes (using the same mechanism)
 3. It no longer emits any function records without line records.
 4. **bugfix**: do not skip to write line records between sibling inlined functions

The bugfix is achieved by creating an isolated line cache for every layer when recursing into inlinees. Since line records might be missing from existing symcaches, the **symcache version was bumped**.

### DIFF
--- a/symcache/tests/snapshots/test_writer__functions_linux.snap
+++ b/symcache/tests/snapshots/test_writer__functions_linux.snap
@@ -1,5 +1,5 @@
 ---
-created: "2019-07-01T13:31:24.623765Z"
+created: "2019-07-02T08:47:28.746328Z"
 creator: insta@0.8.1
 source: symcache/tests/test_writer.rs
 expression: FunctionsDebug(&symcache)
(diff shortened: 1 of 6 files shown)


symcache/tests/snapshots/test_writer__functions_linux.snap
```diff
@@ -478,13 +471,9 @@ expression: FunctionsDebug(&symcache)
             487b DumpOSInformation
             488d LogAppend
             48cc LogAppend
-            48e4 LogAppend
-            48fc LogAppend
-            4914 LogAppend
```
jan-auer: All functions that are removed from the snapshots did not contain any line records.

jan-auer:
Verified with an offending debug symbol.

PR #156, opened 2019-07-03
ref(minidump): Explicitly undefine register rules in CFI

jan-auer:
More elaborate description will follow tomorrow. 
I'm not 100% sure if we should actually add the zeros or leave the old behavior. Regardless, the default value for `.ra` needs to go.


jan-auer:
This needs to be revisited. As it turns out, this requires changes in `gimli`.

PR #165, opened 2019-10-09
Allow to pass msvc options to the msvc demangler

calixteman:


jan-auer:
@calixteman sorry for the exceptionally long response cycle on this one. Do you still need this patch? 

If so, I would slightly prefer to create more abstract flags, which we can then also pass into the other demanglers.

Swatinem:
Hi @calixteman sorry this was laying around for 2+ years by now.

I would close this for now. We expose a few abstracted options for the demangler. If you would like to have full freedom I would suggest to use the msvc demangler directly.

We have the abstraction layer in symbolic-demangle specifically to not have to expose the inner workings.

PR #174, opened 2019-12-12
feat(unreal): Expose context and logs parsing

RaduW:
We need to expose context and logs parsing in order to be able to store them separately (in some envelope) and then, at a latter time, parse them independently.


untitaker:
for context @jan-auer the intent here is to be able to parse the collection of unreal files into envelope attachments, then pull out and merge those attachments into the event. The reason for this indirection is because of order dependencies.

.gitignore
```diff
@@ -13,3 +13,6 @@ __pycache__/
 
 # C++
 .vscode/ipch
+
+# IDE files
+.idea/
```
jan-auer: _nit_: Please terminate with a newline.

unreal/src/logs.rs
```diff
@@ -30,10 +30,7 @@ pub struct Unreal4LogEntry {
     pub message: String,
 }
 
-pub(crate) fn parse_logs(
-    log_slice: &[u8],
-    limit: usize,
-) -> Result<Vec<Unreal4LogEntry>, Unreal4Error> {
+pub fn parse_logs(log_slice: &[u8], limit: usize) -> Result<Vec<Unreal4LogEntry>, Unreal4Error> {
```
jan-auer: _nit_: Can we make this an associated function?

```rust
impl Unreal4LogEntry {
  pub fn parse(slice: &[u8], limit: usize) -> Result<Vec<Self>, Unreal4Error> {
  }
}
```

PR #176, opened 2019-12-13
fix(unreal): Fix typo in ContextRuntimeProperties

RaduW:
Fixed typo changed `Unreal4ContextRuntimeProperties::misc_primary_cpu_brand` to `misc_primary_gpu_brand`.


jan-auer:
Adding @bruno-garcia to double-check.

jan-auer:
Discussed offline: Since this is a small change, we're adding the old attribute as deprecated again. Then, we don't have to perform a major bump.

PR #187, opened 2020-02-04
feat: Support react-native hermes sourcemaps

Swatinem:
This treats Hermes SourceMaps slightly differently, resolving the scope name via the bytecode offset.


sourcemap/src/lib.rs
```diff
@@ -114,46 +115,60 @@ impl SourceMapView {
     /// If the sourcemap is an index it is being flattened.  If flattening
     /// is not possible then an error is raised.
     pub fn from_json_slice(buffer: &[u8]) -> Result<Self, ParseSourceMapError> {
+        use sourcemap::DecodedMap::*;
```
jan-auer: Please do not import enum variants. Even though you're matching only on tuple variants in this case, having such imports is a common cause of issues with typos.

sourcemap/src/lib.rs
```diff
@@ -114,46 +115,60 @@ impl SourceMapView {
...
         Ok(SourceMapView {
-            sm: match sourcemap::decode_slice(buffer)? {
-                sourcemap::DecodedMap::Regular(sm) => sm,
-                sourcemap::DecodedMap::Index(smi) => smi.flatten()?,
+            dm: match sourcemap::decode_slice(buffer)? {
+                Regular(sm) => Regular(sm),
+                Index(smi) => Regular(smi.flatten()?),
+                Hermes(smh) => Hermes(smh),
+                _ => panic!("invalid sourcemap type"),
```
jan-auer: Please return an error rather than panicking.

sourcemap/src/lib.rs
```diff
@@ -114,46 +115,60 @@ impl SourceMapView {
...
             },
         })
     }
 
+    fn as_sm(&self) -> &sourcemap::SourceMap {
+        use sourcemap::DecodedMap::*;
+        match &self.dm {
+            Regular(sm) => sm,
+            Hermes(smh) => smh,
+            _ => unreachable!(),
```
jan-auer: Rather than having `unreachable!()` here, can we introduce our own enum instead? Or, alternatively, store the raw sourcemap plus a plain enum for `lookup_token_with_function_name`.

HazAT:
Confirmed, just tested it end to end, it works!

jan-auer:
Excellent stuff 🎉 

When bumping the `sourcemap` dependency, please also update the `cabi/Cargo.lock`!

PR #189, opened 2020-02-18
fix: Using Hermes SourceMaps with non-hermes stackframes

Swatinem:
This was a misunderstanding on my part. The `SourceMapHermes` actually is not specific to Hermes at all. It’s just a SourceMap with additional scoping information. So it can be used without Hermes.


HazAT:
This implies a sentry and sentry-cli release?
Correct?

Swatinem:
@HazAT  just `sentry`, the CLI is not affected.

HazAT:
Sweet

HazAT:
:shipit: cc @Swatinem

PR #192, opened 2020-03-03
fix(debuginfo): Decompress GNU compressed debug sections correctly

jan-auer:
This fixes decompression of `.zdebug*` GNU sections.

### DIFF
--- a/debuginfo/src/elf.rs
+++ b/debuginfo/src/elf.rs
@@ -254,17 +254,32 @@ impl<'d> ElfObject<'d> {
 
     /// Decompresses the given compressed section data, if supported.
     fn decompress_section(&self, section_data: &[u8]) -> Option<Vec<u8>> {
-        let container = self.elf.header.container().ok()?;
-        let endianness = self.elf.header.endianness().ok()?;
-        let context = Ctx::new(container, endianness);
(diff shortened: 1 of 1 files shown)


debuginfo/src/elf.rs
```diff
@@ -254,17 +254,32 @@ impl<'d> ElfObject<'d> {
 
     /// Decompresses the given compressed section data, if supported.
     fn decompress_section(&self, section_data: &[u8]) -> Option<Vec<u8>> {
-        let container = self.elf.header.container().ok()?;
-        let endianness = self.elf.header.endianness().ok()?;
-        let context = Ctx::new(container, endianness);
+        let (size, compressed) = if section_data.starts_with(b"ZLIB") {
+            // The GNU compression header is a 4 byte magic "ZLIB", followed by an 8-byte big-endian
+            // size prefix of the decompressed data. This adds up to 12 bytes of GNU header.
```
untitaker: I was confused by the last sentence here because the "up to" made it sound like it can be smaller while the code didn't allow for that.

jan-auer: Sorry, saw this too late. I can clarify this on master, if you like. 

In my mind it read as "the numbers are adding up" to a sum :)

untitaker: either way is fine

PR #196, opened 2020-03-17
build: Bump dependencies

jan-auer:


jan-auer:
Looks like we'll release `pdb` tomorrow. I'm waiting for that before continuing this.

Issue #194, opened 2020-03-12
[debuginfo] Lines are collapsed in DwarfUnit::resolve_lines


calixteman:
Identical consecutive lines are collapsed when resolving lines for a given range.
It's a pretty good idea to reduce data size but we lose information about addresses which belong to an inlinee.
For example, if we've:
```
0x1234: Line: 12, File: foo.c
0x1256: Line: 12, File: foo.c
```
we'll get only `0x1234: Line: 12, File: foo.c` but we don't know anymore that `0x1256` belongs to the original inlinee.
There are 2 ways to fix that:
- remove the collapsing;
- add an option when getting functions iterator to not collapse consecutive lines.

Wdyt ?

PR #200, opened 2020-03-30
feat(debuginfo): Allow to optionally collapse identical lines

calixteman:
It aims to fix: https://github.com/getsentry/symbolic/issues/194

### DIFF
--- a/debuginfo/src/dwarf.rs
+++ b/debuginfo/src/dwarf.rs
@@ -507,7 +507,11 @@ impl<'d, 'a> DwarfUnit<'d, 'a> {
     }
 
     /// Resolves line records of a DIE's range list and puts them into the given buffer.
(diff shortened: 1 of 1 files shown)


debuginfo/src/dwarf.rs
```diff
@@ -988,6 +999,20 @@ impl<'s> Iterator for DwarfUnitIterator<'s> {
 
 impl std::iter::FusedIterator for DwarfUnitIterator<'_> {}
 
+/// Options to use with functions_option.
+#[derive(Debug)]
+pub struct DwarfFunctionIteratorOption {
+    collapse_lines: bool,
```
Swatinem: this should be `pub`, or have some kind of builder pattern, otherwise you have no outside way to set it to false right now, do you?

calixteman: good catch, thx

calixteman:
Closed in favor of https://github.com/getsentry/symbolic/pull/237

Issue #195, opened 2020-03-12
[debuginfo] Don't return an error when it isn't possible to resolve a name


calixteman:
Some functions are skipped because an error is returned when trying to resolve a function name using a reference.
I met some case where the references (via abstract_origin) are wrong (they don't point to a DIE).
If such a function is an inlinee then the inliner is skipped which is pretty bad.
I think that in such a case the `resolve_function_name` should return something like `<name omitted>`, it's probably better to have an incomplete information instead of removing the inliner.
We could do the same for `resolve_file`.

PR #201, opened 2020-03-30
[debuginfo] No error in functions iterator on name errors

calixteman:
It aims to fix: https://github.com/getsentry/symbolic/issues/195

### DIFF
--- a/debuginfo/src/dwarf.rs
+++ b/debuginfo/src/dwarf.rs
@@ -507,11 +507,11 @@ impl<'d, 'a> DwarfUnit<'d, 'a> {
     }
 
     /// Resolves line records of a DIE's range list and puts them into the given buffer.
-    fn resolve_lines(&self, ranges: &[Range]) -> Result<Vec<LineInfo<'d>>, DwarfError> {
+    fn resolve_lines(&self, ranges: &[Range]) -> Vec<LineInfo<'d>> {
         // Early exit in case this unit did not declare a line program.
         let line_program = match self.line_program {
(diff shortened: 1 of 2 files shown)


debuginfo/src/dwarf.rs
```diff
@@ -643,18 +647,21 @@ impl<'d, 'a> DwarfUnit<'d, 'a> {
...
             // allows for a single allocation per call to `resolve_lines`.
-            let lines = self.resolve_lines(&range_buf)?;
+            let lines = self.resolve_lines(&range_buf);
 
             let function = Function {
                 address: function_address,
                 size: function_size,
-                name: Name::with_language(name.unwrap_or_default(), self.language),
+                name: Name::with_language(
+                    name.unwrap_or_else(|| Cow::from("<name omitted>")),
```
jan-auer: Could you leave this an empty string and coerce in the caller? We rely on this being empty in Sentry since we show our own placeholders which have semantic meaning.

debuginfo/src/dwarf.rs
```diff
@@ -524,7 +524,11 @@ impl<'d, 'a> DwarfUnit<'d, 'a> {
             lines.reserve(rows.len());
 
             for row in rows {
-                let file = self.resolve_file(row.file_index)?.unwrap();
+                let file = self
+                    .resolve_file(row.file_index)
+                    .ok()
+                    .flatten()
```
jan-auer: Since both usages now only require an option, can you change the function signature to `Option<FileInfo>`?

PR #202, opened 2020-04-01
fix(debuginfo): Avoid infinite recursion because of self references

calixteman:


debuginfo/src/dwarf.rs
```diff
@@ -370,7 +374,12 @@ impl<'d, 'a> UnitRef<'d, 'a> {
 
         if let Some(attr) = reference_target {
             let resolved = self.resolve_reference(attr, |ref_unit, ref_entry| {
-                ref_unit.resolve_function_name(ref_entry)
+                if self.unit.offset != ref_unit.unit.offset || entry.offset() != ref_entry.offset()
+                {
+                    ref_unit.resolve_function_name(ref_entry)
+                } else {
+                    Err(DwarfErrorKind::SelfReference.into())
```
jan-auer, line 381: Just to confirm, are you sure you don't just want to return `Ok(None)` here?

calixteman: Or maybe just return an Option instead of a Result since  `resolve_function_name` is only used in `functions()` and remove the `ok().flatten()`. I can do that in a follow-up PR. Wdyt ?

jan-auer: I'd keep the result if the Dwarf is really invalid (used in `resolve_reference`). But it's already a `Result<Option<_>>`, so just returning `Ok(None)` here should be the ideal compromise imho.

PR #209, opened 2020-04-06
docs: Inline docs from sub-crates

flub:
In rust 2018 edition re-exports of dependent crates are not
automatically inlined.  So we need to give this a hint.


flub:
Neither did I before today :)

PR #211, opened 2020-04-06
test(symcache): Use criterion for benchmark testing

jan-auer:


symcache/benches/bench_writer.rs
```diff
@@ -1,45 +1,55 @@
...
-    b.iter(|| {
+fn bench_write_linux(c: &mut Criterion) {
+    c.bench_function("write_linux", |b| {
         let buffer = ByteView::open("../testutils/fixtures/linux/crash.debug").expect("open");
-        let object = Object::parse(&buffer).expect("parse");
-        SymCacheWriter::write_object(&object, Cursor::new(Vec::new()))
-            .expect("write_object")
-            .into_inner()
+        b.iter(|| {
+            let object = Object::parse(&buffer).expect("parse");
```
untitaker, line 13: Not sure if this should be outside of iter

Swatinem: does it make sense to benchmark parse and write separately? According to the name, it should only be write? Or is the parse step lazy?

jan-auer: Excellent point. The intention is to measure the full roundtrip here, to come close to what Symbolicator is doing. Additionally, `Object::parse` is not doing much, most of the work is happening inside the writer anyway (by opening a `DebugSession`), and I want to test the full roundtrip, including paging in the file.

The names could potentially be a bit better, but [that's another story for another time](https://www.urbandictionary.com/define.php?term=but%20that%27s%20another%20story%20for%20another%20time).

PR #212, opened 2020-04-07
feat(common): Add Path utilities for dSYM structures

jan-auer:
Implements utilities for dealing with `dSYM` folder structures. They can check for a dSYM folder structure, resolve the dSYM parent from a deep path, and resolve the deep path from the parent folder.

`sentry-cli` partially implements this logic already, so we can reduce some duplicate code here. Also, the examples used to require that one declares the full path to a dSYM file, but this is just unnecessary.  

_Users_jauer_Coding_symbolic_target_doc_symbolic_common_trait DSymPathExt html

### DIFF
--- a/common/src/path.rs
+++ b/common/src/path.rs
@@ -342,7 +342,7 @@ impl DSymPathExt for Path {
         let framework = self.file_name()?;
 
         let mut parent = self.parent()?;
-        if !parent.ends_with("/Contents/Resources/DWARF") {
+        if !parent.ends_with("Contents/Resources/DWARF") {
             return None;
         }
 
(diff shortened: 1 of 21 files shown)


common/src/path.rs
```diff
@@ -259,6 +260,104 @@ pub fn shorten_path(path: &str, length: usize) -> Cow<'_, str> {
...
+    fn is_dsym_dir(&self) -> bool;
+
+    /// Resolves the path of the debug file in a `dSYM` directory structure.
+    ///
+    /// Returns `Some(path)` if this path is a dSYM directory according to [`is_dsym_dir`], and a
+    /// file of the same name is located at `Contents/Resources/DWARF/`.
+    ///
+    /// # Examples
+    ///
+    /// ```no_run
```
Swatinem, line 285: oh, and I thought these were actually real tests. Would you mind changing them in a way that they use any of the fixtures that we have? Would be nice to have coverage here.

examples/dump_cfi.rs
```diff
@@ -18,7 +18,8 @@ fn print_error(error: &Error) {
 fn dump_cfi<P: AsRef<Path>>(path: P) -> Result<(), Error> {
     let path = path.as_ref();
 
-    let buffer = ByteView::open(path)?;
+    let dsym_path = path.resolve_dsym();
+    let buffer = ByteView::open(dsym_path.as_deref().unwrap_or(path))?;
```
Swatinem, line 22: Is this `.as_deref().unwrap_or()` a common pattern that we would like to document?

jan-auer: I guess not, that's up to the user to choose what they want to do if the directory is not a dSYM. It's like `path.parent()`.

common/src/path.rs
jan-auer: These are doc tests, so I cannot use our fixtures (and it also makes little sense). These docs are modelled after the `std::path::Path` docs.

However, I agree that this should be tested, let me add some tests.

PR #214, opened 2020-04-07
fix(debuginfo): Do not assume sorted compilation units

jan-auer:
The `Object` interface was built on the invalid assumption that compilation units are generally sorted by non-overlapping address ranges. In fact, they are often overlapping and functions can be declared in any order.

This PR changes the guarantees and no longer sorts `functions()` returned by the `DebugSession` iterator. This will allow to further optimize this iterator in the future and remove some intermediate vectors. 

The `SymcacheWriter` is now more resilient against this, when merging the symbol table into written functions.


jan-auer:
The two test diffs are expected since we no longer sort by the function address in the iterator.

symcache/src/writer.rs
```diff
@@ -216,35 +214,41 @@ where
...
             .context(SymCacheErrorKind::BadDebugFile)?;
 
         for function in session.functions() {
             let function = function.context(SymCacheErrorKind::BadDebugFile)?;
+            writer.add_function(function)?;
+        }
+
+        // Sort the files to efficiently add symbols from the symbol table in linear time
+        // complexity. When the writer finishes, it will sort again with the added symbols.
+        writer.ensure_sorted();
```
jan-auer, line 228: This is where the magic happens:
1. We insert functions in the original order. No point in assuming they are sorted.
2. Sort them
3. Do one pass through functions and symbols and add the symbols
4. Flush the writer, which will sort again and do its thing.

symcache/src/writer.rs
```diff
@@ -563,20 +565,23 @@ where
...
-        // Only sort if functions do not already appear in order. They are sorted primarily by their
-        // start address, and secondarily by the index in which they appeared originally in the
-        // file.
         // To compute parent offsets after that, one can simply binary search by the parent_ref
         // handle, allowing for efficient unique lookups.
-        if !self.sorted {
-            self.functions
-                .sort_unstable_by_key(|handle| handle.original);
-        }
+        self.ensure_sorted();
```
jan-auer, line 584: Worst case, we might sort twice here, even though it's not necessary. However, we cannot reset the `sorted` flag since we need to fix up inline functions, and it should be fast enough in general.

debuginfo/src/object.rs
```diff
@@ -379,9 +379,11 @@ pub enum ObjectDebugSession<'d> {
 impl<'d> ObjectDebugSession<'d> {
     /// Returns an iterator over all functions in this debug file.
     ///
-    /// The iteration is guaranteed to be sorted by function address and includes all compilation
```
Swatinem, line 382: Well, if we would be very pedantic, removing this guarantee would be a breaking change?

debuginfo/src/pdb.rs
```diff
@@ -1009,7 +1009,7 @@ impl<'s> Unit<'s> {
 
         let mut functions = Vec::new();
         let mut stack = FunctionStack::new();
-        let mut proc_offsets = SmallVec::<[_; 3]>::new();
+        let mut proc_offsets = SmallVec::<[_; 10]>::new();
```
Swatinem, line 1012: can you add a comment explaining the magic number here?

debuginfo/src/object.rs
jan-auer: That guarantee was always broken, this is what this PR is about. I'm thinking since it's broken and the symcache writer already partially handled unsorted functions, we can just remove this here and do a minor bump with a warning notice. 

The next major bump will have changes to the iterators themselves.

debuginfo/src/pdb.rs
jan-auer: Actually, this change is bogus.

PR #215, opened 2020-04-07
build(demangle): Bump cpp_demangle to fix clone suffixes

jan-auer:
Bumps `cpp_demangle` to fix parsing of symbols with multiple clone suffixes.

Ref https://github.com/gimli-rs/cpp_demangle/pull/194


jan-auer:
Do not merge until the PR in cpp_demangle is merged and published to crates.io.

PR #217, opened 2020-04-09
fix(common): Avoid panics for UTF-8 characters in paths

jan-auer:
Fixes panics when joining or normalizing paths when they contain multi-byte unicode characters. Along with this, some common code is refactored to reduce code duplication.

The behavior of some of these functions is suboptimal, particularly `join_path` and `shorten_path`. The next step could be to refactor those methods internally to fix the known broken tests and odd behavior. Next, we could create a proper library for a unified `Path` type.

### DIFF
--- a/common/src/path.rs
+++ b/common/src/path.rs
@@ -117,7 +117,7 @@ pub fn join_path(base: &str, other: &str) -> String {
         "{}{}{}",
         base.trim_end_matches(is_path_separator),
         if is_windows { '\\' } else { '/' },
-        other.trim_end_matches(is_path_separator)
+        other.trim_start_matches(is_path_separator)
     )
 }
 
--- a/common/src/path.rs
+++ b/common/src/path.rs
@@ -1,38 +1,83 @@
 use std::borrow::Cow;
 use std::path::{Path, PathBuf};
 
-fn is_absolute_windows_path(s: &str) -> bool {
-    // UNC
-    if s.len() > 2 && (&s[..2] == "\\\\" || &s[..2] == "//") {
-        return true;
+trait ToChar {
+    fn to_char(self) -> char;
+}
+
+impl ToChar for char {
+    fn to_char(self) -> char {
+        self
     }
+}
 
-    // other paths
-    let mut char_iter = s.chars();
-    let (fc, sc, tc) = (char_iter.next(), char_iter.next(), char_iter.next());
+impl ToChar for u8 {
+    fn to_char(self) -> char {
+        char::from(self)
+    }
+}
 
-    match fc.unwrap_or_default() {
-        'A'..='Z' | 'a'..='z' => {
-            if sc == Some(':') && tc.map_or(true, |tc| tc == '\\' || tc == '/') {
-                return true;
-            }
+impl<T: ToChar + Copy> ToChar for &'_ T {
+    fn to_char(self) -> char {
+        (*self).to_char()
+    }
+}
+
+/// Returns `true` if the given character is any valid directory separator.
+#[inline]
+fn is_path_separator<C: ToChar>(c: C) -> bool {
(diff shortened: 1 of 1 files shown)


common/src/path.rs
```diff
@@ -1,38 +1,83 @@
 use std::borrow::Cow;
 use std::path::{Path, PathBuf};
 
-fn is_absolute_windows_path(s: &str) -> bool {
-    // UNC
-    if s.len() > 2 && (&s[..2] == "\\\\" || &s[..2] == "//") {
-        return true;
+trait ToChar {
+    fn to_char(self) -> char;
```
jan-auer, line 5: This trait is introduced to allow abstracting over `u8`, `&u8`, `char` and `&char`. This makes it possible to build generic functions that work in iterators and matchers on both binary slices and strings. Unfortunately, I was not able to find another way to do this with builtin traits like `Borrow` and `Into<char>`.

common/src/path.rs
```diff
@@ -1,38 +1,83 @@
...
-                return true;
-            }
+impl<T: ToChar + Copy> ToChar for &'_ T {
+    fn to_char(self) -> char {
+        (*self).to_char()
+    }
+}
+
+/// Returns `true` if the given character is any valid directory separator.
+#[inline]
```
flub, line 27: out of curiosity, do you have any reference on when this is useful?

common/src/path.rs
```diff
@@ -65,25 +110,19 @@ pub fn join_path(base: &str, other: &str) -> String {
...
-        )
-    }
+    // Always trim by both separators, since as soon as the path is Windows, slashes also count as
+    // valid path separators. However, use the main separator for joining.
+    let is_windows = is_windows_path(base) || is_windows_path(other);
+    format!(
+        "{}{}{}",
+        base.trim_end_matches(is_path_separator),
+        if is_windows { '\\' } else { '/' },
+        other.trim_end_matches(is_path_separator)
```
Swatinem, line 120:
```suggestion
        other.trim_start_matches(is_path_separator)
```

Also please add a test that covers this.

common/src/path.rs
```diff
@@ -442,6 +498,9 @@ mod tests {
         assert_eq!(shorten_path("/foobarbaz/blahblah", 16), ".../blahblah");
         assert_eq!(shorten_path("/foobarbazblahblah", 12), "...lahblah");
         assert_eq!(shorten_path("", 0), "");
+
+        assert_eq!(shorten_path("아이쿱 조합원 앱카드", 9), "아...");
+        assert_eq!(shorten_path("아이쿱 조합원 앱카드", 20), "...ᆸ카드");
```
Swatinem, line 503: is this a right-to-left language? I’m confused why the result is flipped.

common/src/path.rs
jan-auer: Since it's an internal function, it is most likely inlined anyway during release optimizations, but I'm trying to hint a little more strongly that I want these predicates inlined. I don't have any good reference at the moment, but will look for some.

common/src/path.rs
jan-auer: This is just the way the function works, it's what I meant with "odd" behavior. For short `length <= 10`, it truncates in the back, and for `length > 10` it truncates in the front.  I'm planning to change it in the next step, but for now this PR should only fix the panics.

common/src/path.rs
jan-auer: Good. Catch. I'm glad the tests didn't catch this.

jan-auer: I think this doesn't even make sense.

common/src/path.rs
Swatinem: ah, I should have paid more attention then ;-) nvm

PR #218, opened 2020-04-09
feat(minidump): Get CFI info from eh_frame even if err in debug_frame

calixteman:
In case of error in `.debug_frame` section the CFI info in `.eh_frame` weren't retrieved.

### DIFF
--- a/minidump/src/cfi.rs
+++ b/minidump/src/cfi.rs
@@ -232,11 +232,14 @@ impl<W: Write> AsciiCfiWriter<W> {
 
         // First load information from the DWARF debug_frame section. It does not contain any
         // references to other DWARF sections.
-        if let Some(section) = object.section("debug_frame") {
+        // Don't return on error because eh_frame can contain some information
+        let debug_frame_result = if let Some(section) = object.section("debug_frame") {
             let frame = DebugFrame::new(&section.data, endian);
             let info = UnwindInfo::new(object, section.address, frame);
-            self.read_cfi(&info)?;
-        }
+            self.read_cfi(&info)
+        } else {
+            Ok(())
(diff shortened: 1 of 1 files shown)


calixteman:
> Thank you! Patch looks good, thanks for ensuring that the error is eventually returned.
> 
> However, are you sure that the caller can swallow the error and still use the partially written CFI? Generally, the writer can fail at places that may leave invalid or partial records behind, which may lead to errors when feeding them into Breakpad.
>

I'd say that's ok.
Most of the CFI for firefox at least are in the eh_frame section (the same gcc & clang).
So in this case skipping CFI from debug_frame section is not bad.
The idea here is just to let the user decides what he wants to do: skip everything in case of error or keep what he has.
FYI, I fixed a bug in gimli which was causing trouble because of a "bad" debug_frame section:
https://github.com/gimli-rs/gimli/commit/53c802a69066b211984ef06fd96ba20946d47ca3
 
> I was briefly thinking about mentioning this in the description of the `process` method. Since it's quite uncertain that the CFI is still valid, however, I am now leaning towards not doing that. What do you think?

From my pov, CFI here is valid but incomplete so not so bad.

jan-auer:
> FYI, I fixed a bug in gimli which was causing trouble

Thanks, I'll bump `gimli` as soon as it is released.

> CFI here is valid but incomplete so not so bad

Okay, I'm merging this now. From the interface, I still think that if we error at the wrong point, we might have emitted only a partial record (i.e. stopped in the middle of printing a register name). In that case, you'd be left with invalid CFI. 

As you said, based on the error, the user can then decide if they still want to use the written CFI or not. 👍

PR #220, opened 2020-04-20
fix(debuginfo): Calculate line record sizes for DWARF

jan-auer:
Calculates line record sizes for DWARF by comparing with the previous row emitted from the line number program. Additionally, the sequence end point is used to calculate the length for the last row/record.

### DIFF
--- a/debuginfo/src/dwarf.rs
+++ b/debuginfo/src/dwarf.rs
@@ -176,6 +176,7 @@ struct DwarfRow {
     address: u64,
     file_index: u64,
     line: Option<u64>,
+    size: Option<u64>,
 }
 
 /// A sequence in the DWARF line program.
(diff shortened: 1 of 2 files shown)


debuginfo/src/dwarf.rs
```diff
@@ -202,6 +203,13 @@ impl<'d, 'a> DwarfLineProgram<'d> {
 
         while let Ok(Some((_, &program_row))) = state_machine.next_row() {
             let address = program_row.address();
+
+            if let Some(last_row) = sequence_rows.last_mut() {
+                if address >= last_row.address {
+                    last_row.size = Some(address - last_row.address);
```
flub, line 209: This allows a row-size of 0.  I assume this is intentional

PR #221, opened 2020-04-20
fix(debuginfo): Detect all scopes when collecting inlinees

jan-auer:
Fixes an issue where in some cases the inline tree might have been composed wrong. This was particularly the case in DWARF information from Rust programs, as those often contain nested scopes within functions.

Previously, we only considered functions entries when flushing inlinees into the function stream. This was violated in a tree like this:

```
> parent
  > inlinee 1
  > scope
    > inlinee 2
```

Invalid nesting:
```
> parent
  > inlinee 1
    > inlinee 2
```

Correct nesting:
```
> parent
  > inlinee 1
  > inlinee 2
```

### DIFF
--- a/debuginfo/src/dwarf.rs
+++ b/debuginfo/src/dwarf.rs
@@ -613,17 +613,17 @@ impl<'d, 'a> DwarfUnit<'d, 'a> {
                 _ => skipped_depth = None,
             }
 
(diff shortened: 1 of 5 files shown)


jan-auer:
Test diffs are actually OK. Since certain inlinees now have moved out of a function to become its sibling, the function ranges have changed as a result.

PR #208, opened 2020-04-05
feat: Support Swift 5.2 mangling

jan-auer:
Updates to Swift 5.2.1. Along with this, there are a few changes:

- Added a script that auto-imports files and headers. Still requires to build the entirety of Swift, though :(
- Requiring a C++14 compiler now. It's no longer sustainable to keep backwards compatibility.
- Switched to Manylinux2010 since the Manylinux1 images cannot run a C++14 compiler.

Fixes #204

PR #236, opened 2020-05-18
meta(demangle): Recommend to skip swift build

jan-auer:
Applies https://github.com/getsentry/symbolic/pull/208#discussion_r426354077, courtesy of @kastiglione.

> I think you can use `--skip-build` here. The headers in `build/Ninja-DebugAssert/*-macosx-x86_64/include` are still generated when `--skip-build` is used. This avoids building all of swift.

Thanks!

### DIFF
--- a/demangle/vendor/swift/README.md
+++ b/demangle/vendor/swift/README.md
@@ -32,7 +32,7 @@ patch is maintained in `1-arguments.patch`.
    5. Build the complete swift project (be very patient, this may take long):
       ```
       $ cd swift
-      $ ./utils/build-script
+      $ ./utils/build-script --skip-build
       ```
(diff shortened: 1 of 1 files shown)


demangle/vendor/swift/README.md
```diff
@@ -32,7 +32,7 @@ patch is maintained in `1-arguments.patch`.
    5. Build the complete swift project (be very patient, this may take long):
       ```
       $ cd swift
-      $ ./utils/build-script
+      $ ./utils/build-script -S
```
flub, line 35: fwiw I always prefer long option names in docs or scripts, `--skip-build` explains a lot more then `-S`.

Issue #228, opened 2020-04-27
Invalid line size and extra line with version 7.3.0


calixteman:
I've this data in a so file:
```
Name: _ZN7mozilla14VolatileBufferD2Ev
Address: 0x4f44130
Size: 0x24
Line: LineInfo { address: 0x4f44130, size: 0x9, file: FileInfo { name: "memory/volatile/VolatileBufferFallback.cpp", dir: "/home/calixte/dev/mozilla/mozilla-central.hg" }, line: 34 }
Line: LineInfo { address: 0x4f44139, size: 0x4, file: FileInfo { name: "memory/volatile/VolatileBufferFallback.cpp", dir: "/home/calixte/dev/mozilla/mozilla-central.hg" }, line: 37 }
Line: LineInfo { address: 0x4f44142, size: 0x1e, file: FileInfo { name: "memory/volatile/VolatileBufferFallback.cpp", dir: "/home/calixte/dev/mozilla/mozilla-central.hg" }, line: 38 }
Line: LineInfo { address: 0x4f44154, size: 0xc, file: FileInfo { name: "dist/include/mozilla/Mutex.h", dir: "/home/calixte/dev/mozilla/mozilla-central.hg/obj-x86_64-pc-linux-gnu" }, line: 57 }
Inlinees:
  Name: _ZN7mozilla16OffTheBooksMutexD2Ev
  Address: 0x4f44142
  Size: 0x12
  Line: LineInfo { address: 0x4f44142, size: 0x1e, file: FileInfo { name: "dist/include/mozilla/Mutex.h", dir: "/home/calixte/dev/mozilla/mozilla-central.hg/obj-x86_64-pc-linux-gnu" }, line: 57 }
```
Few bugs:
 i) the sum of the lines size exceeds the function size (0x9+0x4+0x1e+0xc=0x37 > 0x24)
 ii) this function contains an inlinee (at address 0x4f44142) and the line number in the main function is 38 when it's 57 in the inlinee
 iii) there is a spurious line at 0x4f44154

With a previous version, I get:
```
Name: _ZN7mozilla14VolatileBufferD2Ev
Address: 0x4f44130
Size: 0x24
Line: LineInfo { address: 0x4f44130, size: 0x9, file: FileInfo { name: "memory/volatile/VolatileBufferFallback.cpp", dir: "/home/calixte/dev/mozilla/mozilla-central.hg" }, line: 34 }
[...]

Issue #193, opened 2020-03-06
[debuginfo] Spurious line address in inlinee 


calixteman:
Just get the dbg here: https://symbols.mozilla.org/libxul.so/20C9B0FB137F982CC2C45FC3557AB8300/libxul.so.dbg.gz
and have a look at function at address `0x8e24d0`.
The first inlinee is at address `0x8e24f8` and the first line is at address `0x8e24e9` which is out of range.
We can this see spurious line in all the stack of the first inlinee.
Is it a bug ? or miss I something ? or did I misunderstand something ?

PR #237, opened 2020-05-19
fix(debuginfo): Inlinees management in DWARF

calixteman:
It aims to fix:
 - addresses which don't belong to the inlinee (it's due to the fact that in get_rows we get row just before inlinee start address),
 - the inlinee addresses in the parent.

So the idea is basically to get rows for non-inlined functions and then for an inlinee get line info from the parent according to the ranges we've for the inlinee.
I let a TODO, I'll fix in a follow-up patch: some non-inlined functions can have several range which messes up the computation of size/end. 

Fixes https://github.com/getsentry/symbolic/issues/228
Fixes https://github.com/getsentry/symbolic/issues/194
Fixes https://github.com/getsentry/symbolic/issues/193

Ref https://github.com/getsentry/symbolic/pull/223


jan-auer:
@calixteman Thanks for getting this started. I'm not so sure about the changes in test snapshots, since they show line records that seem completely out of line.

I think that the change you were doing in #200 is the actual fix, essentially removing this check here:

https://github.com/getsentry/symbolic/blob/d16bc475779ed7dfcf626491c824e8dedb0c7d8b/debuginfo/src/dwarf.rs#L560-L562

That way, we should no longer get overlapping ranges (DWARF usually splits ranges in the parent when there's an inlinee boundary), and all the complicated logic from #223 should no longer be necessary.

calixteman:
I didn't pay attention to the tests but there is something obviously wrong here.
Anyway the patch aims to fix too the spurious addresses which out-of-range: it's due the fact that get_rows gets a row were the range is inside and then the row.begin is used (probably it should be max(row.begin, range.begin)).

calixteman:
There is still something in the tests... let me fix that

calixteman:
Normally this patch should fix what you tried to fix in https://github.com/getsentry/symbolic/pull/223.
And it fixes too the spurious out-of-range lines and I think we've at the end better informations.
So if you see a way to make this code simpler, then I'd be happy.
Anyway we need to find a way to handler non-inlined function on several ranges.

debuginfo/src/dwarf.rs
```diff
@@ -635,6 +661,16 @@ impl<'d, 'a> DwarfUnit<'d, 'a> {
...
             let (call_line, call_file) = self.parse_ranges(entry, range_buf)?;
 
+            // TODO: fix that
+            // We've a non-inlined function which has two ranges or more
+            // So probably splited because of cold paths
+            // Anyway the function_size & function_end are wrong here
+            // so probably the workaround to generate several functions here:
+            // one for each range
+            if !inline && range_buf.len() != 1 {
+                continue;
```
jan-auer, line 671: Can you join this with the next check for an empty `range_buf`, or at least also set `skipped_depth`? This way, you can remove the second part of this TODO, where you skip in the inline child.

I've quickly run a few tests, and it seems that this doesn't produce any false positives on the files that I've checked it on. However, maybe we want to check for explicit _split ranges_? At least in inline functions, it's quite common to get adjacent ranges, and I'm not sure if we can trust that this doesn't happen for top-level functions as well.

debuginfo/src/dwarf.rs
```diff
@@ -675,100 +711,136 @@ impl<'d, 'a> DwarfUnit<'d, 'a> {
...
-            };
-
             if inline {
                 // An inlined function must always have a parent. An empty list of funcs
                 // indicates invalid debug information.
                 let parent = match stack.peek_mut() {
                     Some(parent) => parent,
-                    None => return Err(DwarfErrorKind::UnexpectedInline.into()),
+                    // TODO: (part of the todo above)
+                    None => continue, // return Err(DwarfErrorKind::UnexpectedInline.into()),
```
jan-auer, line 720: This can be reverted if you set `skipped_depth` (see comment above).

debuginfo/src/dwarf.rs
```diff
@@ -553,21 +552,48 @@ impl<'d, 'a> DwarfUnit<'d, 'a> {
             let rows = line_program.get_rows(range);
             lines.reserve(rows.len());
 
-            for row in rows {
-                let file = self.resolve_file(row.file_index).unwrap_or_default();
-                let line = row.line.unwrap_or(0);
+            if let Some((first, rows)) = rows.split_first() {
+                let file = self.resolve_file(first.file_index).unwrap_or_default();
+                let line = first.line.unwrap_or(0);
+                let size = first.size.map(|s| s + first.address - range.begin);
```
jan-auer, line 558: Can we reduce the code duplication here? If I'm reading this correctly, then there are two changes in this:
 1. `last` is moved into the loop around `ranges`. 
 2. You patch line size of the previous record. You could probably do that essentially in every iteration when you have a new record, before the `continue` block.

debuginfo/src/dwarf.rs
```diff
@@ -675,100 +711,136 @@ impl<'d, 'a> DwarfUnit<'d, 'a> {
...
-                        }
-                    };
-
-                    // Patch all following line records pointing them to the call location. Stop at
-                    // the boundary of the inlinee.
-                    for index in start..lines.len() {
-                        let record = &mut lines[index];
-                        if record.address >= function_end {
-                            break;
+                    for range in range_buf.iter() {
```
jan-auer, line 735: In this loop, we now perform two binary searches per range. Especially for functions with many functions, it may be considerably faster to just iterate sequentially through parent lines.

What if we stick with the initial binary search of the previous implementation, but then we change the for loop (`for index in start..lines.len()` in the original code) to check for membership in the respective range, or otherwise advance the range? This might help to keep the diff smaller.

debuginfo/src/dwarf.rs
jan-auer: Discussed offline: Could you also add a comment that explains why the first row is corrected to align with the start of the `range`.

jan-auer:
This last commit fixes the Python tests. When splitting a line info record in two, we must not skip the second half, since this one may need to be split further. The only two cases where skipping is valid is when adding records at the border of a range (since those come out of nowhere).

PR #239, opened 2020-05-26
fix(debuginfo): Three-way split line records for inlinee-ranges

calixteman:
When an inline range is strictly included in a parent line record, it must be split into three parts and the end of the original line record must be retained for the last part. This fixes two distinct issues in the previous implementation:

1. The last record was always aligned with the end of the inlinee-range, which either made it too long or left a gap.
2. The third split was also patched with call site information, instead of retaining the original line info.

**Before patching**:
```
          +-------------------+
 Parent:  |         P         |
          +---+-----------+---+
Inlinee:      |     I     |
              +-----------+
```

**After patching**:
```
          +---+-----------+---+
 Parent:  | P |     C     | P |
          +-------------------+
Inlinee:      |     I     |
              +-----------+
```

(`P` parent line record, `C` Call site record, `I` inlinee line record)

Follow-up on https://github.com/getsentry/symbolic/pull/237

### DIFF
--- a/debuginfo/src/dwarf.rs
+++ b/debuginfo/src/dwarf.rs
@@ -776,6 +776,22 @@ impl<'d, 'a> DwarfUnit<'d, 'a> {
                                 continue;
                             }
 
+                            // Split the parent record if it exceeds the end of this range. We can
+                            // assume that record.size is set here since we passed the previous
+                            // condition.
+                            let split = if record_end > range_end {
+                                record.size = Some(range_end - record.address);
+
+                                Some(LineInfo {
+                                    address: range_end,
+                                    size: Some(record_end - range_end),
+                                    file: record.file.clone(),
+                                    line: record.line,
+                                })
+                            } else {
+                                None
+                            };
+
                             if record.address < range_begin {
                                 // Fix the length of this line record to go up to the start of the
                                 // inline function. This effectively splits the previous record in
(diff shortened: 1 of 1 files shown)


debuginfo/src/dwarf.rs
```diff
@@ -800,28 +816,12 @@ impl<'d, 'a> DwarfUnit<'d, 'a> {
...
-                                    address: range_end,
-                                    size: Some(record_end - range_end),
-                                    file: record.file.clone(),
-                                    line: record.line,
-                                })
+                                index += 1;
                             } else {
-                                None
+                                record.file = file.clone();
+                                record.line = line;
```
jan-auer, line 822: I believe that this still needs to run unconditionally. Otherwise, it looks like the middle record would not have correct file/line info.

jan-auer:
After discussing this for a while, we figured that your initial patch is indeed the correct solution for this, with one minor change to calculate the length when splitting at the beginning: 

```rust
let size = record_end.min(range_end) - range_begin;
```

In the future, a better approach would be to change this even more:
- Operate on ranges with a `start` and `end`, rather than `start` and `size`. We could use Rust's builtin ranges for this.
- Factor out the range intersection logic into a reusable component that can be unit-tested

In another step, we can consider if we even need to patch the parent line records nowadays.

PR #240, opened 2020-05-26
feat: Expose proguard remapper via c-abi

Swatinem:
Exposes the new `proguard::Mapper` via `cabi` and `py`.

Keeps the old `MappingView` and workspace crate untouched, so this can be released as a minor version, while deprecating and removing the other stuff at some later point.


py/tests/test_proguard.py
```diff
@@ -36,3 +36,20 @@ def test_mmap(res_path):
...
+    mapper = ProguardMapper.open(os.path.join(res_path, "proguard.txt"))
+    assert mapper.has_line_info
+    assert mapper.uuid == uuid.UUID("5cd8e873-1127-5276-81b7-8ff25043ecfd")
+
+    assert (
+        mapper.remap_class("android.support.constraint.ConstraintLayout$a")
+        == "android.support.constraint.ConstraintLayout$LayoutParams"
+    )
+
+    remapped = mapper.remap_frame("android.support.constraint.a.b", "a", 116)
```
bruno-garcia, line 51: Would be nice to have a test that covers inlined frames

cabi/include/symbolic.h
```diff
@@ -28,7 +28,7 @@ enum SymbolicErrorCode {
   SYMBOLIC_ERROR_CODE_OBJECT_ERROR_UNSUPPORTED_OBJECT = 2101,
   SYMBOLIC_ERROR_CODE_OBJECT_ERROR_BAD_BREAKPAD_OBJECT = 2102,
   SYMBOLIC_ERROR_CODE_OBJECT_ERROR_BAD_ELF_OBJECT = 2103,
-  SYMBOLIC_ERROR_CODE_OBJECT_ERROR_BAD_MACH_OOBJECT = 2104,
+  SYMBOLIC_ERROR_CODE_OBJECT_ERROR_BAD_MACH_O_OBJECT = 2104,
```
Swatinem, line 31: not sure if this will cause any problems?

py/tests/test_proguard.py
Swatinem: I added this to the fixture, and tested it ;-)

cabi/include/symbolic.h
flub: I was wondering the same...

Swatinem: I manually reverted this. Although its an auto-generated file, so people will run into this again at some point.

cabi/Cargo.toml
```diff
@@ -30,3 +30,4 @@ serde_json = "1.0.40"
 failure = "0.1.5"
 apple-crash-report-parser = { version = "0.4.0", features = ["with_serde"] }
 symbolic = { version = "7.3.3", path = "..", features = ["debuginfo", "demangle", "minidump", "proguard", "sourcemap", "symcache", "unreal-serde"] }
```
jan-auer, line 32: Can the old C-ABI be shimmed with the new proguard version? It would be extremely nice to get rid of `symbolic-proguard` completely and deprecate the entire crate.

cabi/src/proguard.rs
```diff
@@ -77,3 +77,129 @@ ffi_fn! {
         Ok(SymbolicProguardMappingView::as_rust(view).has_line_info())
     }
 }
+
+use proguard::{ProguardMapper, ProguardMapping, StackFrame};
+
+/// Represents a Java Stack Frame.
+#[repr(C)]
+pub struct SymbolicJavaStackFrame {
+    pub klass: SymbolicStr,
```
jan-auer, line 86:
```suggestion
    pub class_name: SymbolicStr,
```

cabi/src/proguard.rs
```diff
@@ -77,3 +77,129 @@ ffi_fn! {
...
+        method: *const SymbolicStr,
+        line: usize,
+    ) -> Result<SymbolicProguardRemapResult> {
+        let mapper = &SymbolicProguardMapper::as_rust(mapper).mapper;
+        let frame = StackFrame::new((*class).as_str(), (*method).as_str(), line);
+
+        let mut frames: Vec<_> = mapper.remap_frame(&frame).map(|frame| {
+            SymbolicJavaStackFrame {
+                klass: frame.class().into(),
+                method: frame.method().into(),
```
jan-auer, line 149: This `SymbolicStr::from` only borrows the `&str`, but the underlying `Cow` will be dropped after the `map` closure has finished. You should clone the string, therefore.

```suggestion
                klass: frame.class().to_owned().into(),
                method: frame.method().to_owned().into(),
```

The same applies to `file` below, as well as `symbolic_proguardmapper_remap_class`. I'm a little bit curious why this does not segfault in the tests.

cabi/src/proguard.rs
```diff
@@ -77,3 +77,129 @@ ffi_fn! {
...
+}
+
+ffi_fn! {
+    /// Creates a proguard mapping view from a path.
+    unsafe fn symbolic_proguardmapper_open(
+        path: *const c_char
+    ) -> Result<*mut SymbolicProguardMapper> {
+        let path = CStr::from_ptr(path).to_str()?;
+        let source = std::fs::read_to_string(path)?;
+        let mapping = ProguardMapping::new(std::mem::transmute(source.as_str()));
```
jan-auer, line 119: This `transmute` hides a lifetime cast. Would you consider to use [`SelfCell`](https://docs.rs/symbolic/7.3.3/symbolic/common/struct.SelfCell.html) for this? Otherwise, it would be good to add a few comments about safety and lifetimes here, since this is quite easy to miss.

I believe this *should* be safe, though, since a constant `String` has a stable memory address.

cabi/include/symbolic.h
jan-auer: This is indeed strange. It would be a breaking change in the C-ABI, and it will generate a different name for the matching error in the Python layer. I think you can fix this with some sort of configuration for `cbindgen`, although the new version is actually correct.

jan-auer:
One more ask: What if we mark the old C-ABI functions and Python functions as deprecated? Once we do the next major bump, it's easier to find all the places to remove.

cabi/src/proguard.rs
Swatinem: > I'm a little bit curious why this does not segfault in the tests.

The borrowed string lives for the complete lifetime of the mapper. But yes, better be safe here.

Swatinem:
👍 I was able to shim the old functions quite easily, and also copied the `SelfCell` code over to the cabi.

PR #241, opened 2020-05-31
feat(demangle): Improve docs for symbolic-demangle

jan-auer:
Updates docs of `symbolic-demangle`, as they were out of date and sometimes incomplete. Particularly, this includes:

- More examples for the demangle functions
- Correct description of flags and return types
- A README generated from the crate docs using `cargo-readme`


demangle/Cargo.toml
```diff
@@ -25,5 +26,11 @@ msvc-demangler = "0.8.0"
 rustc-demangle = "0.1.16"
 symbolic-common = { version = "7.3.3", path = "../common" }
 
+[dev-dependencies]
+symbolic = { path = "..", features = ["demangle"] }
```
Swatinem, line 30: Just a heads up… I know this pattern is super useful, as I have also used it for the rust SDK, but you are not able to properly publish with this: https://github.com/rust-lang/cargo/issues/4242 😭

demangle/src/lib.rs
```diff
@@ -1,26 +1,27 @@
...
+//! - C++ (GCC-style compilers and MSVC)
+//! - Rust (both `legacy` and `v0`)
+//! - Swift (up to Swift 5.2)
+//! - ObjC (only symbol detection)
 //!
-//! As the demangling schemes for different languages are different, the
-//! feature set is also inconsistent.  In particular Rust for instance has
-//! no overloading so argument types are generally not expected to be
-//! encoded into the function name whereas they are in Swift and C++.
+//! As the demangling schemes for the languages are different, the supported demangling features are
```
Swatinem, line 10: how did you manage to format the markdown here? ;-)

demangle/src/lib.rs
```diff
@@ -181,23 +184,65 @@ fn try_demangle_objcpp(ident: &str, opts: DemangleOptions) -> Option<String> {
...
+    /// ```
+    ///
+    /// [module level documentation]: index.html
     fn detect_language(&self) -> Language;
 
     /// Demangles the name with the given options.
+    ///
+    /// Returns `None` in one of the following cases:
+    ///  1. The language cannot be detected.
+    ///  2. The language cannot be demangled.
```
Swatinem, line 216: the language?

demangle/Cargo.toml
jan-auer: Alright, let me revert this then. Too bad, it made the examples much nicer.

demangle/src/lib.rs
jan-auer: I'm using an extension called "Rewrap" in VSCode.

demangle/src/lib.rs
jan-auer: probably better to say "is not supported"

demangle/Cargo.toml
Swatinem: Yes it definitely does! I’m also not yet sure what I should do about this for the SDK, since I depend on `sentry::init` and `sentry::test` a lot in the doctests 😭

jan-auer: For the Sentry SDK, doing something similar to `futures` would probably be worth it. In symbolic, I guess people can extrapolate how to rewrite the imports until there's a proper solution for this in cargo.

PR #244, opened 2020-06-10
Update cpp_demangle::DemangleOptions so simbolic_demangle can compile.

o0Ignition0o:
cpp_demangle::DemangleOptions now has a no_return_type field which causes an issue in symbolic-demangle lib.rs
the breaking change occured in a patch version which means running `cargo
update` breaks people's builds.


khuey:
Thanks for reporting this to me btw.

PR #246, opened 2020-06-14
doc(common): Add examples and improve common docs

jan-auer:
Updates doc comments all over `symbolic-common` with doc comments and improved descriptions, and also deprecates some APIs that are not future-proof:

- `pointer_size`, `instruction_alignment` and `ip_register_name` have moved from `Arch` to `CpuFamily`.
- `Arch::register_name` as been moved to `CpuFamily::cfi_register_name`.
- `CpuFamily::cfi_register_name` returns `None` instead of `Some("")` for some unknown registers.
- Field access on `InstructionInfo` has been deprecated and replaced with a builder.

### DIFF
--- a/common/src/byteview.rs
+++ b/common/src/byteview.rs
@@ -50,16 +50,17 @@ impl Deref for ByteViewBacking<'_> {
 /// underlying file handle until the `ByteView` is dropped:
 ///
 /// ```
-/// # fn main() -> Result<(), std::io::Error> {
-/// # use std::io::Write;
+/// use std::io::Write;
 /// use symbolic_common::ByteView;
 ///
-/// # let mut file = tempfile::tempfile()?;
-/// # file.write_all(b"1234");
+/// fn main() -> Result<(), std::io::Error> {
+///     let mut file = tempfile::tempfile()?;
+///     file.write_all(b"1234");
 ///
-/// let view = ByteView::map_file(file)?;
-/// assert_eq!(view.as_slice(), b"1234");
-/// # Ok(()) }
+///     let view = ByteView::map_file(file)?;
+///     assert_eq!(view.as_slice(), b"1234");
+/// Ok(())
+/// }
 /// ```
 #[derive(Clone, Debug)]
 pub struct ByteView<'a> {
@@ -120,13 +121,14 @@ impl<'a> ByteView<'a> {
     /// # Example
     ///
     /// ```
-    /// # fn main() -> Result<(), std::io::Error> {
-    /// # use std::io::Write;
+    /// use std::io::Write;
     /// use symbolic_common::ByteView;
     ///
-    /// # let mut file = tempfile::tempfile()?;
-    /// let view = ByteView::map_file(file)?;
-    /// # Ok(()) }
+    /// fn main() -> Result<(), std::io::Error> {
+    ///     let mut file = tempfile::tempfile()?;
+    ///     let view = ByteView::map_file(file)?;
+    ///     Ok(())
+    /// }
     /// ```
     pub fn map_file(file: File) -> Result<Self, io::Error> {
         let backing = match unsafe { Mmap::map(&file) } {
@@ -157,13 +159,14 @@ impl<'a> ByteView<'a> {
     /// # Example
     ///
     /// ```
-    /// # fn main() -> Result<(), std::io::Error> {
     /// use std::io::Cursor;
     /// use symbolic_common::ByteView;
     ///
-    /// let reader = Cursor::new(b"1234");
-    /// let view = ByteView::read(reader)?;
-    /// # Ok(()) }
+    /// fn main() -> Result<(), std::io::Error> {
+    ///     let reader = Cursor::new(b"1234");
+    ///     let view = ByteView::read(reader)?;
+    ///     Ok(())
+    /// }
     /// ```
     ///
     /// [`open`]: struct.ByteView.html#method.open
@@ -180,11 +183,12 @@ impl<'a> ByteView<'a> {
     /// # Example
     ///
     /// ```no_run
-    /// # fn main() -> Result<(), std::io::Error> {
     /// use symbolic_common::ByteView;
     ///
-    /// let view = ByteView::open("test.txt")?;
-    /// # Ok(()) }
+    /// fn main() -> Result<(), std::io::Error> {
+    ///     let view = ByteView::open("test.txt")?;
+    ///     Ok(())
+    /// }
     /// ```
     pub fn open<P: AsRef<Path>>(path: P) -> Result<Self, io::Error> {
         let file = File::open(path)?;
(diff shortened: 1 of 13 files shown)


common/src/heuristics.rs
```diff
@@ -8,53 +12,241 @@ const SIGSEGV: u32 = 11;
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use symbolic_common::{Arch, InstructionInfo};
+    ///
+    /// let caller_address = InstructionInfo::new(Arch::X86, 0x1337)
+    ///     .caller_address();
+    /// ```
+    pub fn new(arch: Arch, instruction_address: u64) -> Self {
```
flub, line 144: When reading this in use `new` reads a little weird.  Though I'm struggling to come up with a better name, so I guess it's fine.

common/src/byteview.rs
```diff
@@ -101,22 +150,58 @@ impl<'a> ByteView<'a> {
...
         let mut buffer = vec![];
         reader.read_to_end(&mut buffer)?;
         Ok(ByteView::from_vec(buffer))
     }
 
     /// Constructs a `ByteView` from a file path by memory mapping the file.
+    ///
+    /// # Example
+    ///
+    /// ```no_run
```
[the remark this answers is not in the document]
jan-auer: I had that actually, and then thought it's sort of redundant to the other example. The actual reason I went with this is for the `"test.txt"` literal path. 

If you prefer me to change it?

common/src/cell.rs
```diff
@@ -219,8 +222,20 @@ where
 {
     /// Creates a new `SelfCell`.
     ///
-    /// The callback receives a pointer to the owned data. Note that a borrow to that data can only
-    /// safely be used to derive the object and **must not** leave the callback.
+    /// # Safety
```
Swatinem, line 225: I wonder if you can just guarantee this via lifetimes somehow?

common/src/path.rs
```diff
@@ -179,14 +224,38 @@ pub fn clean_path(path: &str) -> Cow<'_, str> {
...
+///
+/// **Note**: This is the same as [`split_path`], except that it operates on byte slices.
+///
+/// # Examples
+///
+/// Split the last component of a UNIX path:
+///
+/// ```
+/// assert_eq!(
+///     symbolic_common::split_path_bytes("/a/b/c".as_bytes()),
```
Swatinem, line 244:
```suggestion
///     symbolic_common::split_path_bytes(b"/a/b/c"),
```
this might be easier to read, for the other examples as well

common/src/heuristics.rs
```diff
@@ -8,53 +12,241 @@ const SIGSEGV: u32 = 11;
...
+    #[doc(hidden)]
+    #[deprecated = "Use InstructionInfo::new() instead"]
     pub arch: Arch,
-    /// This is true if the frame is the cause of the crash.
+    #[doc(hidden)]
+    #[deprecated = "Use is_crashing_frame() instead"]
     pub crashing_frame: bool,
-    /// If a signal is known that triggers the crash, it can be stored here.
+    #[doc(hidden)]
+    #[deprecated = "Use signal() instead"]
```
Swatinem, line 123: Do we need accessors for this? If so, `set_signal` might be better.

common/src/heuristics.rs
```diff
@@ -8,53 +12,241 @@ const SIGSEGV: u32 = 11;
...
+///   instruction. All ARM64 instructions are 4 bytes long; subtract 4 bytes from the return address
+///   to derive the likely address of the calling instruction.
+///
+/// # More Information
+///
+/// The above information was taken and slightly updated from the now-gone *PLCrashReporter Wiki*.
+/// An old copy can still be found in the [internet archive].
+///
+/// [internet archive]:
+/// https://web.archive.org/web/20161012225323/https://opensource.plausible.coop/wiki/display/PLCR/Automated+Crash+Report+Analysis
```
Swatinem, line 108: does this need to be on one line rather?

```suggestion
/// [internet archive]: https://web.archive.org/web/20161012225323/https://opensource.plausible.coop/wiki/display/PLCR/Automated+Crash+Report+Analysis
```

common/src/byteview.rs
```diff
@@ -101,22 +150,58 @@ impl<'a> ByteView<'a> {
 
     /// Constructs a `ByteView` from any `std::io::Reader`.
     ///
-    /// This currently consumes the entire reader and stores its data in an internal buffer. Prefer
-    /// `ByteView::open` when reading from the file system or `ByteView::from_slice` /
-    /// `ByteView::from_vec` for in-memory operations. This behavior might change in the future.
+    /// **Note**: This currently consumes the entire reader and stores its data in an internal
+    /// buffer. Prefer [`open`] when reading from the file system or [`from_slice`] / [`from_vec`]
+    /// for in-memory operations. This behavior might change in the future.
```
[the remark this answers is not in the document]
jan-auer: It's actually unlikely that this ever changes, that's why it's "might".

common/src/cell.rs
jan-auer: Sadly, not. Any sort of guarantee would require self referential types.

common/src/heuristics.rs
jan-auer: I'd say we probably don't, since the only use of this is to get the caller address and not to store this information.

common/src/heuristics.rs
jan-auer: Argh, my auto formatter constantly destroys these links.

common/src/types.rs
```diff
@@ -477,7 +710,19 @@ impl<'a> Name<'a> {
...
+    /// Constructs a new mangled name with a known [`Language`].
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use symbolic_common::{Language, Name};
+    ///
+    /// let name = Name::with_language("_ZN3foo3barEv", Language::Cpp);
+    /// println!("{:#}", name); // prints "_ZN3foo3barEv [C++]"
+    /// # assert_eq!(format!("{:#}", name), "_ZN3foo3barEv [C++]");
```
Swatinem, line 722: I would personally prefer to just have the `assert_eq` call shown there. Similarly for the `Language` and `Arch` examples.

common/src/byteview.rs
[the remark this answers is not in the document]
flub: In my experience examples which aren't run tend to be broken eventually, which is why i suggested this as it doesn't seem too hard or noisy.  but i don't feel strongly, it's just a suggestion.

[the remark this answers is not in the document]
jan-auer: See https://doc.rust-lang.org/stable/std/fs/struct.File.html#method.create

common/src/path.rs
jan-auer: Updated it where it's possible. Unfortunately `b""` is not a slice, it's an array.

common/src/heuristics.rs
jan-auer: Yeah, `new` is not very descriptive, but it seems to be the accepted name for such functions.

PR #250, opened 2020-06-24
C-ABI: Add method to create symbolic archive from bytes buffer

unclead:
It would be nice to have the method to create an archive from bytes buffer.

For instance, I have a zip archive with dSYM files and I have to unpack the zip archive to a temp directory, then call the method symbolic_archive_open for each file in order to get symbolic archive instance.

Instead of it, I want to just read the slice of bytes from the zip archive and create the symbolic archive instance from it without the necessity to save the buffer to a temporary file to open it further using current exposed methods

p.s. I've tried to generate heard file calling make file command but it failed due to rust compilation error, so I just added the new method to the file manually (I'm not good in rust to fix such errors)

### DIFF
--- a/cabi/include/symbolic.h
+++ b/cabi/include/symbolic.h
@@ -28,7 +28,7 @@ enum SymbolicErrorCode {
   SYMBOLIC_ERROR_CODE_OBJECT_ERROR_UNSUPPORTED_OBJECT = 2101,
   SYMBOLIC_ERROR_CODE_OBJECT_ERROR_BAD_BREAKPAD_OBJECT = 2102,
   SYMBOLIC_ERROR_CODE_OBJECT_ERROR_BAD_ELF_OBJECT = 2103,
-  SYMBOLIC_ERROR_CODE_OBJECT_ERROR_BAD_MACH_OOBJECT = 2104,
+  SYMBOLIC_ERROR_CODE_OBJECT_ERROR_BAD_MACH_O_OBJECT = 2104,
   SYMBOLIC_ERROR_CODE_OBJECT_ERROR_BAD_PDB_OBJECT = 2105,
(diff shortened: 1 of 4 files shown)


py/tests/test_debug.py
```diff
@@ -11,10 +11,18 @@
...
     archive = Archive.open(binary_path)
     obj = archive.get_object(arch="x86_64")
     assert obj.features == set(["symtab", "unwind"])
 
+    with open(binary_path, "rb") as f:
+        buf = f.read()
+
+    archive = Archive.from_bytes(buf)
+    obj = archive.get_object(arch="x86_64")
+    assert obj.features == set(["symtab", "unwind"])
```
jan-auer, line 24: Thanks for adding a test! Would you mind moving this into a separate test function to keep the concerns separated?

unclead: done

cabi/include/symbolic.h
```diff
@@ -28,7 +28,7 @@ enum SymbolicErrorCode {
   SYMBOLIC_ERROR_CODE_OBJECT_ERROR_UNSUPPORTED_OBJECT = 2101,
   SYMBOLIC_ERROR_CODE_OBJECT_ERROR_BAD_BREAKPAD_OBJECT = 2102,
   SYMBOLIC_ERROR_CODE_OBJECT_ERROR_BAD_ELF_OBJECT = 2103,
-  SYMBOLIC_ERROR_CODE_OBJECT_ERROR_BAD_MACH_OOBJECT = 2104,
+  SYMBOLIC_ERROR_CODE_OBJECT_ERROR_BAD_MACH_O_OBJECT = 2104,
```
Swatinem, line 31: heads up! I manually reverted this line when working on proguard

jan-auer: 👍  Thanks. I evaluated this again and thought it's OK to change this now. WDYT?

PR #252, opened 2020-06-29
feat(debuginfo): Expose binary names for ELF and MachO

calixteman:


debuginfo/src/elf.rs
```diff
@@ -81,6 +81,11 @@ impl<'d> ElfObject<'d> {
             .map(|slice| CodeId::from_binary(slice))
     }
 
+    /// The binary's soname, if any.
+    pub fn soname(&self) -> Option<&'d str> {
```
jan-auer, line 85: @calixteman What if we call this `name`, too? Since this is also called `name` on MachO and Breakpad, we could expose this via the `ObjectLike` trait later, too.

PR #255, opened 2020-07-20
fix: Update cpp_demangle to increase its recursion limit

Swatinem:
Requires https://github.com/gimli-rs/cpp_demangle/pull/207


jan-auer:
@Swatinem Thanks! Added a link to the PR in the description.

jan-auer:
The linter error is actually valid, and we can fix this in a separate PR.

Swatinem:
using #262 instead

PR #263, opened 2020-09-11
ref: Replace failure with thiserror in symbolic-unreal

Swatinem:


unreal/src/error.rs
```diff
@@ -1,36 +1,30 @@
...
     /// The crash file contains unexpected trailing data after the footer.
-    #[fail(display = "unexpected trailing data")]
+    #[error("unexpected trailing data")]
     TrailingData,
 
     /// Can't process a log entry.
-    #[fail(display = "invalid log entry")]
-    InvalidLogEntry(std::str::Utf8Error),
+    #[error("invalid log entry")]
+    InvalidLogEntry(#[from] std::str::Utf8Error),
```
jan-auer, line 25: This should not be `#[from]`, as the underlying error is too generic. I believe the same counts for `BadCompression`.

It's too easy where we add a new variant that also contains a `Utf8Error` and then we have to break the `From` impl.

jan-auer (changes requested):
We need to retain the `Fail` derive with a default feature flag. Otherwise, we need to release a breaking change, and for that we need a bit more legwork.

Swatinem:
Closing this for now. We either have to feature-flag this or make a breaking change release

Issue #259, opened 2020-08-28
RUSTSEC-2020-0036: failure is officially deprecated/unmaintained


extrawurst:
`cargo audit` spits out:

```
warning: 1 warning found

Crate:  failure
Title:  failure is officially deprecated/unmaintained
Date:   2020-05-02
URL:    https://rustsec.org/advisories/RUSTSEC-2020-0036
```

Rooted in `symbolic-common` using `failure`.
Did not check the code yet, but how deep is it entangled?

this bubbled up in my `gitui` project: https://github.com/extrawurst/gitui/issues/248

jan-auer:
Thanks! We are looking to remove `failure` in favor of `std::error::Error` soon. 

In the past, the only blocker for us has been missing stabilization of `Error::backtrace`, which we rely on heavily in our production services. However, we've recently come to the conclusion that libraries like `symbolic` should not capture backtraces on their own, so I believe that we can now start to replace `failure`.

extrawurst:
Any update on this?

Swatinem:
I’m starting to work on this, first in `symbolicator`, which has no semver requirements and can therefore internally break whatever it wants. After gaining some experience in this, we will try to tackle `symbolic`. The challenge here is to do it without breaking the API, for example by doing it behind a feature flag.

Swatinem:
`failure` was removed some time ago, and we don’t even have it in transitive dependencies anymore.

PR #264, opened 2020-09-17
ref: Replace failure with std::error::Error

jan-auer:
Ref https://github.com/getsentry/symbolic/issues/259


flub (approved):

jan-auer:
This is essentially ready for review, thanks a bunch to @Swatinem for wrapping this up. 

I'll also do another pass through all changes, so please don't merge just yet.

symbolic-cabi/src/core.rs
```diff
@@ -435,9 +438,11 @@ pub unsafe extern "C" fn symbolic_err_get_last_message() -> SymbolicStr {
     use std::fmt::Write;
     LAST_ERROR.with(|e| {
         if let Some(ref err) = *e.borrow() {
-            let mut msg = err.to_string();
-            for cause in err.iter_causes() {
+            let mut err = err.as_ref();
+            let mut msg = format!("Error: {}", err);
```
jan-auer, line 442: Let's just use the plain `.to_string()` here.

symbolic-cabi/src/core.rs
```diff
@@ -449,21 +454,7 @@ pub unsafe extern "C" fn symbolic_err_get_last_message() -> SymbolicStr {
...
-                write!(&mut out, "stacktrace: {}", backtrace).ok();
-                SymbolicStr::from_string(out)
-            } else {
-                Default::default()
-            }
-        } else {
-            Default::default()
-        }
-    })
+    LAST_ERROR.with(|_| Default::default())
```
jan-auer, line 457:
```suggestion
    SymbolicStr::default()
```

symbolic-debuginfo/src/elf.rs
```diff
@@ -4,10 +4,11 @@ use std::borrow::Cow;
 use std::fmt;
 use std::io::Cursor;
 
-use failure::Fail;
 use flate2::{Decompress, FlushDecompress};
 use goblin::elf::compression_header::{CompressionHeader, ELFCOMPRESS_ZLIB};
-use goblin::{container::Ctx, elf, error::Error as GoblinError, strtab};
+pub use goblin::error::Error as GoblinError;
```
jan-auer, line 9: _nit_: Lets move this under the imports as this is now a re-export. Either inherit docs or add a doc string.

symbolic-debuginfo/src/object.rs
```diff
@@ -60,40 +60,39 @@ macro_rules! map_result {
 
 /// An error when dealing with any kind of [`Object`](enum.Object.html).
 #[non_exhaustive]
-#[allow(clippy::large_enum_variant)]
-#[derive(Debug, Fail)]
+#[derive(Debug, Error)]
```
jan-auer, line 63: On most errors we started deriving clone, except on this one. I'm considering to remove all `Clone` implementations, but let's check if we need them in Symbolicator.

symbolic-debuginfo/src/pdb.rs
```diff
@@ -1066,7 +1046,7 @@ impl<'s> Unit<'s> {
                     let parent_offset = proc_offsets
                         .last()
                         .map(|&(_, offset)| offset)
-                        .ok_or_else(|| PdbError::from(PdbErrorKind::UnexpectedInline))?;
+                        .ok_or_else(|| PdbError::UnexpectedInline)?;
```
jan-auer, line 1049:
```suggestion
                        .ok_or(PdbError::UnexpectedInline)?;
```

symbolic-debuginfo/src/sourcebundle.rs
```diff
@@ -858,7 +865,7 @@ where
     pub fn write_object<O>(self, object: &O, object_name: &str) -> Result<bool, SourceBundleError>
     where
         O: ObjectLike,
-        O::Error: Fail,
+        O::Error: Into<ObjectError>,
```
jan-auer, line 868: I'm realizing that this is fairly restrictive now, essentially making it impossible that anyone outside of symbolic implements `ObjectLike` and then uses it with this class. I think we'll have to go back to `Box<dyn>` for this, or we simply swallow the cause.

Swatinem: note to self: The `O::Error` constraint should be `dyn Error + Send + Sync + 'static`, and in the places where we save it its boxed.

symbolic-debuginfo/src/object.rs
Swatinem: note to self: remove `Clone` on all the other custom error types…

jan-auer (commented):
Looks good, let's ship this.

Swatinem (approved):

PR #272, opened 2020-09-24
ref(common): Change InstructionInfo setters to Option

jan-auer:


symbolic-cabi/src/symcache.rs
```diff
@@ -199,16 +199,13 @@ ffi_fn! {
...
-            real_info.signal(info.signal);
-        }
-        if info.ip_reg != 0 {
-            real_info.ip_register_value(info.ip_reg);
-        }
+        let address = InstructionInfo::new(arch, info.addr)
+            .is_crashing_frame(info.crashing_frame)
+            .signal(Some(info.signal).filter(|&s| s != 0))
+            .ip_register_value(Some(info.ip_reg).filter(|&r| r != 0))
+            .caller_address();
```
jan-auer, line 206: This change shows how this can be used differently now. The `Some().filter()` pattern is probably not super nice, but it is easy to read here.

Swatinem (approved):
I wonder what learnings I can take over to the sentry SDK?

jan-auer:
> I wonder what learnings I can take over to the sentry SDK?

My learnings here were:
1. Avoid external branching if you expect options as input. That allows to write everything in one statement with chained methods. 
2. Use `&mut self` if the "final" action method takes `&mut self` or `&self`. If the final method consumes, then I'd also build the builder consuming.

PR #273, opened 2020-09-25
ci: Switch to GitHub Actions

Swatinem:


Swatinem:
https://zeus.ci/gh/getsentry/symbolic/builds/2042 looking good

scripts/docker-manylinux.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -ex
+
+BUILD_DIR="/work"
+
+docker run \
+        -w /work/py \
+        -v `pwd`:/work \
+        quay.io/pypa/manylinux1_${BUILD_ARCH} \
```
jan-auer: This needs to be manylinux2010

scripts/docker-manylinux.sh
```diff
@@ -0,0 +1,15 @@
...
+        sh manylinux.sh
+
+# Fix permissions for shared directories
+USER_ID=$(id -u)
+GROUP_ID=$(id -g)
+sudo chown -R ${USER_ID}:${GROUP_ID} target/
```
jan-auer, line 15: Can you check if we can remove this? It may only be a relic in Relay, since in current symbolic Travis builds we do not need to do this. 

Note that for release builds we do not cache the target folder in Relay.

.github/workflows/release.yml
```diff
@@ -0,0 +1,97 @@
...
+          npx zeus upload --build ${{ github.run_id }} --job ${{ github.job }} --type "application/zip+wheel" py/dist/*
+          npx zeus job update --build ${{ github.run_id }} --job ${{ github.job }} --status passed
+
+  python-wheel-linux:
+    strategy:
+      fail-fast: false
+      matrix:
+        build-arch: [i686, x86_64]
+
+    name: Create Python wheel for Linux ${{ matrix.build-arch }}
```
jan-auer, line 45: Can we give these jobs a shorter name? For example "Python Linux ${{ matrix.build-arch }}"

https://github.com/getsentry/relay/blob/e5c85a6b83b309f13d598b1e34816c13bbb3f970/.github/workflows/build_library.yml#L15

jan-auer (commented):
Partial review.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,119 @@
...
+  pull_request:
+
+jobs:
+  lint-rust:
+    name: Rust Style/Linting
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
```
jan-auer, line 17: I am a little curious how this can compile clippy without submodules. I'm guessing it may not run the build script, but can you confirm this?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,119 @@
...
+        os: [ubuntu-latest, macos-latest]
+
+    name: Rust Test on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+
+    env:
+      CXX: clang
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
```
jan-auer, line 56: Here we should definitely check out submodules.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,119 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout sources
+        uses: actions/checkout@v2
+
+      - name: Setup Python
+        uses: actions/setup-python@v2
+
+      - run: make style-python
+      - run: make lint-python
```
jan-auer, line 86: In Relay, we have all lints in a single job. Mind if we combine these?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,119 @@
...
+
+      - name: Install rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: stable
+          override: true
+
+      - name: Setup Python
+        uses: actions/setup-python@v2
```
jan-auer, line 113: Let's pin the python version to 3.7 (as minimum requirement). We can also test for latest in addition, if you like.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,119 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+
+    name: Python Test on ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+
+    env:
+      CXX: clang
```
jan-auer, line 98: Can you check if we still need this? I remember this was a requirement on Travis because the default C++ compiler was too old, but that may not be the case here.

jan-auer (approved):
Zeus build is here: https://zeus.ci/gh/getsentry/symbolic/builds/2047

All artifacts are available. Updated craft to expect a macOS 10.15 wheel.

jan-auer:
Updated required status checks.

PR #275, opened 2020-09-30
ref: Introduce unmangled Name and better DemangleOptions

Swatinem:
DemangleOptions now takes explicit flags with support for arguments and return types.


symbolic-demangle/src/lib.rs
```diff
@@ -48,38 +50,63 @@ extern "C" {
...
+    arguments: bool,
+    return_type: bool,
+}
+
+impl DemangleOptions {
+    /// Create new DemangleOptions defaulting to very simple formatting.
+    pub const fn new() -> Self {
+        Self {
+            arguments: false,
+            return_type: false,
```
jan-auer, line 92: I'm starting to think that we should probably print everything by default, and then have a shorthand to turn off most of the unwanted stuff. A really good example for this is [`DemangleFlags`](https://docs.rs/msvc-demangler/0.8.0/msvc_demangler/struct.DemangleFlags.html): There is a "shorthand" for complete and name only demangling.

What if we provide two base constructors: `DemangleOptions::complete` and `DemangleOptions::name_only`, and they turn everything to `true` and `false`, respectively. Maybe even no `Default` implementation.

FWIW, I'm not a big fan that `NAME_ONLY` is additive to `COMPLETE`, but that's just how `msvc_demangler` works.

symbolic-common/src/types.rs
```diff
@@ -769,6 +797,22 @@ impl<'a> Name<'a> {
...
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use symbolic_common::Name;
+    ///
+    /// let name = Name::unmangled("foo::bar");
+    /// assert_eq!(name.mangled(), false);
+    /// ```
+    pub fn mangled(&self) -> bool {
```
jan-auer, line 812: Just because the name is not `unmangled` does not necessarily mean it is mangled. Two alternatives:

- `potentially_mangled`
- `unmangled`

On a side note, "unmangled" reads a bit strange to me. What if we call it "demangled"?

jan-auer (changes requested):

symbolic-demangle/src/lib.rs
```diff
@@ -48,38 +50,63 @@ extern "C" {
...
-}
-
 /// Options for [`Demangle::demangle`].
 ///
+/// The resulting output depends very much on the language of the mangled [`Name`], and not all
+/// options are fully supported by each language.
+///
+/// The default will output a very simple demangling, which means only the function name in most
+/// cases.
+/// Turning on all the options will result in a fully demangled type.
```
flub, line 60: What's the motivation for leaving everything off by default?  Perhaps that could also be part of the documentation.  And why is there no simple way of enabling everything?

flub (approved):

symbolic-demangle/src/lib.rs
```diff
@@ -48,38 +50,63 @@ extern "C" {
...
+        }
+    }
 
     /// Determines whether function arguments should be demangled.
     ///
     /// Defaults to `false`.
-    pub with_arguments: bool,
-}
+    pub const fn arguments(mut self, arguments: bool) -> Self {
+        self.arguments = arguments;
```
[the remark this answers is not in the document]
Swatinem: Swift supports argument names, and argument types (+ return types) separately.

Swatinem:
I changed the `DemangleOptions` to have two constructors now, and the `Name` to have a tri-state for its mangling. Not quite sure about the naming for that however.

symbolic-common/src/types.rs
```diff
@@ -702,10 +730,36 @@ impl<'a> Name<'a> {
         Name {
             string: string.into(),
             lang: Language::Unknown,
+            mangling: Default::default(),
```
jan-auer, line 733: Let's be explicit in all places where we use `NameMangling::Unknown`, since this now has significant semantics.

```suggestion
            mangling: NameMangling::Unknown,
```

symbolic-common/src/types.rs
```diff
@@ -702,10 +730,36 @@ impl<'a> Name<'a> {
         Name {
```
jan-auer, line 730: Should we make the `new` function take a `mangling` argument instead? 

When constructing a name, we will know in most cases whether the name is mangled or not, particularly:
- Symbol tables are always mangled
- Dwarf can have both, but we know based on which attribute we read
- In PDBs it depends on whether we read it from the GSI or from the MDI.
- In PEs, symbols are always mangled.
- Breakpad symbols are never mangled

This means, we can explicitly invoke the constructor with the mangling and can encourage doing so by not offering a constructor that implicitly fills in unknown. 

If you prefer split constructors, you can also create three for `mangled`, `unmangled` and `unknown`, it seems a bit too much in this case.

symbolic-common/src/types.rs
```diff
@@ -769,6 +824,22 @@ impl<'a> Name<'a> {
         self.lang
     }
 
+    /// The Names mangling state.
```
jan-auer, line 827:
```suggestion
    /// Returns the `Name`'s mangling state.
```

symbolic-demangle/src/lib.rs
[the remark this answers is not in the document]
jan-auer: In this case, how do you feel about a second option for controlling argument names, that works just for Swift?

symbolic-demangle/src/lib.rs
```diff
@@ -113,16 +141,16 @@ fn is_maybe_swift(_ident: &str) -> bool {
...
-    };
+
+    // the flags are bitflags
+    let mut flags = MsvcFlags::COMPLETE;
+    if !opts.return_type {
+        flags |= MsvcFlags::NO_FUNCTION_RETURNS;
+    }
+    if !opts.arguments {
+        // a `NO_ARGUMENTS` flag is there in the code, but commented out
+        flags |= MsvcFlags::NAME_ONLY;
```
jan-auer, line 152: Will this mean that `arguments(false)` also disables the return type for MSVC?

symbolic-demangle/src/lib.rs
```diff
@@ -183,19 +214,16 @@ fn try_demangle_swift(ident: &str, opts: DemangleOptions) -> Option<String> {
...
-                1
-            } else {
-                2
-            }
-        }
-        DemangleFormat::Full => 0,
-    };
+    let mut features = 0;
+    if opts.return_type {
+        features |= 1 << 1;
```
jan-auer, line 219: Can we introduce shared constants for these to make this searchable across C++/Rust code?

symbolic-demangle/src/lib.rs
Swatinem: I think so, yes.

examples/addr2line/src/main.rs
```diff
@@ -93,10 +93,7 @@ fn execute(matches: &ArgMatches<'_>) -> Result<()> {
         if matches.is_present("functions") {
             if let Some(symbol) = symbol_map.lookup(addr) {
                 print_name(
-                    symbol
-                        .name
-                        .as_ref()
-                        .map(|n| Name::new(n.as_ref(), NameMangling::Unknown)),
+                    symbol.name.as_ref().map(|n| Name::from(n.as_ref())),
```
jan-auer, line 96: Symbol maps are known to be mangled.

symbolic-common/src/types.rs
```diff
@@ -804,37 +782,37 @@ impl<'a> Name<'a> {
...
+    /// Converts this name into a [`Cow`].
     ///
     /// # Example
     ///
     /// ```
-    /// use symbolic_common::{Name, NameMangling};
+    /// use symbolic_common::Name;
     ///
-    /// let name = Name::new("_ZN3foo3barEv", NameMangling::Unknown);
+    /// let name = Name::from("_ZN3foo3barEv");
```
jan-auer, line 801: In these examples, you know that the name is mangled. I would propose to encourage explicit usage of `NameMangling` here:

```suggestion
    /// let name = Name::new("_ZN3foo3barEv", NameMangling::Mangled, Language::Cpp);
```

symbolic-cabi/src/demangle.rs
```diff
@@ -23,10 +23,7 @@ ffi_fn! {
         lang: *const SymbolicStr,
     ) -> Result<SymbolicStr> {
         let name = get_name(ident, lang);
-        let demangled = name.try_demangle(DemangleOptions {
-            with_arguments: true,
-            format: DemangleFormat::Short,
-        });
+        let demangled = name.try_demangle(DemangleOptions::complete());
```
jan-auer: This should probably use `name_only` and then set `arguments(true)`.

symbolic-debuginfo/src/dwarf.rs
```diff
@@ -825,7 +825,11 @@ impl<'d, 'a> DwarfUnit<'d, 'a> {
             let function = Function {
                 address: function_address,
                 size: function_size,
-                name: Name::with_language(name.unwrap_or_default(), self.language),
+                name: Name::new(
+                    name.unwrap_or_default(),
+                    NameMangling::Unknown,
```
jan-auer, line 830: This we can infer by knowing where the name comes from. This will require a slightly larger refactor:

 - If the name comes from the symbol table, then it is `Mangled`
 - If the name comes from `DW_AT_linkage_name` or `DW_AT_MIPS_linkage_name`, it is `Mangled`
 - If the name comes from `DW_AT_name`, then it is `Unmangled`, to my recollection. This we have to verify again, though.

jan-auer (commented):
Could you also update `BreakpadFunctionIterator` to return `NameMangling::Unmangled`? Breakpad files never contain mangled symbols as per definition.

jan-auer (approved):

Swatinem:
talked with jan: rename `argument_types` to `parameters`, and merge it ;-)

PR #277, opened 2020-10-07
fix: Add a SymbolIterator and Lifetimes to ObjectLike trait

Swatinem:
This avoids both boxing, and unsafely transmuting the symbol iterators.

### DIFF
--- a/symbolic-minidump/src/cfi.rs
+++ b/symbolic-minidump/src/cfi.rs
@@ -131,9 +131,9 @@ struct UnwindInfo<U> {
 }
 
 impl<U> UnwindInfo<U> {
-    pub fn new<'d, O, R>(object: &O, addr: u64, mut section: U) -> Self
+    pub fn new<'d: 'o, 'o, O, R>(object: &O, addr: u64, mut section: U) -> Self
     where
-        O: ObjectLike<'d>,
+        O: ObjectLike<'d, 'o>,
         R: Reader,
         U: UnwindSectionExt<R>,
     {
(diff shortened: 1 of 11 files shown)


symbolic-minidump/src/cfi.rs
```diff
@@ -236,9 +236,9 @@ impl<W: Write> AsciiCfiWriter<W> {
         Ok(())
     }
 
-    fn process_dwarf<'o, O>(&mut self, object: &O) -> Result<(), CfiError>
+    fn process_dwarf<'d: 'o, 'o, O>(&mut self, object: &O) -> Result<(), CfiError>
```
jan-auer, line 239: Is the `'d: 'o` bound required?

jan-auer (approved):

symbolic-minidump/src/cfi.rs
Swatinem:
```
error[E0495]: cannot infer an appropriate lifetime for lifetime parameter `'d` due to conflicting requirements
   --> symbolic-minidump\src\cfi.rs:259:24
    |
259 |             let info = UnwindInfo::new(object, section.address, 
frame);
    |                        ^^^^^^^^^^^^^^^
    |
note: first, the lifetime cannot outlive the lifetime `'d` as defined 
on the method body at 239:22...
   --> symbolic-minidump\src\cfi.rs:239:22
    |
239 |     fn process_dwarf<'d, 'o, O>(&mut self, object: &O) -> Result<(), CfiError>
    |                      ^^
note: ...but the lifetime must also be valid for the lifetime `'o` as 
defined on the method body at 239:26...
   --> symbolic-minidump\src\cfi.rs:239:26
    |
239 |     fn process_dwarf<'d, 'o, O>(&mut self, object: &O) -> Result<(), CfiError>
    |                          ^^
note: ...so that the types are compatible
```
seems so, yes

PR #279, opened 2020-10-08
ref: Add File/FunctionIterator and lifetimes to DebugSession

Swatinem:
continuation of #277

### DIFF
--- a/symbolic-debuginfo/src/base.rs
+++ b/symbolic-debuginfo/src/base.rs
@@ -602,9 +602,15 @@ pub type DynIterator<'a, T> = Box<dyn Iterator<Item = T> + 'a>;
 ///  - Read headers of compilation units (compilands) to resolve cross-unit references.
 ///
 /// [`ObjectLike::debug_session`]: trait.ObjectLike.html#tymethod.debug_session
-pub trait DebugSession {
+pub trait DebugSession<'session> {
     /// The error returned when reading debug information fails.
     type Error;
+
+    /// An iterator over all functions in this debug file.
+    type FunctionIterator: Iterator<Item = Result<Function<'session>, Self::Error>>;
+
+    /// An iterator over all source files referenced by this debug file.
+    type FileIterator: Iterator<Item = Result<FileEntry<'session>, Self::Error>>;
 
     /// Returns an iterator over all functions in this debug file.
     ///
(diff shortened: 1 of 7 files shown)


symbolic-debuginfo/src/base.rs
```diff
@@ -602,21 +602,27 @@ pub type DynIterator<'a, T> = Box<dyn Iterator<Item = T> + 'a>;
 ///  - Read headers of compilation units (compilands) to resolve cross-unit references.
 ///
 /// [`ObjectLike::debug_session`]: trait.ObjectLike.html#tymethod.debug_session
-pub trait DebugSession {
+pub trait DebugSession<'session> {
```
jan-auer, line 605: Usually, we don't need the `'data` lifetime here, but do we need the `'object` one?

jan-auer: On taking a second look, I think we can get away without, since we can still bind it in the associated types on the concrete implementation. In that case, `'object: 'session` would need to be fulfilled.

Swatinem: Yes, I think that should be enough.

jan-auer: What if we do the same for the `ObjectLike` trait, to only have the `'object` lifetime there?

Swatinem: To return references tied to the objects lifetime?

jan-auer (approved):
Merged the clippy fix from master. This is g2g now.

symbolic-debuginfo/src/base.rs
jan-auer: Correct. This would give us the benefit of not having to clone in `PdbObject::symbols`, and I think that use cases where you need the `'data` life time but cannot hold the objects are very scarce in hindsight.

symbolic-debuginfo/src/breakpad.rs
```diff
@@ -1079,15 +1079,17 @@ impl<'d> BreakpadDebugSession<'d> {
     }
 }
 
-impl<'d> DebugSession for BreakpadDebugSession<'d> {
+impl<'data, 'session> DebugSession<'session> for BreakpadDebugSession<'data> {
     type Error = BreakpadError;
+    type FunctionIterator = BreakpadFunctionIterator<'session>;
+    type FileIterator = BreakpadFileIterator<'session>;
```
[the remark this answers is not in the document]
Swatinem: I tried in an earlier experiment. Apparently one or two of the concrete implementations returns data tied to the session. Maybe we can revisit at a later point.

PR #281, opened 2020-10-16
fix: Update breakpad to version which fixes debug_id

flub:
This version of breakpad should compute the ELF debug_identifier
correctly for big-endian minidumps processed in a little-endian
system.


jan-auer (approved):

Swatinem:
While you are at it, can you update the directory here:
https://github.com/getsentry/symbolic/blob/390e3c7817cacd1772b8e59d38d12cfa36c6c671/.gitmodules#L1
I don’t think it matters, but just to be consistent.

jan-auer (approved):
LGTM. Diffs are expected due to changes in the stack scanner. Particularly, it now skips frames with no module mapping more aggressively.

flub:
LGTM

PR #282, opened 2020-10-19
feat(demangle): Update swift demangler to 5.3

Swatinem:


symbolic-demangle/vendor/swift/update.py
```diff
@@ -104,7 +104,7 @@ def main():
 
     print("> Cleaning up previous import")
     for entry in os.listdir(vendor_dir):
-        if os.path.isdir(entry):
+        if os.path.isdir(entry) and not workspace_dir.startswith(os.path.join(vendor_dir, entry)):
```
Swatinem, line 107: guess what happened, lol…

jan-auer: Well, yeah. You shouldn't check out swift inside of this folder, but I guess there's no harm in doing so.

jan-auer (approved):

PR #283, opened 2020-10-19
meta: Bump all semver-major dependencies

Swatinem:


symbolic-unreal/Cargo.toml
```diff
@@ -27,7 +27,7 @@ serde = ["serde_", "chrono/serde"]
 
 [dependencies]
 anylog = "0.5.0"
-bytes = "0.4.12"
+bytes = "0.5.6"
```
jan-auer, line 30: This will make it incompatible with Relay, can you check?

Swatinem: yes, I think that needs updating too. Too bad I can’t use that git ref right now until we fix the breakpad submodule :-(

Swatinem: relay already has both `0.4` and `0.5`, and the parameter of `symbolic-unreal` is not used anywhere.

jan-auer: Thanks for verifying.

jan-auer (approved):

PR #286, opened 2020-11-11
ci: Use GHA instead of zeus

Swatinem:


.github/workflows/ci.yml
```diff
@@ -29,8 +29,6 @@ jobs:
           python-version: 3.7
 
       - uses: Swatinem/rust-cache@v1
-        with:
-          key: lints
```
jan-auer, line 33: What changed in the action so that the key is no longer required?

jan-auer (approved):

jan-auer:
Please run a test in a release branch before merging.

.github/workflows/ci.yml
Swatinem: It now uses the job name automatically ;-)

jan-auer: Is the job name defaulted if not set explicitly, like in this case?
Don't you have to upgrade to a newer version of `rust-cache` then?

Swatinem:
https://github.com/getsentry/symbolic/actions/runs/357922856 so there is the artifact, seems like all the python stuff in there

.github/workflows/ci.yml
Swatinem: github will just use the latest commit from the `v1` branch, and it *always* attached the job name to the cache, in *addition* to the manually provided key
* before: `v0-rust-test-python-test-python-1.47.0-x86_64-unknown-linux-gnu-18bf6b4f01a6-2277f239961d34e39a31`
* after: `v0-rust-test-python-1.47.0-x86_64-unknown-linux-gnu-18bf6b4f01a6-2277f239961d34e39a31`

PR #289, opened 2020-11-20
ci: Run rust nighlty on a schedule

flub:
This ensures we're aware of issues with newer rust compilers sooner.
We run clippy and tests in the same step since caching matters more
than speed of completion as this does not block PRs.

No explicit caching since the compiler changes every time anyway.

#skip-changelog

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -14,7 +14,7 @@ jobs:
       matrix:
         rust: [nightly, beta]
     steps:
-      - uses: actions/checkout@master
+      - uses: actions/checkout@v2
         with:
           submodules: recursive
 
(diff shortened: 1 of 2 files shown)


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,37 @@
+name: Nightly CI
+
+on:
+  schedule:
+    - cron: '11 7 * * 1,4'
+
+env:
+  RUSTFLAGS: -Dwarnings
```
jan-auer, line 8: The rest of the `ci.yml` test suite does not deny warnings. Either add it there, too, or remove it here, please.

jan-auer (changes requested):

jan-auer (approved):

PR #293, opened 2020-11-26
fix(debuginfo): Prefer DWARF names for Dart functions

jan-auer:
We generally trust the symbol table to contain more accurate names than DW_AT_name and other attributes on DIEs in DWARF debug information. Some C++ compilers have been known to output truncated or simplified names into these attributes, whereas the symbol table always contains accurately mangled names.

The Dart compiler is an exception to this case. Its name mangling appears to be lossy and there is no generally available demangler. Thus, we need to prefer the demangled DW_AT_name.

Example:

- Mangled Symbol: `Precompiled__pM_658260265_sEa__anonymous_closure__8632`
- Demangled Symbol: `_MyAppState.build.<anonymous closure>`

Obtained via `cargo run -p addr2line -- -fCi -e app.symbols 0x211b7c` from [this example](https://github.com/getsentry/sentry-dart/blob/6be1283348b5f303de38099669a9fe9edafd6716/flutter/example/lib/main.dart#L59).

---

Before:
image

After:
image

### DIFF
--- a/symbolic-debuginfo/src/dwarf.rs
+++ b/symbolic-debuginfo/src/dwarf.rs
@@ -10,7 +10,7 @@
 use std::borrow::Cow;
 use std::fmt;
 use std::marker::PhantomData;
-use std::ops::Deref;
+use std::ops::{Deref, RangeBounds};
 
 use fallible_iterator::FallibleIterator;
 use gimli::read::{AttributeValue, Range};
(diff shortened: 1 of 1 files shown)


symbolic-debuginfo/src/dwarf.rs
```diff
@@ -419,10 +424,20 @@ impl<'d, 'a> DwarfUnit<'d, 'a> {
...
         };
 
+        let producer = match entry.attr_value(constants::DW_AT_producer)? {
+            Some(AttributeValue::String(string)) => Some(string),
+            _ => None,
+        };
+
+        // Trust the symbol table more to contain accurate mangled names. However, since Dart's name
+        // mangling is lossy, we need to load the demangled name instead.
+        let prefer_dwarf_names = producer.as_deref() == Some(b"Dart VM");
```
jan-auer, line 434: This is part 1 of the change: Detecting the Dart producer.

mitsuhiko (approved):

symbolic-debuginfo/src/dwarf.rs
```diff
@@ -684,25 +717,15 @@ impl<'d, 'a> DwarfUnit<'d, 'a> {
...
-                    .symbol_map
-                    .lookup_range(function_address..function_end)
-                    .and_then(|symbol| {
-                        symbol
-                            .name
-                            .clone()
-                            .map(|n| Name::new(n, NameMangling::Mangled, self.language))
-                    })
-            } else {
+            let symbol_name = if self.prefer_dwarf_names || inline {
```
jan-auer, line 720: This is part 2 of the change: Skipping symbol table lookup if DWARF names are preferred.

PR #296, opened 2020-11-30
ref(symcache): Switch to error with kind

jan-auer:

### DIFF
--- a/symbolic-symcache/src/cache.rs
+++ b/symbolic-symcache/src/cache.rs
@@ -3,7 +3,7 @@ use std::fmt;
 use symbolic_common::{Arch, AsSelf, DebugId, Language, Name, NameMangling};
 
 use crate::error::SymCacheError;
-use crate::format;
+use crate::{format, SymCacheErrorKind};
 
 /// A platform independent symbolication cache.
 ///
@@ -267,7 +267,7 @@ impl<'a> SymCache<'a> {
             if let Some((line_addr, file_id, line)) = self.run_to_line(fun, addr)? {
                 // A missing file record indicates a bad symcache.
                 let file_record = read_file_record(self.data, self.header.files, file_id)?
-                    .ok_or(SymCacheError::BadCacheFile)?;
+                    .ok_or(SymCacheErrorKind::BadCacheFile)?;
 
                 // The address was found in the function's line records, so use
                 // it directly. This should is the default case for all valid
--- a/symbolic-symcache/src/cache.rs
+++ b/symbolic-symcache/src/cache.rs
@@ -2,8 +2,8 @@ use std::fmt;
 
 use symbolic_common::{Arch, AsSelf, DebugId, Language, Name, NameMangling};
 
-use crate::error::SymCacheError;
-use crate::{format, SymCacheErrorKind};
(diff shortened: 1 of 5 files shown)


symbolic-symcache/src/cache.rs
```diff
@@ -3,7 +3,7 @@ use std::fmt;
 use symbolic_common::{Arch, AsSelf, DebugId, Language, Name, NameMangling};
 
 use crate::error::SymCacheError;
-use crate::format;
+use crate::{format, SymCacheErrorKind};
```
Swatinem, line 6: can you move the `…Kind` import to the correct line please

symbolic-symcache/src/error.rs
```diff
@@ -1,3 +1,4 @@
+use std::error::Error;
```
Swatinem, line 1: we import two `Error` types now. Can we name `thiserror::Error` directly in the derive?

symbolic-symcache/src/format.rs
```diff
@@ -8,6 +8,7 @@ use std::marker::PhantomData;
 use symbolic_common::{DebugId, Uuid};
 
 use crate::error::SymCacheError;
+use crate::SymCacheErrorKind;
```
Swatinem, line 11:
```suggestion
use crate::error::{SymCacheError, SymCacheErrorKind};
```

symbolic-symcache/src/writer.rs
```diff
@@ -9,7 +9,7 @@ use symbolic_common::{Arch, DebugId, Language};
 use symbolic_debuginfo::{DebugSession, FileInfo, Function, LineInfo, ObjectLike, Symbol};
 
 use crate::error::{SymCacheError, ValueKind};
-use crate::format;
+use crate::{format, SymCacheErrorKind};
```
Swatinem, line 12: and here

symbolic-symcache/src/error.rs
```diff
@@ -26,55 +27,98 @@ impl fmt::Display for ValueKind {
...
+    }
+}
+
+/// An error returned when handling [`SymCache`](struct.SymCache.html).
+#[derive(Debug, Error)]
+#[error("{kind}")]
+pub struct SymCacheError {
+    kind: SymCacheErrorKind,
+    #[source]
+    source: Option<Box<dyn Error + Send + Sync + 'static>>,
```
Swatinem, line 96: `std::io::Error` basically does double-boxing, which means that `sizeof` is one pointer, whereas we are 2 pointers big because of alignment.

symbolic-symcache/src/error.rs
jan-auer: Since one is a derive and one a trait, this is fine. Compare this to `serde`. Since the derive applies to the very same error type, I'd leave it like this.

symbolic-symcache/src/error.rs
jan-auer: Discussed offline that this would indeed make the happy-path more efficient, but at the expense of more code. In this case, we conclude to keep the code base simpler.

Swatinem (approved):

PR #297, opened 2020-11-30
ref(unreal): Switch to error with kind

jan-auer:


symbolic-unreal/src/container.rs
```diff
@@ -12,6 +12,7 @@ use scroll::{ctx::TryFromCtx, Endian, Pread};
 use crate::context::Unreal4Context;
 use crate::error::Unreal4Error;
 use crate::logs::Unreal4LogEntry;
+use crate::Unreal4ErrorKind;
```
Swatinem, line 15:
```suggestion
use crate::error::{Unreal4Error, Unreal4ErrorKind};
use crate::logs::Unreal4LogEntry;
```

Swatinem (approved):

PR #299, opened 2020-11-30
ref(debuginfo): Switch to error with kind

jan-auer:

### DIFF
--- a/symbolic-debuginfo/src/dwarf.rs
+++ b/symbolic-debuginfo/src/dwarf.rs
@@ -8,6 +8,7 @@
 //! [`MachObject`]: ../macho/struct.MachObject.html
 
 use std::borrow::Cow;
+use std::error::Error;
 use std::fmt;
 use std::marker::PhantomData;
 use std::ops::{Deref, RangeBounds};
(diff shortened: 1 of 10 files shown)


Swatinem (approved):
Haven’t looked at the descriptions & docs in detail, high level looks very good, and getting rid of the details in `ObjectError` sounds good

symbolic-debuginfo/src/object.rs
```diff
@@ -49,58 +50,75 @@ macro_rules! map_inner {
...
+}
 
-    /// An error in DWARF debugging information.
-    #[error("failed to process dwarf info")]
-    Dwarf(#[from] DwarfError),
+impl fmt::Debug for ObjectError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self.repr {
+            ObjectErrorRepr::Transparent(ref inner) => inner.fmt(f),
+            _ => self.repr.fmt(f),
```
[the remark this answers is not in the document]
jan-auer: Good question. This always used to resolve correctly in the past, but it's probably worth disambiguating this.

symbolic-debuginfo/src/dwarf.rs
```diff
@@ -51,29 +52,76 @@ fn offset(addr: u64, offset: i64) -> u64 {
...
-    #[error("unexpected inline function without parent")]
     UnexpectedInline,
 
     /// The debug_ranges of a function are invalid.
-    #[error("function with inverted address range")]
     InvertedFunctionRange,
 
     /// The DWARF file is corrupted. See the cause for more information.
-    #[error("corrupted dwarf debug data")]
-    CorruptedData(#[from] GimliError),
```
Swatinem, line 76: `GimliError` was exposed because of the pub usage here. Maybe we can hide all the `gimli` re-exports again?

jan-auer: Good catch, I'll create a follow-up.

jan-auer: https://github.com/getsentry/symbolic/pull/300

PR #302, opened 2020-12-11
ci(gha): Build a wheel for macOS 11

### LINKED ISSUE #301: Python wheel for macOS Big Sur
Would be great to have a Python wheel for macOS 11.0 to avoid compiling from source. Thank you!

jan-auer:
On macOS Big Sur, python2 reports as `10.16` but python3 reports as `11.0`. For this reason, the `10.15` wheel does not install under python 3 and requires a separate wheel build.

Attempts to fix #301.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -5,6 +5,9 @@ on:
     branches:
       - "release/**"
 
+  # FIXME: Temporarily build pull requests.
+  pull_request:
+
 jobs:
   python-wheel-mac:
     strategy:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -5,10 +5,21 @@ on:
     branches:
       - "release/**"
 
+  # FIXME: Temporarily build pull requests.
+  pull_request:
```
flub, line 9: Don't forget to resolve this before merging

flub (approved):

.github/workflows/release.yml
jan-auer: Thank you, I would've surely forgot. :)

jan-auer:
Closing since with an up-to-date `pip` this is no longer required.

PR #305, opened 2020-12-23
fix(pytest): Pin pytest 5 for debugfail customizations

jan-auer:
With pytest 6, the internal representation of tests seems to have changed. Our overrides in `conftest.py` no longer emit compatible output which causes pytest to fail with an `INTERNALERROR`.

Until we get that fixed, this pins pytest to v5.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ test-rust:
 .PHONY: test-rust
 
 test-python: .venv/bin/python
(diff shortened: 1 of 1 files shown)


flub (approved):
It doesn't even show up on pytest's changelog.  Their API surface is so huge and uncharted...

PR #308, opened 2021-01-05
ci(gha): Switch to release workflow

jan-auer:
Switches the release process to the centralized release action. Craft is run directly on the host and requires `cargo` in it's bump-version script.

To match conventions from other repositories, the former release build workflow has been renamed to "Release build" and moved to `build.yml`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,19 +1,30 @@
 name: Release
 
 on:
-  push:
-    branches:
-      - "release/**"
+  workflow_dispatch:
+    inputs:
+      version:
+        description: Version to release
+        required: true
+      force:
+        description: Force a release even when there are release-blockers (optional)
+        required: false
 
 jobs:
-  python-wheel-mac:
-    name: Python macOS
-    runs-on: macos-10.15
-
+  release:
+    runs-on: ubuntu-latest
+    name: "Release a new version"
     steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -1,79 +1,51 @@
...
-      - run: make sdist
-
-      - uses: actions/upload-artifact@v2
-        with:
-          name: ${{ github.sha }}
-          path: py/dist/*
+      - name: Craft Prepare
+        run: npx @sentry/craft prepare --no-input "${{ env.RELEASE_VERSION }}"
+        env:
+          GITHUB_API_TOKEN: ${{ github.token }}
```
Swatinem, line 38: So `RELEASE_VERSION` is defined via the prepare-release action?
Also, why are you using `github.token` here, and `GH_RELEASE_PAT` otherwise, whats the difference?

Swatinem (approved):

.github/workflows/release.yml
jan-auer: github.token is able to push to this repository, but `GH_RELEASE_PAT` is able to operate on the `getsentry/publish` repo.

jan-auer:
@BYK only difference is that `getsentry/action-prepare-release` is pinned to the current `main` branch SHA.

PR #314, opened 2021-01-17
build: Update gimli to 0.23.0

jan-auer:


jan-auer:
Tests fail with an error from upstream gimli:

```
---- test_elf_executable stdout ----
Error: ElfError { source: Some(Malformed("Section 122 size (4198160) + addr (312) is out of bounds. Overflowed: false")) }
```

jan-auer:
Reverted goblin due to a regression in 0.3.0

flub (approved):

PR #316, opened 2021-01-28
fix(debuginfo): Detect more hidden symbols in swift sources

jan-auer:
A regression in https://github.com/getsentry/symbolic/pull/127 broke detection of BCSymbolMaps. The RegEx `__?hidden#\d+_` was replaced with a literal `__?hidden#` prefix and missed that `?` is an operator rather than a literal question mark.

This RegEx was required due to another issue in the MachO symbol iterator, which unconditionally strips the first leading underscore to match names from DWARF debugging information. For hidden symbols, this does not make sense, however, and instead it should be retained.


symbolic-debuginfo/src/macho.rs
```diff
@@ -252,8 +252,13 @@ impl<'d> MachObject<'d> {
     ///
     /// This is an indication that BCSymbolMaps are needed to symbolicate crash reports correctly.
     pub fn requires_symbolmap(&self) -> bool {
-        self.symbols()
-            .any(|s| s.name().map_or(false, |n| n.starts_with("__?hidden#")))
+        self.symbols().any(|s| {
+            s.name().map_or(false, |n| {
+                n.starts_with("__?hidden#")
+                    || n.starts_with("__hidden#")
+                    || n.starts_with("_hidden#")
```
marandaneto, line 259: wouldn't it be better to be future proof and regex it?
`hidden#` looks like the common pattern, or even taking into consideration 1 or N `_` and having or not `?`

Swatinem (approved):
lgtm, @marandaneto has a valid feedback though. Not sure how likely it is that things are changing again

flub (approved):

symbolic-debuginfo/src/macho.rs
jan-auer: Regexes are a bit expensive at this part, but we could use aho corasick in place of this.

That said, MachO dictates that there can only be a single or two underscores, and I'm not even sure why the symbols I'm looking at don't have the question mark.

marandaneto: wfm then, thx for explaining

marandaneto (approved):

jan-auer:
Verified this on a new build. Turns out that this was a regression in https://github.com/getsentry/symbolic/pull/127, where I replaced the RegEx `__?hidden#\\d+_` with a literal `__?hidden#` prefix and missed that `?` is an operator rather than a literal questionmark.

PR #319, opened 2021-02-02
fix(symcache): Compute correct line offsets

jan-auer:
SymCaches store line records where record specifies the delta to the previous line record. These deltas are stored as `u8` to save space for the common case where only small increments are needed. If two records are more than `0xff` apart, we insert filler line records.

This PR solves two bugs in creating these filler line records:
1. The delta was computed with `(diff & 0xff) as u8` instead of `diff.min(0xff) as u8`. This means that there was an address drift every time we had to insert filler line records.
2. The filler line records are technically still part of the previous range. However, they already carried information of the next line record, leading to wrong line numbers.

**SymCache version is incremented to `6`.** 

The previous implementation originates affe7b9dd119416bfaa6a07d928ea464b00c4bab and 5209ebd7e47d73a273fca3c14978386f63bbad17 and was subsequently amended with split functions in #155.

### DIFF
--- a/symbolic-symcache/src/format.rs
+++ b/symbolic-symcache/src/format.rs
@@ -13,7 +13,7 @@ use crate::error::{SymCacheError, SymCacheErrorKind};
 pub const SYMCACHE_MAGIC: [u8; 4] = *b"SYMC";
 
 /// The latest version of the file format.
-pub const SYMCACHE_VERSION: u32 = 5;
+pub const SYMCACHE_VERSION: u32 = 6;
 
 // Version history:
 //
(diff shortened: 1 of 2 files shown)


jan-auer:
The second part of the fix will not work if we are running into `should_split_function`, since we're exiting the current scope and losing the `last_*` variables. However in such a case the line records are irrelevant. They are filling a large gap (>65k) where other functions will have better matching line records.

I'm trying to check if we can instead store the size of the line record.

Swatinem (approved):

Swatinem (approved):

PR #326, opened 2021-02-17
fix(debuginfo): Support debug_addr indexes in DWARF functions

jan-auer:
The values of the `DW_AT_low_pc` and `DW_AT_high_pc` attributes can also contain
`DW_FORM_addrx*`. We did not support this and used to raise an unreachable
panic.

To handle this more gracefully, symbolic will now return an error. This error
matches what gimli's `Dwarf::attr_addr` returns in the same case.

Note that this will complicate debugging. The `DwarfError` struct does not carry
a backtrace yet, so we will no longer know which attribute causes the error.
Once backtraces are stable in the standard library, we can reintroduce this with
a feature flag.

### DIFF
--- a/symbolic-debuginfo/src/dwarf.rs
+++ b/symbolic-debuginfo/src/dwarf.rs
@@ -516,20 +516,26 @@ impl<'d, 'a> DwarfUnit<'d, 'a> {
             match attr.name() {
                 constants::DW_AT_low_pc => match attr.value() {
                     AttributeValue::Addr(addr) => low_pc = Some(addr),
-                    _ => unreachable!(),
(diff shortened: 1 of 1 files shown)


loewenheim (approved):

jan-auer:
Note, I didn't use [`Dwarf::attr_address`](https://docs.rs/gimli/0.23.0/gimli/read/struct.Dwarf.html#method.attr_address) directly as this would require us to match twice in the `high_pc` case. Gimli itself does this in a strange way here: 

https://github.com/gimli-rs/gimli/blob/36d1e2d41472cc48fe41df839c2059ba551fb329/src/read/dwarf.rs#L342-L350

flub (approved):

PR #336, opened 2021-03-08
Initial support for BCSymbolMap

flub:
There is one TODO left to decide, but otherwise this is pretty much done.


symbolic-debuginfo/src/plist.rs
```diff
@@ -0,0 +1,155 @@
...
+    pub fn test(bytes: &[u8]) -> bool {
+        bytes.starts_with(b"<?xml")
+    }
+
+    /// Parse the plist file, creating a new in-memory representation of it.
+    pub fn parse(id: DebugId, data: &[u8]) -> Result<Self, PListError> {
+        let mut map = HashMap::new();
+        let plist = Element::from_reader(Cursor::new(data))?;
+
+        {
```
jan-auer, line 96: Is this scope still needed?

symbolic-debuginfo/src/plist.rs
```diff
@@ -0,0 +1,155 @@
...
+/// Apple PropertyList structure.
+///
+/// This is an auxiliary file, not an object file.
+///
+/// The PList format is used to map the UUID of a dSYM compiled from bitcode to the original
+/// UUID of the matching [`BCSymbolMap`].
+#[derive(Clone, Debug)]
+pub struct PList {
+    id: DebugId,
+    map: HashMap<String, String>,
```
jan-auer, line 82: Since there's little we can do with `PList`s other than use them for dSYMs, we can simply read the UUID and other relevant attributes, but discard the rest. Particularly, this allows us to provide an infallible signature for `PList::original_uuid`.

symbolic-debuginfo/src/macho.rs
```diff
@@ -1,7 +1,7 @@
 //! Support for Mach Objects, used on macOS and iOS.
 
 use std::borrow::Cow;
-use std::error::Error;
+use std::error::Error as StdError;
```
jan-auer, line 4: Is there a particular reason for this change? It would be best to keep `std::error::Error` without alias and qualify other errors that might conflict, instead.

symbolic-debuginfo/src/macho.rs
```diff
@@ -41,6 +42,7 @@ impl MachError {
 pub struct MachObject<'d> {
     macho: mach::MachO<'d>,
     data: &'d [u8],
+    bcsymbolmap: Option<BCSymbolMap>,
```
jan-auer, line 45: As discussed offline, this should be supplied to the debug session, instead. Contrary to `ObjectLike`, which provides a more-or-less cheap view on the file, `DebugSession` is meant for keeping state and performing lookups and resolution.

This also means that we need to add a symbol iterator to the debug session, which in all cases other than MachO is the same as `*Object::symbols`.

symbolic-debuginfo/src/plist.rs
```diff
@@ -0,0 +1,155 @@
...
+
+use elementtree::Element;
+use thiserror::Error;
+
+use symbolic_common::{DebugId, ParseDebugIdError};
+
+/// The error type for handling a [`PList`].
+#[derive(Debug, Error)]
+#[error("{kind}")]
+pub struct PListError {
```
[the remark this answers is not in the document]
flub: it's kind of because plist is more general i put it here.  it doesn't actually try to reject plists not related to dsyms, only has an `.is_bcsymbol_mapping()` to determine if it can be used.

I imagine when we bump into plist support issues we'd move to an external crate for plist parsing.  Which means maybe this should not end up in the public API like this.  I think moving it to the new bcsymbolmap module might make more sense.  Maybe a more simple ‘OriginalUUIDMapping` with `::parse()`, `::test()` and `::original_uuid()` methods as that hides more of the implementation?

symbolic-debuginfo/src/bcsymbolmap.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+}
+
+/// An in-memory representation of the BCSymbolMap.
+///
+/// This is an auxiliary file, not an object file.
+///
+/// This can be used to provide symbols to a [`MachO`](crate::macho::MachO) object.
+///
+/// TODO(flub): Make this not own the data.
```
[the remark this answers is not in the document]
flub: I'm not sure about lazy parsing, is it so bad to build a `Vec<(usize, &str)>` (or something to that effect) at parse time?

symbolic-debuginfo/src/macho.rs
flub: it's a fairly common pattern i've seen when using anyhow and/or thiserror.  but i don't mind much

symbolic-debuginfo/src/bcsymbolmap.rs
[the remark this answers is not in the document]
jan-auer: Only if we want to do a few number of lookups. Since we now need the entire symbolmap anyway, let's keep it this way 👍

symbolic-debuginfo/src/macho.rs
```diff
@@ -206,6 +257,7 @@ impl<'d> MachObject<'d> {
             symbols: self.macho.symbols(),
             sections,
             vmaddr: self.load_address(),
+            symbolmap: self.bcsymbolmap.clone(), // TODO(flub): Don't clone
```
flub, line 260: We'll have to keep doing this because this iterator does not depend on the lifetime of the object itself.  The symbol map is now only a vector of pointers, so it's not so bad, but it is kind of sad.

symbolic-debuginfo/src/bcsymbolmap.rs
```diff
@@ -0,0 +1,215 @@
...
+/// so that even the DWARF files do not have the actual symbol names.  In this case the
+/// build process will create a `.bcsymbolmap` file which maps the obfuscated symbol names
+/// back to the original ones.  This structure can parse these files and allows providing
+/// this information to the [`MachObject`] so that it has the original symbol names instead
+/// of `__hidden#NNN_` ones.
+///
+/// [`MachObject`]: crate::macho::MachObject
+#[derive(Clone, Debug)]
+pub struct BCSymbolMap<'d> {
+    id: DebugId, // TODO
```
flub, line 67: decide this

symbolic-debuginfo/src/macho.rs
flub: from offline discussion: we have some ideas about what this api would look like but would like to have it validated with other auxiliary files.  the api introduced here is forwards compatible with such an api on the DebugSession so we'll go with this minimal version for now while we build up experience.

symbolic-debuginfo/src/macho.rs
```diff
@@ -65,15 +65,49 @@ impl<'d> MachObject<'d> {
...
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use symbolic_debuginfo::bcsymbolmap::BCSymbolMap;
+    /// use symbolic_debuginfo::macho::MachObject;
+    ///
+    /// let object_data =
+    ///     std::fs::read("tests/fixtures/2d10c42f-591d-3265-b147-78ba0868073f.dwarf-hidden")
+    ///         .unwrap();
```
jan-auer, line 84: This example is great! It could be confusing to see the "test/fixtures..." path in a doc comment, though, so what if we hide that line (and same for the symbolmap below)?

```suggestion
    /// # let object_data =
    /// #     std::fs::read("tests/fixtures/2d10c42f-591d-3265-b147-78ba0868073f.dwarf-hidden")
    /// #         .unwrap();
```

flub: Hmm, it would make the `object_data` variable come out of nowhere with no idea on what its type would be.  I prefer this as it's fully self contained, I don't think it's a big leap to infer that's just a path name.  Plus anyone who knows doctests, which will be most people reading this, will know exactly what's going on.

flub: I could be convinced to do something ugly like:
```
/// let object_data = std::fs::read("some/path/object.dwarf");
/// # let object_data = <the full thing that actually works>
```
I guess, if you really prefer.  But I don't see it as much better to be honest.

symbolic-debuginfo/src/bcsymbolmap.rs
jan-auer: It looks like there's not really a benefit to having it, so we can as well get rid of it here.

symbolic-debuginfo/src/bcsymbolmap.rs
```diff
@@ -0,0 +1,248 @@
...
+    /// Returns the name of a symbol if it exists in this mapping.
+    pub fn get(&self, index: usize) -> Option<&'d str> {
+        self.names.get(index).copied()
+    }
+
+    /// Resolves a name using this mapping.
+    ///
+    /// If the name matches the `__hidden#NNN_` pattern that indicates a [`BCSymbolMap`]
+    /// lookup it will be looked up the resolved name will be returned.  Otherwise the name
+    /// is returned unchanged.
```
jan-auer, line 128: _nit_: Bonus points for giving the lookup methods an illustrative `# Example`.

symbolic-debuginfo/src/lib.rs
```diff
@@ -39,6 +39,7 @@ mod base;
 mod object;
 mod private;
 
+pub mod bcsymbolmap;
```
jan-auer, line 42: Since BCSymbolMaps are tied to `MachO`, could we just move it over? That way you don't have to import from two modules when dealing with those files, and they will all be documented together. We could also add a mention to them in the `macho` module doc comment.

symbolic-debuginfo/src/macho.rs
jan-auer: Turns out, in the standard library they annotate such examples with `norun` and simply ensure it compiles, since there will be tests covering this anyway. Given that this is quite an odd path to read, we could do that, too? 

See [`std::fs::File` source](https://doc.rust-lang.org/stable/src/std/fs.rs.html#65-78).

jan-auer (approved):

symbolic-debuginfo/src/bcsymbolmap.rs
```diff
@@ -0,0 +1,248 @@
...
+    }
+}
+
+impl<'d> BCSymbolMap<'d> {
+    /// Tests whether the buffer could contain a [`BCSymbolMap`].
+    pub fn test(bytes: &[u8]) -> bool {
+        let mut pattern = BC_SYMBOL_MAP_HEADER.as_bytes();
+        if pattern.len() > bytes.len() {
+            pattern = &pattern[..bytes.len()];
+        }
```
Swatinem, line 88: This means that a truncated file would pass this check. Is this really what we want? Or is this a limitation that `test` does get called with a small slice of the file only?

symbolic-debuginfo/src/bcsymbolmap.rs
```diff
@@ -0,0 +1,248 @@
...
+            .next()
+            .ok_or(BCSymbolMapErrorKind::InvalidHeader)?;
+        if header != BC_SYMBOL_MAP_HEADER {
+            return Err(BCSymbolMapErrorKind::InvalidHeader.into());
+        }
+
+        let mut names = Vec::new();
+        for line in lines_iter {
+            names.push(line);
+        }
```
Swatinem, line 114:
```suggestion
        let names = lines_iter.collect();
```

symbolic-debuginfo/src/bcsymbolmap.rs
```diff
@@ -0,0 +1,248 @@
...
+
+        let name = map_iter.next();
+        assert_eq!(name.unwrap(), "-[SentryMessage initWithFormatted:]");
+
+        let name = map_iter.next();
+        assert_eq!(name.unwrap(), "-[SentryMessage setMessage:]");
+    }
+
+    #[test]
+    fn test_data_lifetime() {
```
Swatinem, line 221: nice!

symbolic-debuginfo/tests/fixtures/c8374b6d-6e96-34d8-ae38-efaa5fec424f.bcsymbolmap
```diff
@@ -0,0 +1,3660 @@
...
+sentrycrashsignal_fatalSignals
+sentrycrashsignal_signalCodeName
+sentrycrashsignal_signalName
+_handleSignal
+_sentrycrashcm_signal_getAPI.api
+_g_eventID
+_g_signalStack
+_g_previousSignalHandlers
+/Users/philipphofmann/git-repos/sentry-cocoa/Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_Signal.c
+Sources/SentryCrash/Recording/Monitors/SentryCrashMonitor_Signal.c
```
Swatinem, line 3656: I’m a bit surprised to find (absolute) filenames here. Are those being obfuscated as well? Do we need to resolve them when we extract file/line info?

symbolic-debuginfo/src/macho.rs
```diff
@@ -206,6 +257,7 @@ impl<'d> MachObject<'d> {
             symbols: self.macho.symbols(),
             sections,
             vmaddr: self.load_address(),
+            symbolmap: self.bcsymbolmap.clone(),
```
Swatinem, line 260: I can’t remember if you mentioned this to me in the walkthrough, not sure if its worth avoiding the clone here?

Swatinem (approved):
👍🏻 This looks really straight forward so far.

Can you double check my question regarding filenames?

symbolic-debuginfo/tests/fixtures/c8374b6d-6e96-34d8-ae38-efaa5fec424f.bcsymbolmap
flub: huh, intersting.  i must admit i did not check this file...  Yes, I imagine that this is also obfuscated. I guess question is whether we need to de-obfuscate other things than just symbols as well I guess.  We should look at this sometime.

symbolic-debuginfo/src/bcsymbolmap.rs
jan-auer: Good point, actually all you want to do is `bytes.starts_with(pattern)`, right?

symbolic-debuginfo/src/bcsymbolmap.rs
flub: `ObjectLike::debug_id()` identifies the object.  It is nice to be able to identify a thing, if not someone who has this need has to create their own wrapper, which admittedly is trivial.  It turned out not to be any burden to require this when creating the `BCSymbolMap`, and one could even imagine a constructor like `from_path()` (though because some users might want async and some not, it's probably not so useful).

Anyway, I can't decide which one I prefer and you seem to think without so let's do without.

symbolic-debuginfo/src/bcsymbolmap.rs
flub: Yes, the assumption is that this only gets called with a small slice.  This is kind of mimicking what `ObjectLike::test` does, this test is sufficient to decide between "is this an object? (`Archive::test`)", "is this a BCSymbolMap (`BCSymbolMap::test`)" or "is this a PList? (`PList::test`)".  So in a way the API assumption here is very weak and test is only useful in relation to the other test functions.
On the other hand to call `::parse()`, which is the only thing that actually gives you a full answer to this question, you need to read all the data from disk.  So this is useful as a quick check if it's even worth reading the entire file from disk.  Again this only makes sense if a BCSymbolMap is regularly larger than the filesystem block size, but it seems that is the case.
Anyway, TL;DR is that this is symmetric with `ObjectLike` and makes sense when using it in some of the workflows we do.

symbolic-debuginfo/src/macho.rs
flub: I would like to keep the example as a test-case too.  I've gone with some compromise where I fake-read the data from a path like it appears in an .xcarchive in a comment while hiding where we actually load it from.

symbolic-debuginfo/src/macho.rs
flub: it would be somewhat hard to do effectively.  And in practice users (and symbolicator too) create a DebugSession which for MachO will always construct the entire symbol table anyway so there's no savings as this entire vector would be allocated during that operation anyway.

jan-auer: `Arc` it? That's what we do in PDB's case.

flub: that is an excellent idea.  so obvious it should have been done from the start :)

PR #350, opened 2021-03-24
Add CI to check doc comments

flub:
Plus fix a few existing errors with the doc comments.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,6 @@ on:
 
 env:
   RUSTFLAGS: -Dwarnings
-  RUSTDOCFLAGS: -Dwarnings
 
 jobs:
   lints:
@@ -53,6 +52,8 @@ jobs:
   doc-comments:
     name: Rust doc comments
     runs-on: ubuntu-latest
+    env:
+      RUSTDOCFLAGS: -Dwarnings
     steps:
       - uses: actions/checkout@v2
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -9,6 +9,7 @@ on:
 
 env:
   RUSTFLAGS: -Dwarnings
+  RUSTDOCFLAGS: -Dwarnings
```
jan-auer, line 12: Just noticed that we're inconsistent between Symbolicator and [Relay](https://github.com/getsentry/relay/blob/d6682dd9fcf5e509b84ead60c8652bb40de65cef/.github/workflows/deploy.yml#L16-L17), the latter of which uses:

```
    env:
      RUSTDOCFLAGS: -Dbroken_intra_doc_links
```

I think to remember that `-Dwarnings` did not work, can you confirm it does? 

In both cases, we have it on the job, so let's add it there, please.

flub: seems to fail just fine: https://github.com/getsentry/symbolic/pull/350/checks?check_run_id=2184271390#step:5:27

I think with `-Dbroken_intra_doc_links` you only turn that specific error from a warning into an error, while with `-Dwarnings` you turn all rustdoc warnings into errors.  As it works with all, I'd stick with that really.

jan-auer: Excellent, thanks! I'll update Relay then.

PR #351, opened 2021-03-25
ref(compiler): Upgrade to Rust 1.51

flub:
Fix some clippy warnings and disable some that we can't or want
to fix, because they're from a third party or would break the API.

The Into -> From changes are possible since the orphan rules have
been relaxed in Rust 1.41.  So we no longer work on older rust
versions.

Also bring the periodic CI job to the same level as the symbolicator
one.  Allowing to run it on demand and continue tests when one
fails.

#skip-changelog

### DIFF
--- a/symbolic-debuginfo/src/breakpad.rs
+++ b/symbolic-debuginfo/src/breakpad.rs
@@ -15,6 +15,7 @@ use crate::base::*;
 use crate::private::{Lines, Parse};
 
 mod parser {
+    #![allow(clippy::upper_case_acronyms)]
     use pest_derive::Parser;
 
     #[derive(Debug, Parser)]
--- a/.github/workflows/periodic.yml
+++ b/.github/workflows/periodic.yml
@@ -3,6 +3,7 @@ name: Nightly CI
 on:
   schedule:
     - cron: "11 7 * * 1,4"
+  workflow_dispatch:
 
 env:
   RUSTFLAGS: -Dwarnings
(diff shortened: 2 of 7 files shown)


Swatinem (approved):
Clippy is complaining still

flub:
> Clippy is complaining still

oh i know.  i've been trying to figure out why i don't get the failure notifications of the nightly CI job that's been failing since forever.  I don't get it though, I'm watching this repo and have email notifications for failed action runs set on but can't find a single email regarding the nightly CI failures

flub:
PTAL, this has changed quite a bit since the first version :)

symbolic-debuginfo/src/breakpad.rs
```diff
@@ -15,6 +15,7 @@ use crate::base::*;
 use crate::private::{Lines, Parse};
 
 mod parser {
+    #![allow(clippy::upper_case_acronyms)]
```
jan-auer, line 18: This looks like we might want to file that against `pest`.

jan-auer (approved):

Swatinem (approved):

flub:
Ok, one more round on this because while I was writing the commit message I realised that this bumps the minimal Rust version to 1.41.  Are we fine with that?  Do we document this anywhere?

jan-auer:
Good point, let's add this to the main `symbolic` docstring and regenerate the readme.

PR #356, opened 2021-03-31
feat(debuginfo): Add back PList parsing

flub:
But the type this time round is a simple mapping of UUIDs, the plist
is never representable itself so stays out of the API and allows us to
change the implementation.

#skip-changelog

### DIFF
--- a/symbolic-debuginfo/src/macho/bcsymbolmap.rs
+++ b/symbolic-debuginfo/src/macho/bcsymbolmap.rs
@@ -367,7 +367,7 @@ fn uuid_from_plist(data: &[u8]) -> Result<DebugId, PListError> {
     for element in dict.children() {
         if element.tag().name() == "key" && element.text() == "DBGOriginalUUID" {
             found_key = true;
-        } else if found_key == true {
+        } else if found_key {
             raw_original = Some(element.text().to_string());
             break;
         }
--- a/symbolic-debuginfo/src/macho/bcsymbolmap.rs
+++ b/symbolic-debuginfo/src/macho/bcsymbolmap.rs
@@ -191,6 +191,7 @@ impl<'a, 'd> Iterator for BcSymbolMapIterator<'a, 'd> {
 
 impl FusedIterator for BcSymbolMapIterator<'_, '_> {}
 
+/// Error type when parsing an Apple PropertyList into [`BitcodeUuidMapping`].
 #[derive(Debug, Error)]
 #[error("{kind}")]
 pub struct PListError {
@@ -199,10 +200,17 @@ pub struct PListError {
     source: Option<Box<dyn Error + Send + Sync + 'static>>,
 }
 
+impl PListError {
+    /// The kind of error, giving a little more detail.
+    pub fn kind(&self) -> PListErrorKind {
+        self.kind
+    }
+}
+
 impl From<elementtree::Error> for PListError {
     fn from(source: elementtree::Error) -> Self {
         Self {
-            kind: PListErrorKind::Xml,
+            kind: PListErrorKind::Parse,
             source: Some(Box::new(source)),
         }
     }
@@ -227,11 +235,11 @@ impl From<ParseDebugIdError> for PListError {
 #[non_exhaustive]
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum PListErrorKind {
-    /// The plist did not have the expected XML schema.
+    /// The plist did not have the expected (XML) schema.
     Schema,
-    /// There was an XML parsing error.
-    Xml,
-    /// Failed to parse a PList value.
+    /// There was an (XML) parsing error parsing the plist.
+    Parse,
+    /// Failed to parse a required PList value.
     ParseValue,
     /// Failed to parse UUID from filename.
     ParseFilename,
@@ -241,7 +249,7 @@ impl fmt::Display for PListErrorKind {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
             Self::Schema => write!(f, "XML structure did not match expected schema"),
-            Self::Xml => write!(f, "Invalid XML"),
+            Self::Parse => write!(f, "Invalid XML"),
             Self::ParseValue => write!(f, "Failed to parse a value into the right type"),
             Self::ParseFilename => write!(f, "Failed to parse UUID from filename"),
         }
--- a/symbolic-debuginfo/src/macho/bcsymbolmap.rs
+++ b/symbolic-debuginfo/src/macho/bcsymbolmap.rs
@@ -191,7 +191,7 @@ impl<'a, 'd> Iterator for BcSymbolMapIterator<'a, 'd> {
 
 impl FusedIterator for BcSymbolMapIterator<'_, '_> {}
 
-/// Error type when parsing an Apple PropertyList into [`UuidMapping`].
+/// Error for handling or creating a [`UuidMapping`].
 #[derive(Debug, Error)]
 #[error("{kind}")]
 pub struct UuidMappingError {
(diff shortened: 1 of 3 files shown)


symbolic-debuginfo/src/macho/bcsymbolmap.rs
```diff
@@ -187,6 +191,194 @@ impl<'a, 'd> Iterator for BcSymbolMapIterator<'a, 'd> {
...
+            kind: PListErrorKind::ParseValue,
+            source: Some(Box::new(source)),
+        }
+    }
+}
+
+/// Error kind for [`PListError`].
+#[non_exhaustive]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum PListErrorKind {
```
jan-auer, line 229: This kind is public, but it is not part of any public signature. Since it is only used for the `Display` implementation, it would be better to make this private.

Same applies to the other kind enumeration in this file.

symbolic-debuginfo/src/macho/bcsymbolmap.rs
```diff
@@ -187,6 +191,194 @@ impl<'a, 'd> Iterator for BcSymbolMapIterator<'a, 'd> {
 
 impl FusedIterator for BcSymbolMapIterator<'_, '_> {}
 
+#[derive(Debug, Error)]
+#[error("{kind}")]
+pub struct PListError {
```
jan-auer, line 196: This is called `PListError` but is used for parsing a `BitcodeUuidMapping`. Is this intentional?

Also, this error would benefit from a doc comment.

symbolic-debuginfo/src/macho/mod.rs
```diff
@@ -18,7 +18,7 @@ use crate::private::{MonoArchive, MonoArchiveObjects, Parse};
 
 mod bcsymbolmap;
 
-pub use bcsymbolmap::{BcSymbolMap, BcSymbolMapError};
+pub use bcsymbolmap::{BcSymbolMap, BcSymbolMapError, BitcodeUuidMapping};
```
jan-auer, line 21: Please make this a star-reexport. It's too easy to miss types here, for example here you forgot the `PListError` which is in the return type of `BitcodeUuidMapping::parse_plist`.

```suggestion
pub use bcsymbolmap::*;
```

jan-auer (changes requested):
Thanks, it definitely makes sense to add PList support back into symbolic in this way. 

This is now another utility we have to handle Apple's dSYM files, just like the `DSymPathExt` trait. I'm starting to wonder if it would make sense to move those into a `symbolic-dsym` crate.

symbolic-debuginfo/src/macho/bcsymbolmap.rs
```diff
@@ -187,6 +191,194 @@ impl<'a, 'd> Iterator for BcSymbolMapIterator<'a, 'd> {
...
+    /// use std::path::Path;
+    /// use symbolic_common::DebugId;
+    /// use symbolic_debuginfo::macho::BitcodeUuidMapping;
+    ///
+    /// let filename = Path::new("2d10c42f-591d-3265-b147-78ba0868073f.plist");
+    /// # let filename = Path::new("tests/fixtures/2d10c42f-591d-3265-b147-78ba0868073f.plist");
+    /// let dsym_uuid: DebugId = filename
+    ///     .file_stem().unwrap()
+    ///     .to_str().unwrap()
+    ///     .parse().unwrap();
```
Swatinem, line 282: hm, maybe the doctests should use `?` rather than `unwrap`? So its more realistic when you want to copy-paste this verbatim

symbolic-debuginfo/src/macho/bcsymbolmap.rs
```diff
@@ -187,6 +191,194 @@ impl<'a, 'd> Iterator for BcSymbolMapIterator<'a, 'd> {
...
+            found_key = true;
+        } else if found_key {
+            raw_original = Some(element.text().to_string());
+            break;
+        }
+    }
+
+    match raw_original {
+        Some(raw) => raw.parse().map_err(Into::into),
+        None => Err(PListError::from(PListErrorKind::Schema)),
```
Swatinem, line 378: Just a stylistic suggestion:
You can define an inner fn that returns `Option<&str>`, use short-circuit `?` inside, and then call that with one `ok_or_else` instead of doing so in basically 3-4 places above.

symbolic-debuginfo/src/macho/bcsymbolmap.rs
flub: So I finally looked at this and I think the actual mistake is not exposing a `PListError::kind() -> PListErrorKind` method.  See also https://github.com/getsentry/symbolic/pull/357

symbolic-debuginfo/src/macho/bcsymbolmap.rs
flub: Yes, kind of.  It's only returned from the `.parse_plist*` methods and the error kinds are primarily dealing with plists.  Making it a `BitcodeUuidMappingError` and `BitcodeUuidMappingErrorKind` was also just too wieldy plus it is weird to jam plist errors in such a generic name.  Unless we make a `BitcodeUuidMappingError` which only provides a source but does not try to provide a kind.  Which could also be an option.

symbolic-debuginfo/src/macho/bcsymbolmap.rs
flub: 🤷   people are meant to copy-paste?  in this case you'd use the other api provided instead of copy-pasting this anyway.  let me know if you feel strongly on this, but unless there's a strict doctest guide somewhere on which to chose i don't really mind enough on my own to change this :)

symbolic-debuginfo/src/macho/bcsymbolmap.rs
jan-auer: Then let's call it a `DSymPlist` instead of a `BitcodeUuidMapping`, and the associated error `DSymPlistError`.

symbolic-debuginfo/src/macho/bcsymbolmap.rs
jan-auer: You can have both, see [Examples on `File`](https://doc.rust-lang.org/stable/std/fs/struct.File.html#examples). 

Doc examples are primarily meant to be illustrative and idiomatic examples of how to use this in real scenarios. Since `.unwrap()` is a code smell, we should also not be documenting it. It only feels so natural because we overuse `unwrap` in unit tests to get nice stack traces.

symbolic-debuginfo/src/macho/bcsymbolmap.rs
```diff
@@ -194,6 +198,201 @@ impl<'a, 'd> Iterator for BcSymbolMapIterator<'a, 'd> {
...
+#[derive(Debug, Error)]
+#[error("{kind}")]
+pub struct PListError {
+    kind: PListErrorKind,
+    #[source]
+    source: Option<Box<dyn Error + Send + Sync + 'static>>,
+}
+
+impl PListError {
+    /// The kind of error, giving a little more detail.
```
jan-auer, line 211: Standard documentation for this is:

```rust
/// Returns the corresponding [`PListErrorKind`] for this error.
```

Inspired from `std::io::Error::kind` and used consistently throughout symbolic.

jan-auer (changes requested):
Thanks for the updates, approach and implementation look good.

See comments on naming and doc tests before merging.

symbolic-debuginfo/src/macho/bcsymbolmap.rs
flub: oh yes, doc examples are certainly idiomatic: https://doc.rust-lang.org/stable/std/primitive.str.html#examples-46

in all seriousness, yes we should aim to do better (and this already is!) but this is awkward as we have multiple error types to get rid of and we don't have anyhow in the dependencies.  i don't really mind unwrap calls in examples, the examples are not about teaching error handling...

flub: not to mention the more infrastructure you pile into the hidden part of the doc example the harder it becomes to maintain, especially once the nesting no longer matches.

jan-auer (approved):
Good to go, thanks 👍 

Would be great to see the doc examples without `unwrap` some time later.

PR #357, opened 2021-04-01
fix(bcsymbolmap): Expose error kind

flub:
The erorr kind was erroneously not exposed on the public API.

#skip-changelog


flub:
Another option here is to not expose the error kind at all and only provide a `source` of the error on the public API.  We'd then be using the kind only internally for display purposes.

jan-auer (approved):

PR #371, opened 2021-04-26
fix(cbindgen): Fix test, update generated header and add CI

flub:
This updates the cbindgen generated header with a new version as the
output has changed somewhat significantly.

It also fixes a test and adds some missing tests to CI.


Swatinem (approved):
This is built in docker for use by python, right? Does that still work?

flub:
I don't see why this should have affected that.  I didn't change any building/linking stuff.  The thing I'm struggling with was just never done on CI before.

PR #374, opened 2021-04-28
fix: Handle too many files more gracefully

Swatinem:
Instead of failing hard when hitting the max file limit, we rather emit
a sentinel for every file that exceeds the limit.

I tested this on https://symbols.electronjs.org/Electron%20Framework/8DE7618E965231A7B4CDD1C701077B9B0/Electron%20Framework.sym which previously ran into this limit and failed symcache conversion.

### DIFF
--- a/symbolic-symcache/src/writer.rs
+++ b/symbolic-symcache/src/writer.rs
@@ -381,8 +381,10 @@ where
             return Ok(*index);
         }
 
-        if self.files.len() >= std::u16::MAX as usize {
-            return Err(SymCacheErrorKind::TooManyValues(ValueKind::File).into());
+        // TODO: Instead of failing hard when exceeding the maximum allowed number of files, we rather
+        // emit `u16::MAX` which is already treated as a sentinel value for unknown file entries.
+        if self.files.len() >= u16::MAX as usize {
+            return Ok(u16::MAX);
         }
 
         let index = self.files.len() as u16;
--- a/symbolic-symcache/src/cache.rs
+++ b/symbolic-symcache/src/cache.rs
@@ -2,7 +2,7 @@ use std::fmt;
 
 use symbolic_common::{Arch, AsSelf, DebugId, Language, Name, NameMangling};
 
-use crate::error::{SymCacheError, SymCacheErrorKind};
+use crate::error::SymCacheError;
 use crate::format;
 
 /// A platform independent symbolication cache.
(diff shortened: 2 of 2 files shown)


symbolic-symcache/src/writer.rs
```diff
@@ -381,8 +381,17 @@ where
             return Ok(*index);
         }
 
+        // Instead of failing hard when exceeding the maximum allowed number of files (u16::MAX),
+        // we rather insert a sentinel "<unknown>" file and use that for all files that exceed the max.
+        const MAX_FILE_INDEX: u16 = std::u16::MAX - 1;
         if self.files.len() >= std::u16::MAX as usize {
-            return Err(SymCacheErrorKind::TooManyValues(ValueKind::File).into());
+            return Ok(MAX_FILE_INDEX);
```
flub, line 388: shouldn't this return `std::u16::MAX`?  I think now you're returning the index of the last real entry, instead of the one with `<unknown>`
(btw, should you do `u16::MAX` these days?  i thought clippy would be unhappy but maybe it doesn't matter)

Swatinem: lol, I was stupid, yes. being at `MAX` means we add the sentinel and return it, being over MAX means we always return the sentinel index (which is MAX)

symbolic-symcache/src/writer.rs
```diff
@@ -381,8 +381,16 @@ where
             return Ok(*index);
         }
 
-        if self.files.len() >= std::u16::MAX as usize {
-            return Err(SymCacheErrorKind::TooManyValues(ValueKind::File).into());
+        // Instead of failing hard when exceeding the maximum allowed number of files (u16::MAX),
+        // we rather insert a sentinel "<unknown>" file and use that for all files that exceed the max.
+        if self.files.len() > u16::MAX as usize {
```
flub, line 386: in this version the final size of `self.files.len()` will be `u16::MAX + 1` will it not?  Does it not have to stay inside `u16`?

flub (commented):
So this took me also a good 5-10mins of staring at it at first, thinkging "this can't be right" but not seeing the issue 🤣 

Anyway, this kind of looks like a good candidate for a unit test as that could easily verify the invariants that are not enforced in the type system and that we're now trying to get right by squinting hard enough.  I haven't looked at how much infrastructure is required exactly, but do you think it's reasonable to add?

symbolic-symcache/src/writer.rs
```diff
@@ -381,8 +381,16 @@ where
             return Ok(*index);
         }
 
-        if self.files.len() >= std::u16::MAX as usize {
-            return Err(SymCacheErrorKind::TooManyValues(ValueKind::File).into());
+        // Instead of failing hard when exceeding the maximum allowed number of files (u16::MAX),
+        // we rather insert a sentinel "<unknown>" file and use that for all files that exceed the max.
```
jan-auer, line 385: Please also mark this as `TODO`, since we now clearly need that we need to support more than 65k files.

symbolic-symcache/src/writer.rs
jan-auer: ⬆️  This version will create `SymCacheErrorKind::TooManyValues` when writing the files segment. Also for safety, please ensure that `self.files` never contains more than `u16::MAX` entries. This shows intent more clearly and is more robust.

symbolic-symcache/src/writer.rs
```diff
@@ -381,8 +381,16 @@ where
...
+        if self.files.len() > u16::MAX as usize {
+            return Ok(u16::MAX);
+        }
+        if self.files.len() == u16::MAX as usize {
+            record = format::FileRecord {
```
jan-auer, line 390: You overwrite the record here, and down below insert it into `file_cache`. This creates an invalid cache entry. You need to cache the original record, and only make the push into `self.files` conditional.

jan-auer (changes requested):

symbolic-symcache/src/writer.rs
Swatinem: I’m surprised how difficult it is to think through. Lets consider u8 since thats easier. `MAX` is 255 and `.len()` is 256, because of 0-based indexing, `MAX` is the largest *index* that you can use.

Swatinem:
Okay, I didn’t realize that `u16::MAX` was already treated as a sentinel value, however when reading that value later on in the pipeline, it would still fail claiming an invalid symcache file. Now I will emit the sentinel value for every file that exceeds the limit, and later on when resolving, it will just output an empty filename.

flub (approved):

jan-auer (approved):

PR #388, opened 2021-06-01
fix: Assume `CompactUnwindOp::None` are stackless functions

Swatinem:
We have seen some of these `CompactUnwindOp::None` correspond to some tiny
stackless functions, such as `__kill` from `libsystem_kernel.dylib` or similar.

Because they don't have a normal CFI record, we would fall back to frame pointers
or stack scanning when unwinding, which will cause us to skip the caller
frame, or fail unwinding completely.

To overcome this problem we will emit a CFI record that basically says that
the function has no stack space of its own. Since these compact unwind records
can be the result of merging multiple of these adjacent functions, they can
span more instructions/bytes than one single symbol.

This can potentially lead to false positives. However in that case, the unwinding
code will detect the bogus return address and fall back to frame pointers or
scanning either way.

### DIFF
--- a/symbolic-minidump/src/cfi.rs
+++ b/symbolic-minidump/src/cfi.rs
@@ -349,7 +349,7 @@ impl<W: Write> AsciiCfiWriter<W> {
                 let frame = EhFrame::new(&section.data, endian);
                 UnwindInfo::new(object, section.address, frame)
             });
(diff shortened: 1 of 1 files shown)


loewenheim (approved):

Gankra:
Hmm, neat!

PR #389, opened 2021-06-01
fix: Hardcode unwind info for _sigtramp

Swatinem:
This is on top of #388 

This specific function has some hand crafted dwarf expressions
that we currently can't process. They encode how to restore the
registers from a machine context accessible via `$rbx`

See: https://github.com/apple/darwin-libplatform/blob/215b09856ab5765b7462a91be7076183076600df/src/setjmp/x86_64/_sigtramp.s#L198-L258


loewenheim (approved):
I assume the additional lines in the test data are these hand-written special case entries?

Swatinem:
> I assume the additional lines in the test data are these hand-written special case entries?

Those are from #388. This PR does not add any lines to the fixtures.

PR #390, opened 2021-06-04
fix: Update msvc-demangler

Swatinem:
This update finally includes our patches and avoids the need to pull it from a git dependency.


jan-auer (approved):
🎉 

Please double-check integration tests in `symbolicator`, as they will need further updates in Sentry.

PR #391, opened 2021-06-04
ref: Update gimli

Swatinem:
Updates gimli to 0.24 and removes the public re-export. To do that we copy-pasted our own Endian enum.

*technically breaking*


jan-auer (commented):
What's the reason for removing the `gimli::RunTimeEndian` re-export? From what I can see, that type is as good as ours and to the user of `symbolic` the origin is opaque.

jan-auer (approved):

PR #399, opened 2021-06-09
fix: Introduce CfiCache versioning

Swatinem:
Uses a simple ASCII preamble.
One thing to note is that this header is being stripped when doing `as_slice` as to not pass it
through to breakpad.

### DIFF
--- a/symbolic-minidump/src/cfi.rs
+++ b/symbolic-minidump/src/cfi.rs
@@ -43,7 +43,12 @@ use symbolic_debuginfo::pe::{PeObject, RuntimeFunction, UnwindOperation};
 use symbolic_debuginfo::{Object, ObjectError, ObjectLike};
 
 /// The magic file preamble to identify cficache files.
-/// Files with version < 2 do not have a magic file preamble.
+///
+/// Files with version < 2 do not have the full preamble with magic+version, but rather start
+/// straight away with a `STACK` record.
+/// The magic here is a `u32` corresponding to the big-endian `CFIC`.
+/// It will be written and read using native endianness, so mismatches between writer/reader will
+/// result in a [`CfiErrorKind::BadFileMagic`] error.
 pub const CFICACHE_MAGIC: u32 = u32::from_be_bytes(*b"CFIC");
 
 /// The latest version of the file format.
--- a/symbolic-minidump/src/cfi.rs
+++ b/symbolic-minidump/src/cfi.rs
@@ -41,8 +41,22 @@ use symbolic_debuginfo::pdb::PdbObject;
 use symbolic_debuginfo::pe::{PeObject, RuntimeFunction, UnwindOperation};
 use symbolic_debuginfo::{Object, ObjectError, ObjectLike};
 
+/// The magic file preamble to identify cficache files.
+/// Files with version < 2 do not have a magic file preamble.
+pub const CFICACHE_MAGIC: [u8; 4] = *b"CFIC";
+
 /// The latest version of the file format.
-pub const CFICACHE_LATEST_VERSION: u32 = 1;
+pub const CFICACHE_LATEST_VERSION: u32 = 2;
+
+// The preamble are 8 bytes, all ASCII.
+// This should be forward-compatible with a binary format:
+// `CFICxxx\n`
+//      ^^^ <- 3 ASCII digits denoting the version, `002` for now.
+
+// Version history:
+//
+// 1: Initial ASCII-only implementation
+// 2: Implementation with a versioned preamble
 
 /// Used to detect empty runtime function entries in PEs.
 const EMPTY_FUNCTION: RuntimeFunction = RuntimeFunction {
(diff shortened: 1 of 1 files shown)


loewenheim (approved):

symbolic-minidump/src/cfi.rs
```diff
@@ -41,8 +41,22 @@ use symbolic_debuginfo::pdb::PdbObject;
...
+pub const CFICACHE_MAGIC: [u8; 4] = *b"CFIC";
+
 /// The latest version of the file format.
-pub const CFICACHE_LATEST_VERSION: u32 = 1;
+pub const CFICACHE_LATEST_VERSION: u32 = 2;
+
+// The preamble are 8 bytes, all ASCII.
+// This should be forward-compatible with a binary format:
+// `CFICxxx\n`
+//      ^^^ <- 3 ASCII digits denoting the version, `002` for now.
```
jan-auer, line 54: Can we make this binary right away, just like the symcache? This time around, it would also be great to fix the byte order rather than relying on native endianness.

symbolic-minidump/src/cfi.rs
```diff
@@ -1025,8 +1042,17 @@ impl<'a> CfiCache<'a> {
     /// Load a symcache from a `ByteView`.
     pub fn from_bytes(byteview: ByteView<'a>) -> Result<Self, CfiError> {
         if byteview.len() == 0 || byteview.starts_with(b"STACK") {
-            let inner = CfiCacheInner::V1(CfiCacheV1 { byteview });
+            let inner = CfiCacheInner::Unversioned(CfiCacheV1 { byteview });
             return Ok(CfiCache { inner });
+        } else if byteview.starts_with(&CFICACHE_MAGIC) && byteview.get(7).cloned() == Some(b'\n') {
```
jan-auer, line 1047: It would be better to (1) read the magic, (2) the version, and then (3) the rest based on the version. Comparing with the SymCache format, its header contains further binary data without a newline, for instance.

jan-auer (commented):
Note that if we roll this out, there is no more way to roll back. Once symbolicator has written the new CFI caches, they will produce errors when read from an old instance.

symbolic-minidump/src/cfi.rs
```diff
@@ -41,9 +42,23 @@ use symbolic_debuginfo::pdb::PdbObject;
 use symbolic_debuginfo::pe::{PeObject, RuntimeFunction, UnwindOperation};
 use symbolic_debuginfo::{Object, ObjectError, ObjectLike};
 
+/// The magic file preamble to identify cficache files.
+/// Files with version < 2 do not have a magic file preamble.
```
jan-auer, line 46: They don't have the `CFIC` preamble, but you validate version 1 files by checking that they start with a breakpad `STACK ` record.

Separately, can we declare that this preamble is subject to endianness and document how it behaves? Especially, that we read and write with native endianness.

symbolic-minidump/src/cfi.rs
```diff
@@ -1014,28 +1030,50 @@ pub struct CfiCache<'a> {
...
+
         let byteview = ByteView::from_vec(buffer);
-        let inner = CfiCacheInner::V1(CfiCacheV1 { byteview });
+        //let inner = CfiCacheInner::Versioned(CFICACHE_LATEST_VERSION, CfiCacheV1 { byteview });
+        let inner = CfiCacheInner::Unversioned(CfiCacheV1 { byteview });
         Ok(CfiCache { inner })
     }
 }
 
+fn write_preamble<W: Write>(mut writer: W, version: u32) -> Result<(), io::Error> {
```
jan-auer, line 1047: Do you know why we're not receiving a warning about dead code here?

jan-auer (approved):

flub (commented):
afaik this now produces a different magic depending if you read or write this file on little or big endian machine.  But the contents is all plain ascii and not affected by endianness.  Does it not make more sense to encode the magic and version in network-endianness?

symbolic-minidump/src/cfi.rs
flub: it's used on line 1096

jan-auer:
> afaik this now produces a different magic depending if you read or write this file on little or big endian machine. But the contents is all plain ascii and not affected by endianness. Does it not make more sense to encode the magic and version in network-endianness?

This is another good reason why we should document intent for the magic, @Swatinem.

@flub We will want to use the same header going forward when we introduce an all-binary CFI cache. However, Arpad and I concluded that it will be much easier to treat it as a host-local cache, just like symcaches, with native endianness data inside. To avoid misuse in case someone moves the files between hosts with different endianness, we encode the magic endianness-aware.

symbolic-minidump/src/cfi.rs
jan-auer: Thanks :)

Swatinem: It is still used when writing versioned caches, just that we never *create* such versioned caches.

PR #401, opened 2021-06-10
fix: Skip over duplicated functions in SymCaches

Swatinem:
The `DwarfDebugSession::functions` Iterator can yield duplicated functions, which, combined with a ton of inlinees can blow out the `ParentOffset` limit.

This does fix the symptom of these duplicated functions, but I would say that the real fix would rather be somewhere in the Iterator that yields these values.

### DIFF
--- a/symbolic-debuginfo/src/dwarf.rs
+++ b/symbolic-debuginfo/src/dwarf.rs
@@ -8,6 +8,7 @@
 //! [`MachObject`]: ../macho/struct.MachObject.html
 
 use std::borrow::Cow;
+use std::collections::BTreeSet;
 use std::error::Error;
 use std::fmt;
 use std::marker::PhantomData;
@@ -1249,6 +1250,7 @@ impl<'data> DwarfDebugSession<'data> {
             functions: Vec::new().into_iter(),
             range_buf: Vec::new(),
             finished: false,
+            seen_ranges: BTreeSet::new(),
         }
     }
 
@@ -1342,6 +1344,7 @@ impl<'s> Iterator for DwarfFileIterator<'s> {
 pub struct DwarfFunctionIterator<'s> {
     units: DwarfUnitIterator<'s>,
     functions: std::vec::IntoIter<Function<'s>>,
+    seen_ranges: BTreeSet<(u64, u64)>,
     range_buf: Vec<Range>,
     finished: bool,
 }
@@ -1356,6 +1359,18 @@ impl<'s> Iterator for DwarfFunctionIterator<'s> {
 
         loop {
             if let Some(func) = self.functions.next() {
+                // We have seen this iterator yield duplicate top-level function entries, which combined
+                // with recursively walking its inlinees can blow past symcache limits.
+                // We suspect the reason is that the iterator does not internally deduplicate top level
+                // functions defined in different compile units. We suspect this might be caused by link-time
+                // deduplication which merges templated code that is being generated multiple times in each
+                // compilation unit.
+                let fn_range = (func.address, func.size);
+                if self.seen_ranges.contains(&fn_range) {
+                    continue;
+                }
+                self.seen_ranges.insert(fn_range);
+
                 return Some(Ok(func));
             }
 
--- a/symbolic-debuginfo/src/dwarf.rs
+++ b/symbolic-debuginfo/src/dwarf.rs
@@ -1266,8 +1266,8 @@ impl<'data> DwarfDebugSession<'data> {
             units: self.cell.get().units(),
             functions: Vec::new().into_iter(),
             range_buf: Vec::new(),
-            finished: false,
             seen_ranges: BTreeSet::new(),
+            finished: false,
         }
     }
 
@@ -1361,8 +1361,8 @@ impl<'s> Iterator for DwarfFileIterator<'s> {
 pub struct DwarfFunctionIterator<'s> {
     units: DwarfUnitIterator<'s>,
     functions: std::vec::IntoIter<Function<'s>>,
-    seen_ranges: BTreeSet<(u64, u64)>,
     range_buf: Vec<Range>,
+    seen_ranges: BTreeSet<(u64, u64)>,
     finished: bool,
 }
 
--- a/symbolic-debuginfo/src/dwarf.rs
+++ b/symbolic-debuginfo/src/dwarf.rs
@@ -713,7 +713,11 @@ impl<'d, 'a> DwarfUnit<'d, 'a> {
     }
 
     /// Collects all functions within this compilation unit.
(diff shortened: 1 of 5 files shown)


jan-auer:
I'll have a closer look at this PR at a later time. Just as an initial warning: Please be extremely careful with this change. IIRC, we used to have such deduplication logic and it caused issues.

Swatinem:
I moved the deduplication step into the `functions` iterator per our discussion and added a comment explaining what we *think* is going on.

Next steps:
* run the offending object through breakpads own `dump_syms`
* test electron before/after and see what the symcache differences are. electron stresses our symcache limits quite a bit

symbolic-symcache/src/writer.rs
```diff
@@ -205,10 +206,15 @@ where
             .debug_session()
             .map_err(|e| SymCacheError::new(SymCacheErrorKind::BadDebugFile, e))?;
 
+        let mut seen_ranges = BTreeSet::new();
         for function in session.functions() {
             let function =
                 function.map_err(|e| SymCacheError::new(SymCacheErrorKind::BadDebugFile, e))?;
-            writer.add_function(function)?;
+            let fn_range = (function.address, function.size);
+            if !seen_ranges.contains(&fn_range) {
```
jan-auer: You can simplify this to:

```rust
if seen_ranges.insert(fn_range) {
  writer.add_function(function)?;
}
```

symbolic-symcache/src/writer.rs
```diff
@@ -205,10 +206,15 @@ where
             .debug_session()
             .map_err(|e| SymCacheError::new(SymCacheErrorKind::BadDebugFile, e))?;
 
+        let mut seen_ranges = BTreeSet::new();
```
jan-auer: Would it be enough to just track the function start address, provided that we just deal with top-level functions?

symbolic-debuginfo/src/dwarf.rs
```diff
@@ -1356,6 +1359,18 @@ impl<'s> Iterator for DwarfFunctionIterator<'s> {
...
         loop {
             if let Some(func) = self.functions.next() {
+                // We have seen this iterator yield duplicate top-level function entries, which combined
+                // with recursively walking its inlinees can blow past symcache limits.
+                // We suspect the reason is that the iterator does not internally deduplicate top level
+                // functions defined in different compile units. We suspect this might be caused by link-time
+                // deduplication which merges templated code that is being generated multiple times in each
+                // compilation unit.
+                let fn_range = (func.address, func.size);
+                if self.seen_ranges.contains(&fn_range) {
```
jan-auer, line 1369: You can simplify this to:

```rust
if !self.seen_ranges.insert((func.address, func.size)) {
  continue;
}
```

symbolic-debuginfo/src/dwarf.rs
```diff
@@ -1356,6 +1359,18 @@ impl<'s> Iterator for DwarfFunctionIterator<'s> {
 
         loop {
             if let Some(func) = self.functions.next() {
+                // We have seen this iterator yield duplicate top-level function entries, which combined
+                // with recursively walking its inlinees can blow past symcache limits.
+                // We suspect the reason is that the iterator does not internally deduplicate top level
+                // functions defined in different compile units. We suspect this might be caused by link-time
+                // deduplication which merges templated code that is being generated multiple times in each
+                // compilation unit.
+                let fn_range = (func.address, func.size);
```
jan-auer, line 1368: Would it be enough to just track the function start address, provided that we just deal with top-level functions?

symbolic-debuginfo/src/dwarf.rs
```diff
@@ -1342,6 +1344,7 @@ impl<'s> Iterator for DwarfFileIterator<'s> {
 pub struct DwarfFunctionIterator<'s> {
     units: DwarfUnitIterator<'s>,
     functions: std::vec::IntoIter<Function<'s>>,
+    seen_ranges: BTreeSet<(u64, u64)>,
```
jan-auer, line 1347: On second thought, MSVC may output similar duplicate entries. Can you check whether this is the case and we need to port this over, too? 

Also, I'm curious on the runtime impact on this solution. The function iterator is already quite slow due to its sorting, but we should be careful adding another _n log n_ contributor.

symbolic-debuginfo/src/dwarf.rs
```diff
@@ -1356,6 +1359,18 @@ impl<'s> Iterator for DwarfFunctionIterator<'s> {
...
             if let Some(func) = self.functions.next() {
+                // We have seen this iterator yield duplicate top-level function entries, which combined
+                // with recursively walking its inlinees can blow past symcache limits.
+                // We suspect the reason is that the iterator does not internally deduplicate top level
+                // functions defined in different compile units. We suspect this might be caused by link-time
+                // deduplication which merges templated code that is being generated multiple times in each
+                // compilation unit.
+                let fn_range = (func.address, func.size);
+                if self.seen_ranges.contains(&fn_range) {
+                    continue;
```
jan-auer, line 1370: This check is happening too late. At this time, you have already collected the function and its full inline tree into a sorted vector. You can avoid this by pushing down `&mut seen_ranges` into `DwarfUnit::functions` a few lines below.

jan-auer (changes requested):

symbolic-debuginfo/src/dwarf.rs
```diff
@@ -769,6 +773,18 @@ impl<'d, 'a> DwarfUnit<'d, 'a> {
...
+            // We have seen duplicate top-level function entries being yielded from the
+            // [`DwarfFunctionIterator`], which combined with recursively walking its inlinees can
+            // blow past symcache limits.
+            // We suspect the reason is that the the same top-level functions might be defined in
+            // different compile units. We suspect this might be caused by link-time deduplication
+            // which merges templated code that is being generated multiple times in each
+            // compilation unit. We make sure to detect this here, so we can avoid creating these
+            // duplicates as early as possible.
+            if !seen_ranges.insert((function_address, function_size)) {
+                continue;
```
jan-auer, line 785: Now, you need to set `skipped_depth = Some(depth);` before continuing. See other branches where we skip a function.

symbolic-debuginfo/src/dwarf.rs
```diff
@@ -1249,6 +1267,7 @@ impl<'data> DwarfDebugSession<'data> {
             functions: Vec::new().into_iter(),
             range_buf: Vec::new(),
             finished: false,
+            seen_ranges: BTreeSet::new(),
```
jan-auer, line 1270: _nit_: Move this before `finished`.

jan-auer (approved):

PR #407, opened 2021-06-15
fix(cfi): Don't build the MachO symbols in the loop

Gankra:
symbol_map is a Vec that is quite expensive to create, so doing it every time we encounter a dwarf FDE entry is extremely quadratic and slow. This causes macos firefox builds to timeout while trying to dump the symbols for XUL -- especially for aarch64 builds (for whatever reason). 

You can test that out by running dump_syms on the XUL inside target.dmg for this build: https://treeherder.mozilla.org/jobs?repo=try&revision=5a2fbeb4112cd87eee7bb3f1b48223fc9b3ff7b9&selectedTaskRun=O2kbKrM-RauVaLGx-7f0MA.0

This change drops dump_syms on XUL from ~2.5 minutes to ~4 seconds on my macbook pro.

### DIFF
--- a/symbolic-minidump/src/cfi.rs
+++ b/symbolic-minidump/src/cfi.rs
@@ -411,6 +411,8 @@ impl<W: Write> AsciiCfiWriter<W> {
             }
             Ok(())
         }
+        // Preload the symbols as this is expensive to do in the loop.
+        let symbols = object.symbol_map();
 
         // Initialize an unwind context once and reuse it for the entire section.
         let mut ctx = UninitializedUnwindContext::new();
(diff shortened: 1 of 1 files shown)


Gankra:
See also: https://bugzilla.mozilla.org/show_bug.cgi?id=1716619

Gankra:
(I think ideally we would not build symbols_map at all and instead do a more precise query, but this is the minimal fix because I don't know if that kind of random access is possible.)

jan-auer (approved):
Good catch!

jan-auer:
> I think ideally we would not build symbols_map at all and instead do a more precise query, but this is the minimal fix because I don't know if that kind of random access is possible.

It would be possible in principle, with changes in `goblin`. That is, at least for MachO. The underlying [`Symbols`](https://docs.rs/goblin/0.4.1/goblin/mach/symbols/struct.Symbols.html) type would support manual binary search if we expose the `nsyms` field in addition to the random-access `get` function.

However, that all requires that the symbols are ordered by address, and I seem to remember that we encountered files where that wasn't the case.

Gankra:
@jan-auer how long would it take to cut an 8.2.1 release with these changes?

PR #409, opened 2021-06-21
fix: Skip broken inlinees in PDBs

Swatinem:
Some PDBs fail processing because some inlinees result in invalid checksum errors. I suspect this might be because of broken file index annotations. Instead of failing the complete file, we silently skip these now.

Should solve https://getsentry.atlassian.net/browse/ISSUE-1238, and have that file process successfully.

### DIFF
--- a/symbolic-debuginfo/src/pdb.rs
+++ b/symbolic-debuginfo/src/pdb.rs
@@ -1122,7 +1122,14 @@ impl<'s> Unit<'s> {
                     // skip silently instead of erroring out. Missing a single inline function is
                     // more acceptable in such a case than halting iteration completely.
                     if let Some(inlinee) = inlinees.get(&site.inlinee) {
-                        self.handle_inlinee(site, parent_offset, inlinee, &program)?
+                        // We have seen some PDB files in the wild that can lead to `UnimplementedFileChecksumKind`
+                        // errors, possibly due to some broken file index annotations (see the comment
+                        // in `InlineeLineIterator`). Similarly to all the other cases here, we rather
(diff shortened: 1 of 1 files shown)


flub (approved):

symbolic-debuginfo/src/pdb.rs
```diff
@@ -1122,7 +1122,13 @@ impl<'s> Unit<'s> {
...
                     // more acceptable in such a case than halting iteration completely.
                     if let Some(inlinee) = inlinees.get(&site.inlinee) {
-                        self.handle_inlinee(site, parent_offset, inlinee, &program)?
+                        // We have seen some PDB files in the wild that can lead to `UnimplementedFileChecksumKind`
+                        // errors, possibly due to some broken file index annotations (see the comment
+                        // in `InlineeLineIterator`). Similarly to all the other cases here, we rather
+                        // silently skip these than to fail processing the whole file.
+                        self.handle_inlinee(site, parent_offset, inlinee, &program)
+                            .ok()
+                            .flatten()
```
jan-auer, line 1131: Please add a TODO to log this error. 

Did you take a closer look what kind of checksum is in that PDB? If this is a new checksum kind, we should be implementing the new kind rather than _just_ silently ignoring the error.

Swatinem: I was stepping through this, and also the `name_offset` is completely bogus (extremely large, pointing way past the size of the PDB itself), so I think it really is a broken `file_index`. We log quite a few of these checksum type errors, and they all have different values.

PR #414, opened 2021-06-29
fix: Use the BcSymbolMap for names and paths coming from DWARF

Swatinem:


loewenheim (approved):

symbolic-debuginfo/src/dwarf.rs
```diff
@@ -394,6 +396,7 @@ impl<'d, 'a> UnitRef<'d, 'a> {
         &self,
         entry: &Die<'d, '_>,
         language: Language,
+        bcsymbolmap: Option<&BcSymbolMap<'d>>,
```
flub, line 399: In rust I'm API-wise usually tempted to create a `resolve_function_name_with_bcsymbolmap()` function instead of having a `None` argument to a function.

symbolic-debuginfo/src/dwarf.rs
```diff
@@ -441,14 +444,19 @@ impl<'d, 'a> UnitRef<'d, 'a> {
...
     prefer_dwarf_names: bool,
 }
 
 impl<'d, 'a> DwarfUnit<'d, 'a> {
     /// Creates a DWARF unit from the gimli `Unit` type.
-    fn from_unit(unit: &'a Unit<'d>, info: &'a DwarfInfo<'d>) -> Result<Option<Self>, DwarfError> {
+    fn from_unit(
+        unit: &'a Unit<'d>,
+        info: &'a DwarfInfo<'d>,
+        bcsymbolmap: Option<&'d BcSymbolMap<'d>>,
```
flub, line 458: same wrt the `Option` vs new constuctor

symbolic-debuginfo/src/dwarf.rs
```diff
@@ -1320,6 +1344,31 @@ impl<'s> Iterator for DwarfUnitFileIterator<'s> {
...
+        match s {
+            Cow::Borrowed(s) => bcsymbolmap.resolve(s).into(),
+            // we make an unconditional copy here, since the borrow checker has no idea
+            // if the returned `str` is owned by the `bcsymbolmap`, or by `s`.
+            Cow::Owned(s) => bcsymbolmap.resolve(&s).to_owned().into(),
+        }
+    } else {
+        s
+    }
+}
```
flub, line 1370: I wonder if these should be on the `BcSymbolMap` struct.  but that would mean committing to a public API so perhaps it's fine to leave them here as helpers.  (If you want to go all out you could make a local extension trait, but meh)

flub (approved):

symbolic-debuginfo/src/dwarf.rs
```diff
@@ -1320,6 +1344,31 @@ impl<'s> Iterator for DwarfUnitFileIterator<'s> {
     }
 }
 
+fn resolve_byte_name<'s>(bcsymbolmap: Option<&BcSymbolMap<'s>>, s: &'s [u8]) -> &'s [u8] {
+    if let Some(bcsymbolmap) = bcsymbolmap {
+        std::str::from_utf8(s)
+            .map(|s| bcsymbolmap.resolve(s))
```
jan-auer, line 1350: What would you think about adding a binary version of `resolve`; or even changing it to work with `AsRef<[u8]>`?

symbolic-debuginfo/src/dwarf.rs
```diff
@@ -1248,13 +1265,20 @@ impl<'data> DwarfDebugSession<'data> {
...
         })?;
 
-        Ok(DwarfDebugSession { cell })
+        Ok(DwarfDebugSession {
+            cell,
+            bcsymbolmap: None,
+        })
+    }
+
+    pub(crate) fn set_bcsymbolmap(&mut self, symbolmap: Option<Arc<BcSymbolMap<'data>>>) {
```
jan-auer, line 1274: In the MachO code, this is called `load_symbolmap`. Align the naming and add a doc comment, even though this is not public? 

FWIW, having a dependency on MachO in this implementation is certainly not ideal, but it is the quickest way to go about it. A probably cleaner approach would be to generalize symbol mapping in a trait, but for this one change that may well be an overkill. The last approach dependency-wise would be to wrap the bare DebugSession and transform the `Function` entries we receive back from the session. Since that has it's entirely different problems, though, your solution is about as good as I think it can get now.

symbolic-debuginfo/src/dwarf.rs
```diff
@@ -1320,6 +1344,31 @@ impl<'s> Iterator for DwarfUnitFileIterator<'s> {
...
+    }
+}
+
+fn resolve_cow_name<'s>(bcsymbolmap: Option<&BcSymbolMap<'s>>, s: Cow<'s, str>) -> Cow<'s, str> {
+    if let Some(bcsymbolmap) = bcsymbolmap {
+        match s {
+            Cow::Borrowed(s) => bcsymbolmap.resolve(s).into(),
+            // we make an unconditional copy here, since the borrow checker has no idea
+            // if the returned `str` is owned by the `bcsymbolmap`, or by `s`.
+            Cow::Owned(s) => bcsymbolmap.resolve(&s).to_owned().into(),
```
jan-auer, line 1365: Why do you convert this branch to owned, but not the other? You can always borrow from the symbolmap. IMHO, this entire function should just be (inlined):

```rust
bcsymbolmap.map_or(s, |m| m.resolve(&s).into())
```

Swatinem: `resolve()` is bound to the lifetime of the argument, since it might return the argument unmodified. Thats what I mean by the borrow checker can’t tell the difference.

symbolic-debuginfo/src/dwarf.rs
Swatinem: I tried doing that at first (modifying the `Function` / `FileInfo` structs in the outer iterator), but that would make things a lot more complex actually.

symbolic-debuginfo/src/dwarf.rs
jan-auer: Thanks, I missed that. Looking at [`BcSymbolMap`](https://docs.rs/symbolic/8.2.1/symbolic/debuginfo/macho/struct.BcSymbolMap.html#implementations), we might actually want to extend its API surface indeed. We have a `get(usize)` which returns an `Option`, but no `resolve` that does so. Maybe what we want is:

- `get(&self, usize) -> Option<&'d str>`
- `resolve(&self, &str) -> Option<&'d str>`
- `resolve_default(&self, &'d str) -> &'d str` (note that this is binding to the same `'d` as the data)

Not sure about the binary interface, though. Probably, it's fine to assume you only want to pass UTF-8 strings.

symbolic-debuginfo/src/dwarf.rs
Swatinem: The convenience in these helpers is also that they take care of the `Option`. I simplified them a bit, but they are just too convenient to pass by :-D

PR #425, opened 2021-09-09
Make dsym_parent accept .framework.dSYM

luser:
Currently `DSymPathExt::dsym_parent` expects the bundle name to be exactly `Framework.dSYM`, but for frameworks it can also be `Framework.framework.dSYM`. This change makes it accept either.

I ran into this issue trying to use symbolic with dSYM bundles from our framework at FullStory. The path to the DWARF file in them is `FullStory.framework.dSYM/Contents/Resources/DWARF/FullStory`. I created a new framework project in Xcode and archived it to verify that we weren't doing anything nonstandard. The resulting dSYM is included as a test fixture in this PR, you can see that it's named `Example.framework.dSYM`.


symbolic-common/src/path.rs
```diff
@@ -509,7 +509,19 @@ impl DSymPathExt for Path {
...
-        if parent.file_stem() == Some(framework) && parent.is_dsym_dir() {
+        // Accept both Filename.dSYM and Filename.framework.dSYM as
+        // the bundle directory name.
+        let stem_matches = parent.file_name()
+            .and_then(|name| Path::new(name).file_stem())
+            .map(|stem| {
+                if stem == framework {
+                    return true;
+                }
+                let mut alt = framework.to_owned();
```
luser, line 520: This is annoying. I guess in hindsight I could have done:
```rust
  let alt = Path::new(framework);
  alt.file_stem() == Some(framework) && alt.extension() == Some(".framework")
```

Swatinem (approved):

symbolic-common/src/path.rs
Swatinem: Might as well change, you need to make rustfmt happy anyway :-D

PR #426, opened 2021-09-10
Use pdb_addr2line::TypeFormatter.

### LINKED ISSUE #423: Some pdb functions are missing arguments
When looking up functions in pdb files, some functions are missing their argument types.
For example, you get `_except_handler4` instead of `_except_handler4(_EXCEPTION_RECORD*, _EXCEPTION_REGISTRATION_RECORD*, _CONTEXT*, void*)`, and `is_potentially_valid_image_base` instead of `is_potentially_valid_image_base(void* const)`.

This happens for inlined functions, and for procedures whose argument list is stored in the type info (and for which there is no public symbol at the same address with a mangled name).

### Steps to Reproduce

One way to reproduce this is using the addr2line example from the symbolic repo. Strictly speaking, you might expect it to print no arguments because [it uses `name_only()`](https://github.com/getsentry/symbolic/blob/4e3dc0a9f211588b140bdf3fdf1658fcab8cefcf/examples/addr2line/src/main.rs#L15) when demangling, but for these functions the name is not mangled anyway.

```
% cargo run --release -p addr2line -- -Cfi -e symbolic-debuginfo/tests/fixtures/crash.pdb 0x355b
_except_handler4
  at ??:0
```

```
% cargo run --release -p addr2line -- -Cfi -e symbolic-debuginfo/tests/fixtures/crash.pdb 0x30bf
is_potentially_valid_image_base
  at f:\dd\vctools\crt\vcstartup\src\utility\utility.cpp:35
__scrt_is_nonwritable_in_current_image
  at f:\dd\vctools\crt\vcstartup\src\utility\utility.cpp:99
```

Looking up the same addresses using [pdb-addr2line](https://github.com/mstange/pdb-addr2line) gives the full arguments:

```
% cargo install --examples pdb-addr2line
% pdb-addr2line -Cfi -e symbolic-debuginfo/tests/fixtures/crash.pdb 0x355b
_except_handler4(_EXCEPTION_RECORD*, _EXCEPTION_REGISTRATION_RECORD*, _CONTEXT*, void*)
??:?
% pdb-addr2line -Cfi -e symbolic-debuginfo/tests/fixtures/crash.pdb 0x30bf
is_potentially_valid_image_base(void* const)
f:\dd\vctools\crt\vcstartup\src\utility\utility.cpp:35
__scrt_is_nonwritable_in_current_image(void const*)
f:\dd\vctools\crt\vcstartup\src\utility\utility.cpp:99
```

mstange:
This replaces symbolic's TypeFormatter with the one from the pdb_addr2line
crate, which is more fully-featured. Most importantly, it supports
function arguments, both for inlined functions (which use IdIndex) and
for procedures (which use TypeIndex).

Fixes #423.

As a consequence of this commit, function names for procedures from PDBs
will now always contain arguments. These arguments are obtained from the
type data, and not from demangling. Consequently, their presence is
unaffected by demangler flags, and users cannot opt out of the arguments.
Similarly, we never print the return type, and there is no way for users
to opt in to showing the return type.

This commit also moves some things related to ModuleInfo loading around.
This was needed to make the lifetimes work out successfully.
Previously, the TypeFormatter was created just for the duration of the
function name printing, and it could depend on fields of PdbDebugInfo;
this was ok because the ItemMap caching was outside of the TypeFormatter.
Now all caching is inside of TypeFormatter, so we don't want to re-create
the formatter repeatedly, and instead store it on the PdbDebugInfo. But
this means that some things need to be stored *outside* of PdbDebugInfo
so that TypeFormatter can have a lifetime dependency on those things,
specifically on ModuleInfo and StringTable. So these things move to the
PdbStreams struct.
The ModuleInfo caching now uses a FrozenMap (from the elsa crate) instead
of a Vec of LazyCells. The Vec needed to know its size ahead-of-time and
couldn't grow, but PdbStreams cannot know the number of modules at creation
time. (At least not in a way where the results of the module iteration are
passed on to the other parts of the system that need the module header
list.)

r? @Swatinem

### DIFF
--- a/symbolic-debuginfo/src/pdb.rs
+++ b/symbolic-debuginfo/src/pdb.rs
@@ -1,20 +1,19 @@
 //! Support for Program Database, the debug companion format on Windows.
 
 use std::borrow::Cow;
-use std::cell::{RefCell, RefMut};
-use std::cmp::Ordering;
-use std::collections::btree_map::{BTreeMap, Entry};
+use std::collections::btree_map::BTreeMap;
 use std::error::Error;
 use std::fmt;
 use std::io::Cursor;
 use std::sync::Arc;
 
-use lazycell::LazyCell;
+use elsa::FrozenMap;
 use parking_lot::RwLock;
-use pdb::{
-    AddressMap, FallibleIterator, InlineSiteSymbol, ItemIndex, LineProgram, MachineType, Module,
-    ModuleInfo, PdbInternalSectionOffset, ProcedureSymbol, SymbolData,
+use pdb_addr2line::pdb::{
+    AddressMap, FallibleIterator, InlineSiteSymbol, LineProgram, MachineType, Module, ModuleInfo,
+    PdbInternalSectionOffset, ProcedureSymbol, SymbolData,
 };
+use pdb_addr2line::ModuleProvider;
 use smallvec::SmallVec;
 use thiserror::Error;
 
@@ -31,7 +30,7 @@ const MAGIC_BIG: &[u8] = b"Microsoft C/C++ MSF 7.00\r\n\x1a\x44\x53\x00\x00\x00"
 
 // Used for CFI, remove once abstraction is complete
 #[doc(hidden)]
-pub use pdb;
+pub use pdb_addr2line::pdb;
 
 /// The error type for [`PdbError`].
 #[non_exhaustive]
@@ -98,6 +97,16 @@ impl From<pdb::Error> for PdbError {
 impl From<fmt::Error> for PdbError {
     fn from(e: fmt::Error) -> Self {
         Self::new(PdbErrorKind::FormattingFailed, e)
+    }
+}
+
+impl From<pdb_addr2line::Error> for PdbError {
+    fn from(e: pdb_addr2line::Error) -> Self {
+        match e {
+            pdb_addr2line::Error::PdbError(e) => Self::new(PdbErrorKind::BadObject, e),
+            pdb_addr2line::Error::FormatError(e) => Self::new(PdbErrorKind::FormattingFailed, e),
+            e => Self::new(PdbErrorKind::FormattingFailed, e),
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


loewenheim (approved):
This looks good to me, but I think @Swatinem will have to have a look when he gets back.

symbolic-debuginfo/src/pdb.rs
```diff
@@ -409,119 +419,102 @@ impl<'data, 'object> Iterator for PdbSymbolIterator<'data, 'object> {
...
-    pdb: Arc<RwLock<Pdb<'d>>>,
+    /// The streams, to load module streams on demand.
+    streams: &'d PdbStreams<'d>,
     /// All module headers for repeated iteration.
-    modules: Vec<Module<'d>>,
-    /// Lazy loaded module streams in the same order as headers.
-    module_infos: Vec<LazyCell<Option<ModuleInfo<'d>>>>,
-    /// Cache for module by name lookup for cross module imports.
-    module_exports: RefCell<BTreeMap<pdb::ModuleRef, Option<pdb::CrossModuleExports>>>,
+    modules: Rc<Vec<Module<'d>>>,
```
Swatinem, line 483: Let me double-check if its okay to make this `!Send` (even though the `RefCell`s previously made this `!Sync`).

(This also makes me wonder why `TypeFormatter` needs a `Rc<Vec<_>>` and is not satisfied with a `&[_]` instead.)

Swatinem (commented):
lgtm, though I will check if we need this to be `Send`. (If so, might as well add a test for that separately)

Also @jan-auer voiced some concerns about pulling in another dependency, and suggested to maybe move all of this into the `pdb` crate.

Swatinem (approved):

mstange:
I've rebased the PR. I've also eliminated the use of `Rc`.

@Swatinem, could you take another look? This PR has become relevant to me again because it's blocking https://github.com/mozilla/dump_syms/issues/406.

symbolic-debuginfo/src/pdb.rs
mstange: I took another look at this. A slice is indeed not an option: Let's say it was `&'list [Module<'dbi>]`, what would those two lifetimes be? Who would own the `Vec`, and who would own the `DebugInformation`?

But at least I've eliminated the `Rc` now:

I've changed `Rc<Vec<Module<'d>>` to `Vec<Module<'d>>`, i.e. I've given the `TypeFormatter` full ownership of the `Vec` instead of just shared ownership.
Now everyone who wants to access the list of modules needs to go through the `TypeFormatter` - `TypeFormatter::modules()` exposes the slice.

Swatinem (approved):

PR #427, opened 2021-09-10
Delete unused and misleading fixtures.

mstange:
There are fixtures at the following paths:
 - symbolic-testutils/fixtures/
 - symbolic-debuginfo/tests/fixtures/

For example, both have a crash.pdb which looks very similar on the
surface but is actually slightly different. I lost a bunch of time
because I was investigating the wrong file.

The symbolic-debuginfo tests use fixtures under symbolic-testutils.
Most of the fixtures in symbolic-debuginfo are unused. This commit
removes them.


relaxolotl (approved):
Thanks for removing these! It looks like these may have accidentally snuck in during a large refactor (https://github.com/getsentry/symbolic/pull/127/commits/55e7614db102ef9e6e8f96d9b67e9d9c48983051), but it's clear that these aren't being used anywhere.

PR #430, opened 2021-09-17
fix: Safe multibyte char string splitting

flub:
This was splitting strings in an unsafe way, assuming you could always
split at 33 bytes from the end which is not true in UTF-8 strings.

Fixes SYMBOLICATOR-GE

### DIFF
--- a/symbolic-demangle/src/lib.rs
+++ b/symbolic-demangle/src/lib.rs
@@ -485,5 +485,9 @@ mod test {
             strip_hash_suffix("hello$0123456789abcdef0123456789abcdxx"),
             "hello$0123456789abcdef0123456789abcdxx"
         );
+        assert_eq!(
+            strip_hash_suffix("hello$\u{1000}0123456789abcdef0123456789abcde"),
+            "hello$\u{1000}0123456789abcdef0123456789abcde"
+        );
     }
 }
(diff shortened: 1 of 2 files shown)


symbolic-demangle/src/lib.rs
```diff
@@ -189,12 +189,22 @@ fn strip_hash_suffix(ident: &str) -> &str {
     if len < 33 {
```
relaxolotl: nit: you could probably remove the `else` indentation if you just returned in this if block

symbolic-demangle/src/lib.rs
```diff
@@ -459,4 +470,24 @@ mod test {
...
+            "hello_0123456789abcdef0123456789abcdef",
+        );
+        assert_eq!(
+            strip_hash_suffix("hello\u{1000}0123456789abcdef0123456789abcdef"),
+            "hello\u{1000}0123456789abcdef0123456789abcdef"
+        );
+        assert_eq!(
+            strip_hash_suffix("hello$0123456789abcdef0123456789abcdxx"),
+            "hello$0123456789abcdef0123456789abcdxx"
+        );
```
relaxolotl, line 487: nit: could i bother you to add a unit test with a 2-byte character that comes after the `$` to check for the scenario where a multibyte char bumps the iterator past the expected index?

relaxolotl (approved):

symbolic-demangle/src/lib.rs
flub: clippy beat you to this one 😄

PR #431, opened 2021-09-17
draft: figure out why CI isn't breaking when compilation is failing

relaxolotl:


relaxolotl:
conclusion: CI isn't failing because it's not like it's doing anything wrong. to catch the issue fixed by https://github.com/getsentry/symbolic/pull/429 requires CI to compile the repository twice: once prior to a proposed dependency upgrade, and then a second time with the upgrades applied and the build artifacts (namely `Cargo.lock`) preserved during the second build.

PR #434, opened 2021-09-28
Be incredibly lenient when parsing ELFs

relaxolotl:
This is a proof of concept PR to demonstrate an excessive solution that could be used to sidestep problems like the one mentioned in https://github.com/getsentry/sentry-dart/issues/591, where only a portion of an ELF is malformed but the rest of it can be perfectly parsed. The ELF parser has been changed so that malformed ELFs could still be partially parsed out instead of eagerly failing at the first error.

The PR basically copy-pastes nearly everything from goblin's ELF parser, and replaces all short-circuited errors with a macro that returns the ELF object with everything successfully parsed up until the failure. A good second iteration on this would be to merely skip unparseable pieces but not short-circuit on the first failure, allowing us to, for example parse out dynamic symbols even if regular symbol parsing fails. 

An additional field as been added to all object types to indicate whether it's malformed or not, in order to provide some feedback in the event that an ELF is malformed and cannot be fully parsed. Ideally this is properly used in other objects.

An ideal alternative to this would probably be to extend goblin's repertoire of `lazy_parse` helpers, following the idea mentioned in https://github.com/m4b/goblin/pull/254#issuecomment-770583461 instead. Following that path would allow us to properly lazily parse the file instead of just copying line-for-line what `Elf::parse()` is doing.

### DIFF
--- a/examples/object_debug/src/main.rs
+++ b/examples/object_debug/src/main.rs
@@ -39,6 +39,7 @@ fn inspect_object<P: AsRef<Path>>(path: P) -> Result<(), Box<dyn std::error::Err
                 println!("   symbol table: {}", object.has_symbols());
                 println!("   debug info:   {}", object.has_debug_info());
                 println!("   unwind info:  {}", object.has_unwind_info());
+                println!("   is malformed: {}", object.is_malformed());
             }
             Err(e) => {
                 print!(" - ");
--- a/symbolic-debuginfo/src/base.rs
+++ b/symbolic-debuginfo/src/base.rs
@@ -694,6 +694,9 @@ pub trait ObjectLike<'data, 'object> {
 
     /// Determines whether this object contains embedded sources.
     fn has_sources(&self) -> bool;
+
+    /// Determines whether this object is malformed and was only partially parsed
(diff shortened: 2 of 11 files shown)


Swatinem (approved):

Swatinem:
Fixed test failures (snapshot updates, reading the interpreter correctly).

Should be good to go, if we really want to get this hotfixed.

loewenheim (approved):

relaxolotl:
Updated some docs, and made successful parsing of section headers a requirement to return a partially parsed ELF.

There are some follow-up actions that ideally should be made to refine this hotfix:
- Ensure that all of the well-formed sections that can be parsed are parsed when there is a malformed section, ie make an active attempt to provide as many debug file features (symbol tables, unwinding, etc) as we can
- Ensure that well-formed ELFs with zero usable information behave the same as before even with the previous bullet point
- Unit tests to double check that the object API isn't broken in unexpected ways due to this partial parse

I'll also try to pursue the goal mentioned in the last paragraph of the description: 
> extend goblin's repertoire of lazy_parse helpers, following the idea mentioned in m4b/goblin#254 (comment)

PR #445, opened 2021-10-21
fix: Only speculate stackless function on x64

Swatinem:
The workaround that we put in there was only hardcoded for x64, and should not be used for other 64-bit architectures.


jan-auer:
⬆️  Please add some context to this change, as this diverges from Breakpad's converter.

PR #449, opened 2021-11-17
fix(unreal): Support the UE5 crash reporter

jan-auer:
UE5 introduced a `"CR1"` marker at the beginning of crash archives. It indicates
that the header at the beginning of the file is valid. Previously, that header
was merely a placeholder and a second header at the end of the file contained
valid information. `Unreal4Crash::parse` now supports both versions by checking
for "CR1".

See https://github.com/EpicGames/UnrealEngine/commit/a0471b76577a64e5c4dad89a38dfe7d9611a65ef
Fixes https://github.com/getsentry/sentry/issues/29940
cc @redxdev

### DIFF
--- a/symbolic-unreal/src/container.rs
+++ b/symbolic-unreal/src/container.rs
@@ -89,6 +89,7 @@ impl TryFromCtx<'_, usize> for Unreal4FileMeta {
 
     fn try_from_ctx(data: &[u8], file_offset: usize) -> Result<(Self, usize), Self::Error> {
         let mut offset = 0;
+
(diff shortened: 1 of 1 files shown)


jan-auer:
The file I was given to verify contains PII, so we cannot use it in the test suite. We'll need to create our own by checking out at least the UE version mentioned in the PR description.

cc @relaxolotl @Swatinem

Swatinem (approved):

relaxolotl (approved):

PR #452, opened 2021-11-23
fix(dsym): Correctly resolve dSYM paths if their filename includes .app

### LINKED ISSUE #451: `resolve_dsym` does not properly strip nested extensions
`resolve_dsym` does not properly strip nested extensions, e.g. `.app` in `testApp.app.dSYM`. It attempts to look for a debug file under `testApp.app.dSYM/Contents/Resources/DWARF/testApp.app`, instead of the correct location at `testApp.app.dSYM/Contents/Resources/DWARF/testApp`.

relaxolotl:
It turns out the files of dSYMs generated by XCode are slightly different from the ones that can be downloaded from App Store Connect.

XCode and dwarfdump produce filenames along the lines of `[appName].app.dSYM`, whereas App Store Connect produces `[appName].dSYM`. Despite the different dSYM names, they both contain a debug file that's simply named `[appName]`, as far as I can tell. This updates `resolve_dsym` to recognize and strip `.app` if it's present. I've opted to take the more conservative route of only stripping `.app` instead of attempting to remove as many nested extensions as possible, in case some extensions need to be preserved.

Fixes #451.


Swatinem (approved):
Do we need to commit this 600K+ file, or can we truncate it? We are just asserting its presence after all.

relaxolotl:
good point, i've removed all of the files that aren't explicitly used by the tests that introduce the new set of directories, and stripped the debug file of its contents since that's not being used anyways.

PR #458, opened 2021-11-25
feat: Add new SymCache serializer and ObjectLike converter

Swatinem:
This includes the code that serializes the new SymCache to disk, as well as code to process
`ObjectLike` symbolic-debuginfo objects.

This also has a small shim with the exact same API as the current `SymCacheWriter`.


loewenheim (commented):
Is there a file missing?

Swatinem:
absolutely :-D I also commited that half an hour ago, but seems like I never pushed, lol.

loewenheim (approved):

Issue #472, opened 2021-12-17
Replace walrus with straight wasmparser


Jake-Shadle:
When I first added the dependency on symbolic-debuginfo, I immediately noticed that it added a duplicate dependency on wasmparser (we use wasmtime which is using 0.81, the latest at the time of writing) and saw that the only reason for this was because the wasm code is using walrus as the higher level parser for wasm binaries. Unfortunately, even though it is maintained by many of the same people who maintain wasmtime/wasmparser, we have seen in the past that walrus is mostly passively maintained as opposed to wasmparser which is updated frequently, which leads to frequent cases where duplicate versions of wasmparser are in your dependency tree if you depend on something that uses wasmparser, such as wasmtime.

Not only that, but wasmparser is also depended on directly since walrus doesn't expose the base code address with a note to use the same version that walrus depends on. When I looked at the code, there is really very little information that the wasm parser needed, so it seemed like a better way forward to just put a small parser based on wasmparser into the crate to get rid of the problematic walrus dependency altogether.

I have working solution for this, I'm just waiting on #470 to be merged since it is based off of that one since they are both about reducing dependencies and making symbolic-debuginfo a bit nicer to use from something other than sentry-cli.

PR #474, opened 2021-12-20
Replace walrus

Jake-Shadle:
Opening as a draft since this PR depends on #470 

As stated in #472, this PR removes walrus due to its passively maintained status, and instead uses `wasmparser` directly. This means that not only can `wasmparser` be upgraded to follow the latest version much more easily and remove a dependency, it also means the runtime memory usage and CPU time should be greatly reduced as walrus was doing a _lot_ more work when parsing than this implementation, which (other than the bits needed for validation) only parses the sections/content actually used by this crate.

Resolves: #472

### DIFF
--- a/symbolic-debuginfo/src/wasm.rs
+++ b/symbolic-debuginfo/src/wasm.rs
@@ -14,6 +14,7 @@ mod parser;
 
 /// An error when dealing with [`WasmObject`](struct.WasmObject.html).
 #[derive(Debug, Error)]
+#[non_exhaustive]
 pub enum WasmError {
     /// Failed to read data from a WASM binary
     #[error("invalid wasm file")]
--- a/symbolic-debuginfo/src/wasm.rs
+++ b/symbolic-debuginfo/src/wasm.rs
@@ -43,17 +43,6 @@ impl<'data> WasmObject<'data> {
         data.starts_with(b"\x00asm")
     }
 
-    /// Tries to parse a WASM from the given slice.
-    pub fn parse(data: &'data [u8]) -> Result<Self, WasmError> {
-        let mut obj = parser::parse(data)?;
-
-        // The actual iterator just wraps the vec, so we reverse it and just pop off
-        // the end
-        obj.funcs.reverse();
-
-        Ok(obj)
-    }
-
     /// The container file format, which currently is always `FileFormat::Wasm`.
     pub fn file_format(&self) -> FileFormat {
         FileFormat::Wasm
--- a/symbolic-debuginfo/src/wasm.rs
+++ b/symbolic-debuginfo/src/wasm.rs
@@ -97,7 +97,8 @@ impl<'data> WasmObject<'data> {
     /// Returns an iterator over symbols in the public symbol table.
     pub fn symbols(&self) -> WasmSymbolIterator<'data, '_> {
         WasmSymbolIterator {
-            funcs: Box::new(self.funcs.clone().into_iter()),
+            funcs: self.funcs.clone().into_iter(),
+            _marker: std::marker::PhantomData,
         }
     }
 
@@ -274,7 +275,8 @@ impl<'data> Dwarf<'data> for WasmObject<'data> {
 ///
 /// Returned by [`WasmObject::symbols`](struct.WasmObject.html#method.symbols).
 pub struct WasmSymbolIterator<'data, 'object> {
-    funcs: Box<dyn Iterator<Item = Symbol<'data>> + 'object>,
+    funcs: std::vec::IntoIter<Symbol<'data>>,
+    _marker: std::marker::PhantomData<&'object u8>,
 }
 
 impl<'data, 'object> Iterator for WasmSymbolIterator<'data, 'object> {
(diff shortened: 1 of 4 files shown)


symbolic-debuginfo/src/wasm.rs
```diff
@@ -292,67 +258,39 @@ impl<'data: 'object, 'object> ObjectLike<'data, 'object> for WasmObject<'data> {
...
 }
 
 /// An iterator over symbols in the WASM file.
 ///
 /// Returned by [`WasmObject::symbols`](struct.WasmObject.html#method.symbols).
-pub struct WasmSymbolIterator<'data, 'object> {
-    funcs: std::iter::Peekable<Box<dyn Iterator<Item = &'object walrus::Function> + 'object>>,
-    _marker: std::marker::PhantomData<&'data [u8]>,
+pub struct WasmSymbolIterator<'data> {
+    funcs: Vec<Symbol<'data>>,
```
Swatinem, line 287: or maybe use a `std::vec::IntoIter` so you can avoid the `reverse()`.

Also, not quite sure if removing the lifelite parameter is considered a breaking change? If so, maybe add a `PhantomData`.

symbolic-debuginfo/src/wasm/parser.rs
```diff
@@ -0,0 +1,252 @@
...
+                funcs[body_index].size = size;
+
+                // Though we have an accurate? size of the function body, the old method of symbol
+                // iterating with walrus extends the size of each body to be contiguous with the
+                // next function, so we do the same, other than the final function
+                if body_index > 0 {
+                    funcs[body_index - 1].size = address - funcs[body_index - 1].address;
+                }
+
+                body_index += 1;
```
Swatinem, line 122: so if I understand correctly, the `FunctionSection` gives you the total number of sections? And then each `CodeSectionEntry` fills in the details?

I would rather just `push` here, and use `if let Some() = funcs.last()` for the check above.

Or is the problem rather that you could potentially have a `CustomSection("name")` that wants to fill out the names before we actually get to the `CodeSectionEntry`?

This feels a bit brittle, and you can easily panic if you create a wasm file without a `FunctionSection` (or one with a 0 count), but with a bunch of `CodeSectionEntry`s.

symbolic-debuginfo/src/wasm/parser.rs
```diff
@@ -0,0 +1,252 @@
...
+                                    let fname = map.read()?;
+
+                                    // The names for imported functions are also in this table, but
+                                    // we don't care about them
+                                    if fname.index >= num_imported_funcs {
+                                        if let Some(func) = funcs
+                                            .get_mut((fname.index - num_imported_funcs) as usize)
+                                        {
+                                            func.name =
+                                                dbg!(Some(std::borrow::Cow::Borrowed(fname.name)));
```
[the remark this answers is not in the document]
Jake-Shadle: Oops

symbolic-debuginfo/src/wasm.rs
Jake-Shadle: Not sure what you mean, how would into_iter() avoid the `reverse()`?

symbolic-debuginfo/src/wasm/parser.rs
Jake-Shadle: Yes, the function section gives the total number of locally defined functions in the module, I believe it always has to come before the actual code section where the function bodies reside, but I can change it so it won't panic if that is the case. I also don't think the `name` custom section can come before the code section, but I'm not certain. If there was a case where there is a code section entry without a valid function section before it with a matching count of local functions, the wasm validation will fail before we get to the indexing, but I can change that just in case.

symbolic-debuginfo/src/wasm.rs
Swatinem: I think `IntoIter` yields from the front without needing to move data around.

Jake-Shadle: Sorry I was uncaffeinated at the time, I've preserved the original API now.

symbolic-debuginfo/src/wasm.rs
```diff
@@ -273,14 +273,14 @@ impl<'data> Dwarf<'data> for WasmObject<'data> {
 /// An iterator over symbols in the WASM file.
 ///
 /// Returned by [`WasmObject::symbols`](struct.WasmObject.html#method.symbols).
-pub struct WasmSymbolIterator<'data> {
-    funcs: Vec<Symbol<'data>>,
+pub struct WasmSymbolIterator<'data, 'object> {
+    funcs: Box<dyn Iterator<Item = Symbol<'data>> + 'object>,
```
Swatinem, line 277: I don’t think there is a need to box/dyn this. I think `std::vec::IntoIter` should work.

Swatinem (approved):

symbolic-debuginfo/src/wasm.rs
Jake-Shadle: Oh right, I never write code like this. :sweat:

Swatinem:
Awesome stuff!

Issue #477, opened 2022-01-04
Extremely slow demangling of malformed symbol using excessive memory


5225225:
Test case:

```rust
fn main() {
    symbolic::demangle::demangle("_ZUlzjjlZZL1zStUlSt7j_Z3kjIIjIjL1vfIIEEEjzjjfjzSt7j_Z3kjIIjfjzL4t3kjIIjfjtUlSt7j_Z3kjIIjIjL1vfIIEEEjzjjfjzSt7j_Z3kjIIjfjzL4t3kjIIjfjzL4t7IjIjjzjjzSt7j_Z3kjIIjfjzStfjzSt7j_ZA3kjIIjIjL1vfIIEEEjzjjfjzSt7j_Z3kjIIjIjL1vfIIEEEjzjjfjzSt7j_Z3kjIIjfjzL4t3kjIIjzL4t7IjIjjzjjzSt7j_Z3kjIIjfjzStfjzSt7j_ZA3kjIIjIjL1vfIIEEEjzjjfjzSt7j_Z3kjIIjIjL1vfIIEEEjzjjfjzSt7j_Z3kjIIjfjzL4t3kjIIjfjzL4t7IjIjL1vfIIEEEjzjjSI");
}
```

This is *probably* a cpp_demangle issue, but I can't reproduce this there. At least, that's where the stack trace led.

When fuzzing, this hits a memory limit and then stops. When running normally, it does complete successfully, but uses a few gigabytes of RAM on the way, and I assume it can be easily crafted to OOM any normal system, though I haven't tried it.

Swatinem:
Indeed, this is not an issue upstream. I suspect because upstream has some tighter limits around recursion (both parsing as well as pretty printing). We had to increase those limits to support more complex (but still valid) mangled symbols.

As you say that it blows up memory usage, I suspect its rather a matter of either adding a potentially huge amount of substitutions, or blowing past some limit when pretty printing.
Probably the latter, which I think we can work around by just bounding the output string.

Swatinem:
Interesting. The problem does not seem to be the pretty printing, but rather some part of the parsing. Even when I reset the recursion depth back to the default of `cpp_demangle`, the tests runs for ~15 seconds on my system and uses a ton of memory, whereas the upstream test I added in https://github.com/gimli-rs/cpp_demangle/pull/236 runs essentially instantly.

So this does need some more investigation. While #481 does not fix this particular issue, IMO it is still worth merging and bounding the length of demangled symbols.

[...]

PR #481, opened 2022-01-04
fix: Introduce bounded demangling in an attempt to fix unbounded memo…

Swatinem:
…ry growth

An attempt to help with #477, however that is not solved yet.


flub (commented):
iiuc this will still write as much as it likes into the BoundedString, only on using `Display` or `ToString` traits of the string will it not print the entire thing.  That doesn't seem enough to fix the unbounded-ness of the allocation?

Swatinem:
> this will still write as much as it likes into the BoundedString, only on using `Display` or `ToString` traits of the string will it not print the entire thing

I don’t really understand. No `Display` or `ToString` is involved here.

flub:
oh wait, i totally misread `std::fmt::Write` :/

flub (approved):

PR #482, opened 2022-01-04
fix: Fix test assertion on windows

Swatinem:
On Windows, Files can only be unlinked if there are no active handles to it.


relaxolotl (approved):
i don't see tests being run on windows in CI for this repository; is this the only thing that's blocking us from doing so?

PR #484, opened 2022-01-05
Replace memmap

Jake-Shadle:
`memmap` has been marked as [unmaintained](https://rustsec.org/advisories/RUSTSEC-2020-0077) for over a year now, this just replaces it with the maintained `memmap2` fork.

### DIFF
--- a/symbolic-common/Cargo.toml
+++ b/symbolic-common/Cargo.toml
@@ -22,7 +22,7 @@ all-features = true
 
 [dependencies]
 debugid = "0.7.1"
-memmap = "0.7.0"
+memmap2 = "0.5.0"
 stable_deref_trait = "1.1.1"
 serde_ = { package = "serde", version = "1.0.88", optional = true, features = ["derive"] }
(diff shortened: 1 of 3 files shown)


Swatinem (approved):
Is there some discussion I can read through regarding this is unmaintained/superceded?

Jake-Shadle:
The advisory links to the original issue https://github.com/danburkert/memmap-rs/issues/90 where multiple people tried to get in contact with the maintainer.

Swatinem:
Thanks for clarifying! I guess its time to add automatic checking for this in CI, looks like we only have that for the rust SDK.

extrawurst:
glad this was finally tackled. interested what has changed since: https://github.com/getsentry/symbolic/issues/304

Swatinem:
I think if it gets flagged via RUSTSEC advisories, that is a bit of an incentive to do the transition. Although the advisories can be a bit overeager sometimes.

extrawurst:
when can we get a release with this? :)

PR #486, opened 2022-01-07
ci: Run a weekly cargo-audit job

Swatinem:


.github/workflows/weekly.yml
```diff
@@ -27,8 +29,7 @@ jobs:
           override: true
           components: clippy
 
-      - name: Run clippy ${{ matrix.rust }}
```
relaxolotl, line 30: any reason why we can't keep this name?

.github/workflows/weekly.yml
```diff
@@ -37,3 +38,14 @@ jobs:
         with:
           command: test
           args: --workspace --all-features
+
+  weekly-audit:
+    name: Audit
```
relaxolotl, line 43: any reason why this can't be "security audit"?

relaxolotl (approved):
Something to keep in mind is that `actions-rs/audit-check` hasn't seen much activity from its author for over a year (last release was May 2020). If the default settings work well enough for our needs then this shouldn't be a cause for concern.

Swatinem:
I just kept the job and naming the same as with the sentry-rust repo.

Issue #490, opened 2022-01-12
WASM instruction addresses are incorrect


Jake-Shadle:
After my change in #474 it seems that the instruction addresses are wrong in many cases. Unfortunately the current test case for wasm is too simplistic to catch the error, but I immediately noticed it when symbolicating a stack trace from one of our own modules. I'm just filing this bug for visibility, I plan on fixing this tomorrow (hopefully).

PR #491, opened 2022-01-13
Fix WASM symbolication

Jake-Shadle:
Well, this took me longer than I would have liked to figure out, but I _think_ this resolves an issue with (malformed?) DWARF debug info and `SymCacheConverter`.

For some background, I noticed that when sending wasm stacktraces to sentry.io, the first frame in a very short stacktrace of < 5 frames was always misattributed, but other frames were correct. At first I assumed that I had messed up the WASM parsing in #474 somehow, but when looking at the addresses/symbols reported by [`wasmtime::Trap`](https://docs.rs/wasmtime/latest/wasmtime/struct.Trap.html) and the information parsed by `WasmObject` they were almost the same, other than that both wasmtime and DWARF use the "true" function start address, whereas in the previous code and the refactor it was using the address after the function locals, which this PR fixes.

Thinking that this was maybe the reason for the misattribution, I figured I would sanity check by making a test with `SymbolCache` which I assume is what sentry is using for symbolication server side, by comparing the stacktrace reported by wasmtime, which I know is correct, versus the one spit out by the `SymbolCache`. However, again, the first frame was always misattributed.

Eventually, after adding a bunch of debugging code and trying different things, I noticed this seemingly innocuous line, https://github.com/getsentry/symbolic/blob/d05ca1662b64eced395bc65230ba87acf3120d71/symbolic-symcache/src/new/writer.rs#L175. And indeed, when adding a check for function addresses `> u32::MAX`, there were indeed _many_ of them. These addresses aren't coming from the WASM addresses that were being parsed, but rather from the DWARF debug info, and ends up being a "last one wins" situation where large addresses wrap around and stomp over lower addresses already in the list. In our scenario, the first frame in the stack trace just so happened to also be the first function in the WASM binary, with the lowest possible addresses, and _many_ DWARF functions with large addresses overwrote it.

Anyway, long story short, I added an additional check in `SymCacheConverter::process_symbolic_function` to skip functions that are larger than `u32::MAX` as a quick workaround, which means this particular WASM stacktrace is now symbolicated correctly.

I also happened to notice a copy-paste error in `DwarfUnit::functions` and removed it.

Resolves: #490

### DIFF
--- a/symbolic-debuginfo/src/dwarf.rs
+++ b/symbolic-debuginfo/src/dwarf.rs
@@ -814,12 +814,6 @@ impl<'d, 'a> DwarfUnit<'d, 'a> {
             // should avoid this problem.
             range_buf.sort_by_key(|r| r.begin);
 
-            // In WASM files emitted by emscripted, we have observed a variety of broken ranges.
-            // One of these cases also involves ranges which are not being sorted, resulting in
-            // arithmetic underflow calculating `function_size` (in debug builds). Sorting the ranges
-            // should avoid this problem.
-            range_buf.sort_by_key(|r| r.begin);
-
             let function_address = offset(range_buf[0].begin, self.inner.info.address_offset);
             let function_size = range_buf[range_buf.len() - 1].end - range_buf[0].begin;
             let function_end = function_address + function_size;
--- a/symbolic-symcache/src/new/writer.rs
+++ b/symbolic-symcache/src/new/writer.rs
@@ -162,8 +162,8 @@ impl SymCacheConverter {
     }
 
     pub fn process_symbolic_function(&mut self, function: &Function<'_>) {
-        // skip over empty functions
-        if function.size == 0 {
+        // skip over empty functions or functions whose address is too large to fit in a u32
+        if function.size == 0 || function.address > u32::MAX as u64 {
             return;
         }
 
--- a/symbolic-debuginfo/src/wasm/parser.rs
+++ b/symbolic-debuginfo/src/wasm/parser.rs
@@ -221,7 +221,7 @@ fn get_function_info(
 ) -> Result<(u64, u64), WasmError> {
     let mut body = body.get_binary_reader();
 
-    let real_start = body.original_position();
+    let function_address = body.original_position() as u64;
 
     // locals, we _can_ just skip this, but might as well validate while we're here
     {
@@ -231,12 +231,6 @@ fn get_function_info(
             let ty = body.read_type()?;
             validator.define_locals(pos, count, ty)?;
         }
-    }
-
-    let function_address = body.original_position() as u64;
-
-    if function_address == 632 {
-        dbg!(real_start);
     }
 
     while !body.eof() {
--- a/symbolic-debuginfo/src/wasm/parser.rs
+++ b/symbolic-debuginfo/src/wasm/parser.rs
@@ -221,6 +221,8 @@ fn get_function_info(
 ) -> Result<(u64, u64), WasmError> {
     let mut body = body.get_binary_reader();
 
+    let real_start = body.original_position();
+
     // locals, we _can_ just skip this, but might as well validate while we're here
     {
         for _ in 0..body.read_var_u32()? {
@@ -232,6 +234,10 @@ fn get_function_info(
     }
 
     let function_address = body.original_position() as u64;
+
+    if function_address == 632 {
+        dbg!(real_start);
+    }
 
     while !body.eof() {
         let pos = body.original_position();
--- a/symbolic-debuginfo/tests/snapshots/test_objects__wasm_symbols.snap
+++ b/symbolic-debuginfo/tests/snapshots/test_objects__wasm_symbols.snap
@@ -1,6 +1,7 @@
 ---
 source: symbolic-debuginfo/tests/test_objects.rs
 expression: SymbolsDebug(&symbols)
+
 ---
-              8c internal_func
+              8b internal_func
 


symbolic-debuginfo/src/dwarf.rs
```diff
@@ -814,12 +814,6 @@ impl<'d, 'a> DwarfUnit<'d, 'a> {
             // should avoid this problem.
             range_buf.sort_by_key(|r| r.begin);
 
-            // In WASM files emitted by emscripted, we have observed a variety of broken ranges.
-            // One of these cases also involves ranges which are not being sorted, resulting in
-            // arithmetic underflow calculating `function_size` (in debug builds). Sorting the ranges
-            // should avoid this problem.
-            range_buf.sort_by_key(|r| r.begin);
```
Swatinem, line 821: :-D nice  catch. Must have been a wrong merge or something, lol.

symbolic-symcache/src/new/writer.rs
```diff
@@ -162,8 +162,8 @@ impl SymCacheConverter {
     }
 
     pub fn process_symbolic_function(&mut self, function: &Function<'_>) {
-        // skip over empty functions
-        if function.size == 0 {
+        // skip over empty functions or functions whose address is too large to fit in a u32
+        if function.size == 0 || function.address > u32::MAX as u64 {
```
Swatinem, line 166: I wonder why that is happening. Is your wasm file >4G (which I doubt)?
Or is there an underflow happening in a different place, maybe:

https://github.com/getsentry/symbolic/blob/d05ca1662b64eced395bc65230ba87acf3120d71/symbolic-debuginfo/src/dwarf.rs#L65-L71

CC @jan-auer since I brought up my concerns about a potential underflow in `offset` previously.

This is just a hunch now. But I was arguing to skip/ignore any addr that underflows `offset` right in the `symbolic-debuginfo` abstraction, instead of propagating that into deeper in the processing layer where it could potentially create more confusion.

Swatinem (approved):
lgtm for now, though I would love to find out where that overflow/undeflow is coming from.

symbolic-symcache/src/new/writer.rs
Jake-Shadle: So it looks like it might indeed be bad DWARF info, as it seems like there are many functions that are attributed to `u32::MAX` according to 

https://github.com/getsentry/symbolic/blob/d05ca1662b64eced395bc65230ba87acf3120d71/symbolic-debuginfo/src/dwarf.rs#L547-L548

Jake-Shadle: eg this is what it prints out when parsing the dwarf ranges for this WASM module

```
symbolic-debuginfo/src/dwarf.rs:550] addr = 4294967295
[symbolic-debuginfo/src/dwarf.rs:550] addr = 4294967295
[symbolic-debuginfo/src/dwarf.rs:550] addr = 4294967295
[symbolic-debuginfo/src/dwarf.rs:550] addr = 4294967295
[symbolic-debuginfo/src/dwarf.rs:550] addr = 4294967295
[symbolic-debuginfo/src/dwarf.rs:550] addr = 4294967295
[symbolic-debuginfo/src/dwarf.rs:550] addr = 4294967295
[symbolic-debuginfo/src/dwarf.rs:550] addr = 4294967295
[symbolic-debuginfo/src/dwarf.rs:550] addr = 4294967295
[symbolic-debuginfo/src/dwarf.rs:550] addr = 4294967295
[symbolic-debuginfo/src/dwarf.rs:550] addr = 4294967295
[symbolic-debuginfo/src/dwarf.rs:550] addr = 4294967295
[symbolic-debuginfo/src/dwarf.rs:550] addr = 4294967295
[symbolic-debuginfo/src/dwarf.rs:550] addr = 4294967295
[symbolic-debuginfo/src/dwarf.rs:550] addr = 28390
[symbolic-debuginfo/src/dwarf.rs:550] addr = 4294967295
[symbolic-debuginfo/src/dwarf.rs:550] addr = 4294967295
[symbolic-debuginfo/src/dwarf.rs:550] addr = 28288
[symbolic-debuginfo/src/dwarf.rs:550] addr = 28299
[symbolic-debuginfo/src/dwarf.rs:550] addr = 28310
[symbolic-debuginfo/src/dwarf.rs:550] addr = 28318
[symbolic-debuginfo/src/dwarf.rs:550] addr = 28318
[symbolic-debuginfo/src/dwarf.rs:550] addr = 28318
[symbolic-debuginfo/src/dwarf.rs:550] addr = 28318
[symbolic-debuginfo/src/dwarf.rs:550] addr = 28325
[symbolic-debuginfo/src/dwarf.rs:550] addr = 28334
[symbolic-debuginfo/src/dwarf.rs:550] addr = 28334
[symbolic-debuginfo/src/dwarf.rs:550] addr = 28339
[symbolic-debuginfo/src/dwarf.rs:550] addr = 28365
[symbolic-debuginfo/src/dwarf.rs:550] addr = 28379
[symbolic-debuginfo/src/dwarf.rs:550] addr = 28379
[symbolic-debuginfo/src/dwarf.rs:550] addr = 4294967295
[symbolic-debuginfo/src/dwarf.rs:550] addr = 4294967295
[symbolic-debuginfo/src/dwarf.rs:550] addr = 4294967295
[symbolic-debuginfo/src/dwarf.rs:550] addr = 4294967295
[symbolic-debuginfo/src/dwarf.rs:550] addr = 4294967295
[symbolic-debuginfo/src/dwarf.rs:550] addr = 4294967295
[symbolic-debuginfo/src/dwarf.rs:550] addr = 28219
[symbolic-debuginfo/src/dwarf.rs:550] addr = 28227
[symbolic-debuginfo/src/dwarf.rs:550] addr = 28227
```

So some addresses are fine and valid, but they are interspersed with many that are either wrong in the DWARF itself or being parsed incorrectly and/or truncated.

Swatinem: Interesting. I would be curious to look at the raw DWARF here. Is the debug file you are trying public by any means? Otherwise you can also link to a sentry debug file directly and I can take a look.

Jake-Shadle: Sure, here's the sentry link https://sentry.io/api/0/projects/embark-studios/ark-test/files/dsyms/?id=177326653

Jake-Shadle: Actually, I'm wondering if maybe this is actually normal behavior for DWARF, it's just that for WASM binaries the address offset is negative unlike for elf/macho, which results in these `u32::MAX` addresses increasing past `u32::MAX`, rather than in the other cases where it either stays the same or decreases.

Swatinem: Interesting, so `llvm-dwarfdump` really flags these as `DW_AT_low_pc [DW_FORM_addr]	(0xffffffff (dead code))`
Searching the llvm source for `(dead code)` led me to this:

https://github.com/llvm/llvm-project/blob/6d0be74af5555f7bc56ac72cbd98ff270fd1291b/llvm/test/tools/llvm-dwarfdump/X86/tombstone.s#L5

From the output it would seem the tombstone is `-1` (both 32bit and 64bit it seems)

The low/high pair looks like this:
```
                              DW_AT_low_pc [DW_FORM_addr]	(0xffffffff (dead code))
                              DW_AT_high_pc [DW_FORM_data4]	(0x00000019)
```

So I think we will have some arithmetic issues down the road anyway because start > end in that case.

Either way, given all this I think its fair to just skip anything that is `> u32::MAX` for now, and figure out how to deal with these cases later on.

Swatinem: This is also an interesting snippet:

https://github.com/llvm/llvm-project/blob/8036cf7f5402ea7fc8564a9a2beae512c324bf3d/llvm/lib/DebugInfo/DWARF/DWARFDebugRangeList.cpp#L73-L75

and ofc:

https://github.com/llvm/llvm-project/blob/f677413071a2df10fbb68050624696fd4d4ae1cd/llvm/include/llvm/BinaryFormat/Dwarf.h#L772-L774

Jake-Shadle: Nice investigation! Just curious, but what is the deployment process like for symcache into sentry, particularly sentry.io? Most of the other changes I have made recently we can just use a patch for, but this particular issue manifests on the server.

Swatinem: Its a bit complicated ;-)

We first need to pull the dependency update into symbolicator.
Then we deploy that. But then the fix is only being applied to *new* debug files. Old caches live for as long as they are in use. We have the ability via cache versioning to refresh these in the background, but that is a rather heavy hammer as it is all (symcaches) or nothing. We certainly don’t want to do that too frequently, and I’m not sure this fix is worth regenerating all caches for.

PR #495, opened 2022-01-17
fix: Try replacing chrono usage in symbolic-unreal

Swatinem:
Well, this only goes halfway, since some dependencies still rely on chrono/time. The code changes necessary are a bit annoying though :-(


flub (approved):

loewenheim (approved):

Swatinem:
```
Failed to load coverage: Truncated coverage data
```

meh, maybe using `cargo-llvm-cov` might still be better than `fucov`.

PR #518, opened 2022-03-17
meta(gha): Deploy workflow enforce-license-compliance.yml

chadwhitacre:
I are a bot, here to deploy [enforce-license-compliance.yml](https://github.com/getsentry/.github/blob/36d1230813c1e3344703382a744d151ba2d4b508/.github/workflows/enforce-license-compliance.yml). 🤖


chadwhitacre:
On second thought, deciding to implement better error messaging in pilot repos before bringing here.

PR #523, opened 2022-03-29
fix: Match the complete BcSymbolMap header (NATIVE-281)

Swatinem:
Previously, the match was truncated to the input buffer length, which meant that
empty files always passed the check, which they obviously should not.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ style-rust:
 .PHONY: style-rust
 
 style-python: .venv/bin/python
-	.venv/bin/pip install -U black==22.1.0
+	.venv/bin/pip install -U black==22.3.0
 	.venv/bin/black --check py --exclude 'symbolic/_lowlevel*|dist|build|\.eggs'
 
 # Linting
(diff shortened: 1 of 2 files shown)


loewenheim (approved):

flub (commented):
I can't request changes anymore... but I fear this breaks things

Swatinem:
Oh what I meant to refer to was this:

https://github.com/getsentry/symbolic/blob/10ccd262daf7ba6038bb15e6d1ce9079362f44cd/symbolic-debuginfo/src/object.rs#L127-L129

Symbolic needs at least 16 bytes to actually detect anything. It does not truncate, only checks for that.

Other longer file headers are also checked for precisely:

https://github.com/getsentry/symbolic/blob/10ccd262daf7ba6038bb15e6d1ce9079362f44cd/symbolic-debuginfo/src/pdb.rs#L127

flub:
hum, did some spelunking in the sentry-cli codebase as well and I can't find a reason for the requirement to be able to do partial matches, though I'm like 100% sure this was a requirement and that this was done explicitly like this before.  I think these API's are a bit meh as they are, they require a "large enough" slice but you're left not knowing what "large enough" means is as a caller.  So whatever, I guess it's fine, thanks for having a look as well.

Swatinem:
Hm, maybe we should have a look at the `sentry` codebase itself, as it pulls in parts of the symbolic python API.

flub:
good point, though it seems the test methods aren't even exposed on the python API and sentry re-implements this: https://github.com/getsentry/sentry/blob/master/src/sentry/models/debugfile.py#L421-L433

Swatinem:
👍🏻 thanks for looking into this, in which case I think we are good.

PR #526, opened 2022-03-31
Experimental il2cpp code

flub:
This is not meant to be stable in any way.

### DIFF
--- a/symbolic/src/lib.rs
+++ b/symbolic/src/lib.rs
@@ -73,7 +73,7 @@ pub use symbolic_debuginfo as debuginfo;
 #[doc(inline)]
 #[cfg(feature = "demangle")]
 pub use symbolic_demangle as demangle;
-#[doc(inline)]
+#[doc(inline, hidden)]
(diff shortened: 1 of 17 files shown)


symbolic/src/lib.rs
```diff
@@ -73,6 +73,9 @@ pub use symbolic_debuginfo as debuginfo;
 #[doc(inline)]
 #[cfg(feature = "demangle")]
 pub use symbolic_demangle as demangle;
+#[doc(inline, hidden)]
```
flub, line 76: @Swatinem I've marked this as hidden as well as documented it in the module as do not use.  I think this should be sufficient.

relaxolotl (approved):
only really skimmed this, approving because this is meant to be experimental and very clearly marks itself as such

Swatinem (approved):
A stronger callout, as in naming the feature `unstable_dontuse_il2cpp`, and maybe even introduce the feature to `symbolic-il2cpp` as well so the crate is completely empty without that feature, might be preferable, but I guess its fine either way :-D

PR #529, opened 2022-04-02
ref(usym): Use existing types instead of primitives for fields on UsymSymbols

relaxolotl:
We have types that represent debug IDs and architectures, so instead of returning plain strings for those UsymSymbols now returns those types. 

For the time being, those two fields are lazily converted into their respective types instead of during the parsing phase. I'm working on a different PR that explores the idea of leniently parsing as much of a usym file as possible which will explore the idea of aggressively converting these fields during parsing.

One awkward thing about this new API is that there's currently no way to access the raw debug ID or architecture if they fail to parse. If we feel that it's important to expose these I can explore adding getters for them. They probably won't be particularly useful, though.

### DIFF
--- a/symbolic-il2cpp/src/usym.rs
+++ b/symbolic-il2cpp/src/usym.rs
@@ -318,8 +318,8 @@ impl<'a> UsymSymbols<'a> {
     /// The ID of the assembly.
     ///
     /// This should match the ID of the debug symbols.
-    pub fn id(&self) -> Option<DebugId> {
-        DebugId::from_str(self.id).ok()
+    pub fn id(&self) -> Result<DebugId, UsymError> {
+        DebugId::from_str(self.id).map_err(|e| UsymError::new(UsymErrorKind::BadId, e))
     }
 
     /// The name of the assembly.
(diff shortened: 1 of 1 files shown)


loewenheim (approved):

symbolic-il2cpp/src/usym.rs
```diff
@@ -317,9 +320,8 @@ impl<'a> UsymSymbols<'a> {
     /// The ID of the assembly.
     ///
     /// This should match the ID of the debug symbols.
-    // TODO: Consider making this return debugid::DebugId
-    pub fn id(&self) -> &str {
-        self.id
+    pub fn id(&self) -> Option<DebugId> {
```
flub, line 323: You could do something like return `Result<DebugId, &str>` so you still return the string error if it could not be parsed.  If choosing that way probably should do the same for `arch`

flub (approved):

symbolic-il2cpp/src/usym.rs
relaxolotl: it is done 🎉

PR #536, opened 2022-04-14
docs(usym): Make documentation about the SymCache binary public

relaxolotl:
This moves documentation about the structure and the contents of a SymCache binary into a public location, so there's something available that details the general structure of a symcache and its contents. Previously this wasn't accessible anywhere except in the source code itself.

Run `cargo docs --open -p symbolic-symcache` to preview what the docs will look like.


symbolic-symcache/src/lib.rs
```diff
@@ -1,4 +1,101 @@
-//! Provides SymCache support.
+//! Provides SymCache support. This includes a reader and writer for the binary format, as well as
+//! helper traits and functions to apply transformations to debugging symbols before they are
+//! written to the SymCache.
```
flub, line 3: nit: it is preferable to have a single-line summary first just like for git commits.  The generated docs will use the first line in the module overview parts and this is much nicer to read if its not the start of a paragraph.

symbolic-symcache/src/lib.rs
```diff
@@ -1,4 +1,101 @@
...
+//! # Structure of a SymCache
+//!
+//! A SymCache (version 7) contains the following primary kinds of data, written in the following
+//! order:
+//!
+//! 1. Files
+//! 2. Functions
+//! 3. Source Locations
+//! 4. Address Ranges
+//! 5. String Data
```
flub, line 14: All these parts are detailed below, only this one isn't.  I don't consider this blocking though, as this was already the case and this is just moving things around.  But if you feel like doing a followup PR to fix that... 😉

flub (approved):

Swatinem (approved):

symbolic-symcache/src/lib.rs
relaxolotl: i'll see what i can do about adding a section for string data in a followup, especially since i added this entry in to begin with 👍

PR #539, opened 2022-04-14
ref(usym): Add test with malformed data

relaxolotl:
There's really no reason why we can't handle a scenario where we have non-UTF8 strings, but I'm just going to leave a test and defer doing so until later.


symbolic-il2cpp/src/usym.rs
```diff
@@ -610,6 +610,21 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_header_with_errors() {
+        let file = File::open(fixture("il2cpp/artificial-bad-meta.usym")).unwrap();
+        let data = ByteView::map_file_ref(&file).unwrap();
```
Swatinem, line 616:
```suggestion
        let data = ByteView::open(fixture("il2cpp/artificial-bad-meta.usym")).unwrap();
```

;-)

Swatinem (approved):

symbolic-il2cpp/src/usym.rs
relaxolotl: oh, thanks for this! this goes to show up much copy-pasting i've been doing 🙃

PR #541, opened 2022-04-19
ref(symcache): Deduplicate functions during insertion into a symcache

relaxolotl:
A spinoff of the discussion started in https://github.com/getsentry/symbolic/pull/540#discussion_r850070820. The idea is that we're currently not deduplicating function entries in the symcache, and maybe we should. It's possible to have two entries in the symcache for the same function where the only difference between the two is their `entry_pc` which isn't quite correct.

Inline entries appear to be the biggest offenders of this, based on the test data in symbolic. The changes to the tests in the PR demonstrate this by having a vast majority of their pruned functions be of an inlined flavour. This doesn't appear to be too much of an issue for non-inlined functions as they appear to be already parsed and deduplicated by the time the symcache writer is working with them. For usyms, this isn't so much the case.

usyms offer minimal support to identify addresses/lines/source locations associated with some common function, only really providing a list of "source records" mapped to addresses and function identifiers. The format also provides zero guarantees about functions being contiguous blocks of entries, e.g. function a might have 3 back-to-back source locations (slocs) followed by 1 sloc for function b, then followed by more slocs for function a. Some deduplication logic when inserting functions into the symcache would simplify the converter logic, and cut out the work to properly group usym-parsed records by their functions.

It also looks like the tests were omitting inlined functions from their snapshots because they introduced a large amount of noise into the final results, presumably because there could be multiple duplicate entries for one inlined function. I'm adding these functions back into the snapshots because these are being written to the symcache, so it would be good to start testing and verifying any changes that would impact inlined function entries in the symcache.


symbolic-symcache/src/new/raw.rs
```diff
@@ -67,6 +68,19 @@ pub struct Function {
     pub lang: u32,
 }
 
+impl PartialEq for Function {
+    fn eq(&self, other: &Self) -> bool {
+        self.name_offset == other.name_offset && self.comp_dir_offset == other.comp_dir_offset
```
relaxolotl, line 73: `lang` seems like something that might be important to include here, in case a function somehow has some parts in say, C# and other sections in C++ (looking at you, usyms). I excluded it for now just because I'm guessing the expectation is that all functions in a single symcache are usually for one common language, but I could be wrong.

symbolic-symcache/tests/test_writer.rs
```diff
@@ -66,7 +52,7 @@ fn test_write_header_linux() -> Result<(), Error> {
         },
         arch: Amd64,
         files: 55,
-        functions: 697,
+        functions: 695,
```
relaxolotl, line 55: `test_write_header_linux` and `test_write_functions_linux` source their data from the same fixture, so I'll include a list of removed functions here.

Here's the two axed functions:
```
ffffffff _ZN15google_breakpad13PageAllocator5AllocEm
ffffffff _ZNK15google_breakpad17LinuxPtraceDumper13BuildProcPathEPciPKc
```
In the snapshot, these have entries with addresses that aren't set to the inline func canary address.

symbolic-symcache/tests/test_writer.rs
```diff
@@ -106,7 +92,7 @@ fn test_write_header_macos() -> Result<(), Error> {
         },
         arch: Amd64,
         files: 36,
-        functions: 639,
+        functions: 587,
```
relaxolotl, line 95: Likewise, `test_write_header_macos` and `test_write_functions_macos` also source their data from the same file. Here's a much longer list covering all 52 missing functions:
```
ffffffff _ZN15google_breakpad10scoped_ptrI9sigactionED1Ev
ffffffff _ZN15google_breakpad10scoped_ptrINS_21CrashGenerationClientEED1Ev
ffffffff _ZN15google_breakpad11MachMessage13CalculateSizeEv
ffffffff _ZN15google_breakpad11MachMessage13GetDataPacketEv
ffffffff _ZN15google_breakpad11MachMessage13GetDescriptorEi
ffffffff _ZN15google_breakpad11MachMessage13SetDescriptorEiRKNS_21MachMsgPortDescriptorE
ffffffff _ZN15google_breakpad11MachMessage18SetDescriptorCountEi
ffffffff _ZN15google_breakpad11MachMessage7SetDataEPvi
ffffffff _ZN15google_breakpad11MachMessageC2Ev
ffffffff _ZN15google_breakpad11ReceivePortC2EPKc
ffffffff _ZN15google_breakpad11ReceivePortC2Ej
ffffffff _ZN15google_breakpad11ReceivePortC2Ev
ffffffff _ZN15google_breakpad11ReceivePortD2Ev
ffffffff _ZN15google_breakpad12DynamicImage11GetFileTypeEv
ffffffff _ZN15google_breakpad13DynamicImages20ReadImageInfoForTaskEv
ffffffff _ZN15google_breakpad13DynamicImages23GetExecutableImageIndexEv
ffffffff _ZN15google_breakpad13DynamicImages27GetDyldAllImageInfosPointerEv
ffffffff _ZN15google_breakpad14MachPortSenderC2EPKc
ffffffff _ZN15google_breakpad14MachPortSenderC2Ej
ffffffff _ZN15google_breakpad15MachSendMessageC2Ei
ffffffff _ZN15google_breakpad16ExceptionHandler13ResumeThreadsEv
ffffffff _ZN15google_breakpad16ExceptionHandler14SuspendThreadsEv
ffffffff _ZN15google_breakpad16ExceptionHandler26SendMessageToHandlerThreadENS_20HandlerThreadMessageE
ffffffff _ZN15google_breakpad16ExceptionHandler8TeardownEv
ffffffff _ZN15google_breakpad16ExceptionHandlerC1ERKNSt3__112basic_stringIcNS1_11char_traitsIcEENS1_9allocatorIcEEEEPFbPvEPFbPKcSE_SA_bESA_bSE_
ffffffff _ZN15google_breakpad16ExceptionHandlerD1Ev
ffffffff _ZN15google_breakpad17MinidumpGenerator10WriteStackEPjP18MDMemoryDescriptor
ffffffff _ZN15google_breakpad17MinidumpGenerator12WriteContextEPjP20MDLocationDescriptor
ffffffff _ZN15google_breakpad17MinidumpGenerator13WriteStackX86EPjP18MDMemoryDescriptor
ffffffff _ZN15google_breakpad17MinidumpGenerator14GetThreadStateEjPjS1_
ffffffff _ZN15google_breakpad17MinidumpGenerator16WriteStackX86_64EPjP18MDMemoryDescriptor
ffffffff _ZN15google_breakpad17MinidumpGenerator17CurrentPCForStackEPj
ffffffff _ZN15google_breakpad17MinidumpGenerator20CurrentPCForStackX86EPj
ffffffff _ZN15google_breakpad17MinidumpGenerator20FindExecutableModuleEv
ffffffff _ZN15google_breakpad17MinidumpGenerator23CurrentPCForStackX86_64EPj
ffffffff _ZN15google_breakpad17MinidumpGeneratorD1Ev
ffffffff _ZN15google_breakpad18MinidumpFileWriter4CopyEjPKvl
ffffffff _ZN15google_breakpad18MinidumpFileWriter5CloseEv
ffffffff _ZN15google_breakpad18MinidumpFileWriter8AllocateEm
ffffffff _ZN15google_breakpad18MinidumpFileWriterC2Ev
ffffffff _ZN15google_breakpad18MinidumpFileWriterD2Ev
ffffffff _ZN15google_breakpad19breakpad_exc_serverEP17mach_msg_header_tS1_
ffffffff _ZN15google_breakpad6FileIDC2EPKc
ffffffff _ZN16MacFileUtilities11MachoWalker20WalkHeader64AtOffsetEx
ffffffff _ZN16MacFileUtilities11MachoWalker9ReadBytesEPvmx
ffffffff _ZN16MacFileUtilities11MachoWalkerC2EPKcPFbPS0_P12load_commandxbPvES6_
ffffffff _ZN16MacFileUtilities11MachoWalkerC2EPvmPFbPS0_P12load_commandxbS1_ES1_
ffffffff _ZN16MacFileUtilities11MachoWalkerD2Ev
ffffffff _ZN16MacFileUtilities7MachoID6UpdateEPNS_11MachoWalkerExm
ffffffff _ZN16MacFileUtilities7MachoIDC2EPKc
ffffffff _ZN16MacFileUtilities7MachoIDC2EPKcPvm
ffffffff _ZNSt3__114__split_bufferI18MDMemoryDescriptorRN15google_breakpad16PageStdAllocatorIS1_EEEC2EmmS5_
```

If you'd like to generate these yourself, check out `master`, comment out L22-37 of this file and insert a `Some(f)` above or below that block, and run the tests for this crate. insta will generate a new snapshot based on your test results which will contain all of the functions in the symcache, including the inlined (duplicate and non-duplicate) functions.

Swatinem (changes requested):
I’m sorry to "veto" this change, let me give a more detailed explanation.

Please feel free to reproduce this use-case btw, I haven’t so far ;-)

In C, you can define multiple `static` functions, which are completely local to the compilation unit (aka C source file). They can have duplicated/overlapping names. You can have a `static` function `a` defined in file `a.c`, and a `static` function `a` defined in `other_a.c`. They have the same name and unfortunately compilation directory. Ideally, we should also save the declaration file/line of a function, because that will uniquely identify it. But we don’t for now. So its entry_pc is the only thing we have right now to tell them apart.

And we do want to have two different entries for two different but same-named functions.
Eventually at some point in the future, we would want to return function-relative instructions as well, which would be relative to the `entry_pc`. We don’t do that right now, but might as well in the future. There is even https://getsentry.atlassian.net/browse/FEEDBACK-855 filed already for this, though as we all know, jira is the place where issues go to die.

The fact that only "global" (exported) functions have unique names (essentially the linker handles these); and "local" (`static`) functions do not (the linker ignores them mostly) means that we can have functions with the same name but different entry_pc.

Another aside, there is this thing called "unity builds" in C, where you essentially concatenate different C source files together and compile them as one compilation unit, so you lower the compiler and preprocessor overhead as a way to speed up build times. In those cases you need to make sure that your functions have unique names, as otherwise the compiler would complain. Here is an example of that: https://github.com/getsentry/sentry-native/commit/f2ed3aca4c12597bb6ef71db3d874416722385b6 before that PR, we had a bunch of different `static send_envelope` functions all in different files/compilation units (and thus different `entry_pc`s).

relaxolotl:
Thanks for the thorough explanation, @Swatinem.

> Please feel free to reproduce this use-case btw, I haven’t so far ;-)

To clarify, do you mean the use case described in the PR or the use cases described in the following segments in your comment? Are the scenarios you're listing hypothetical ones that would break this PR, or..? If this is in reference to duplicate functions in usyms, the diff for the usym test snapshot currently demonstrates this: `NewBehaviour.Start()` has two entries that's collapsed down to just one `NewBehaviour.Start()` entry.

FWIW I would be happy if all this PR ended up producing was a list of tests that included these aforementioned scenarios that rely on the `entry_pc` as a discriminant, if they're not overly complicated to construct. I can add in custom logic into the usym converter instead to handle duplicates instead.

> You can have a static function a defined in file a.c, and a static function a defined in other_a.c. They have the same name and unfortunately compilation directory. Ideally, we should also save the declaration file/line of a function, because that will uniquely identify it. But we don’t for now.
 
Is the only thing stopping us from including this just a lack of time/customers writing in about it? Based on what you're saying, it seems like it'd be rather useful to track these two pieces of information, although I could be missing some context on why this isn't being done right now. 

Do you know if this would improve the accuracy of the files associated with stack trace frames as rendered on sentry and shown to end users? Or would this not really have an impact? We could make a case for adding these two in if it allows us to show more accurate filenames in our stack traces.

Swatinem:
> Is the only thing stopping us from including this just a lack of time/customers writing in about it?

I think the declaration file/line is just not exposed via our abstraction.

> Do you know if this would improve the accuracy of the files associated with stack trace frames as rendered on sentry

I don’t think it would.

> or the use cases described in the following segments in your comment? Are the scenarios you're listing hypothetical ones that would break this PR, or..?

Yes, the example you quoted. And it is not hypothetical but very real. *However*, it would break function-relative addresses. Which we don’t have yet, but there is that open jira ticket to eventually add the feature; which I expect would be quite straight forward.

relaxolotl:
I've opened up a new PR with a test that directly checks the use case brought up in the comments: #545. Given that, I'm going to close this PR as it is no longer needed.

PR #542, opened 2022-04-20
ref(symcache): Move transformer tests out to a separate file

relaxolotl:
SymCache transformers got their own dedicated module in #538, so they probably should get their own dedicated test module.

### DIFF
--- a/symbolic-symcache/tests/test_transformers.rs
+++ b/symbolic-symcache/tests/test_transformers.rs
@@ -1,0 +1,91 @@
+use std::fmt;
+use std::io::Cursor;
+
+use symbolic_common::{ByteView, SelfCell};
+use symbolic_debuginfo::macho::BcSymbolMap;
+use symbolic_debuginfo::Object;
+use symbolic_symcache::transform::{self, Transformer};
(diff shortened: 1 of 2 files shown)


symbolic-symcache/tests/test_transformers.rs
```diff
@@ -0,0 +1,91 @@
...
+        sl: transform::SourceLocation<'f>,
+    ) -> transform::SourceLocation<'f> {
+        self.0.get().transform_source_location(sl)
+    }
+}
+
+#[test]
+fn test_transformer_symbolmap() -> Result<(), Error> {
+    let buffer = ByteView::open(
+        "../symbolic-debuginfo/tests/fixtures/2d10c42f-591d-3265-b147-78ba0868073f.dwarf-hidden",
```
relaxolotl, line 61: in theory this could be moved to fixtures, but it would have to be copied. it looks like doctests reference these files, so they can't just be deleted. i'm thinking it's better just to keep this as-is to avoid any drift between the doctest data and the unit test data.

Swatinem (approved):

loewenheim (approved):

PR #544, opened 2022-04-21
ref(symcache): Transformers no longer have a 'static lifetime

relaxolotl:
While exploring ways to write tests for #543, I stumbled upon a todo which if fixed would make my life a lot easier. 

I started digging around and noticed that `Transformer`s were required to have a static lifetime: This was likely due to a suggestion made by clippy to be explicit about the lifetime on the `Box` wrapping around `Transformer`s in the symcache, and `Box`es have static lifetimes by default. This has generated some large-ish workarounds to satisfy this lifetime.

The lifetime on a `Transformer` doesn't need to be that long to begin with, so I've just simply restricted it to the lifespan of the writer it's attached to. Some shuffling needed to be done with the code instantiating these transformers to avoid issues related to data being potentially accessed in a deconstructor after it is dropped (?).

### DIFF
--- a/examples/symcache_debug/src/main.rs
+++ b/examples/symcache_debug/src/main.rs
@@ -75,24 +75,14 @@ fn execute(matches: &ArgMatches) -> Result<()> {
 
         let mut writer = SymCacheWriter::new(Cursor::new(Vec::new()))?;
 
-        if let Some(bcsymbolmap_file) = matches.value_of("bcsymbolmap_file") {
-            let bcsymbolmap_path = Path::new(bcsymbolmap_file);
-            let bcsymbolmap_buffer = ByteView::open(bcsymbolmap_path)?;
-            let bcsymbolmap =
-                OwnedBcSymbolMap(SelfCell::try_new(bcsymbolmap_buffer, |s| unsafe {
-                    BcSymbolMap::parse(&*s)
-                })?);
-            writer.add_transformer(bcsymbolmap);
+        if let Some(transformer) = bcsymbolmap_transformer {
+            writer.add_transformer(transformer);
         }
 
         #[cfg(feature = "il2cpp")]
         {
-            if let Some(linemapping_file) = matches.value_of("linemapping_file") {
-                let linemapping_path = Path::new(linemapping_file);
-                let linemapping_buffer = ByteView::open(linemapping_path)?;
-                if let Some(linemapping) = LineMapping::parse(&linemapping_buffer) {
-                    writer.add_transformer(linemapping);
-                }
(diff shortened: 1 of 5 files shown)


Swatinem (commented):
This is technically a breaking change, as you have noticed yourself, as you need to shuffle code around to create the transformer *before* the symcache writer.

The far more better thing to do would be to just remove the lifetime parameter from the `BcSymbolMap`. That too would be a breaking change though.

Turning the `OwnedBcSymbolMap` into a `BcSymbolMapView` with a `::open()` constructor would be the way to go there I would say.

flub (approved):
I think this is a great improvement and makes sense to me.  It is a breaking change, but for a pretty young feature and we can do breaking changes.  Making BcSymbolMap `'static` instead seems weird, because it is inherently not static.

Swatinem:
IMO making `BcSymbolMap` `'static` is possible, and might even be quite straight forward. I recently wrote https://swatinem.de/blog/magic-asref/ about using `Buf: AsRef<[u8]>` instead of a `&'data [u8]`, so you can have a static mmapped `ByteView` in there.
Then you have the problem of it being self-referencial. In that case, one could just have a `Vec<Range<usize>>`, and perform a `get` on the underlying buffer. You are chasing a few more pointers around, but you avoid it being self-referencial by doing that.

Either way, if we wanna do breaking changes, lets make an explicit list of what we want to break and plan for that.
Things like cleaning up the `symcache` code and removing all the deprecated stuff from there. Revisiting #509 also comes to mind. One reason that PR went nowhere was because of the `SelfCell` indirection of `BcSymbolMap`. Not relying on that would make us able to make transformers `&mut`.

flub:
You're secretly arguing that all public APIs should always be `Send + 'static` here because someone might want to spawn things with it.  You might as well get rid of rust lifetimes and always require `'static` at this point, many more dynamic languages do this.

If a transformer wants to allow using things with a lifetime, then making it have a lifetime itself is a valid approach.  So far we've seen two possible transformers and both have had a lifetime associated in their logical Rust-representation, so this seems like a very sensible solution.

We don't necessarily have to coordinate all breaking changes at once, having fewer makes migrations easier.  E.g. is anyone else using the symcache?  If so already removing the old symcache might be too soon.  Making transformers `mut` could also be nice I guess, I don't really understand why a self-referential struct should make this impossible though.

Swatinem:
On second thought though, the Writer itself is a one-time use thing that you don’t keep around for long, so introducing a lifetime on it sounds fair.

loewenheim (approved):

PR #545, opened 2022-04-23
ref(symcache): Add a test for mostly overlapping functions

relaxolotl:
I made an assumption in #541 that apparently was incorrect: Because we don't track file name functions, there are situations where the only distinguishing difference between two otherwise identical functions is their `entry_pc`. It's important to ensure that this doesn't change, so I've added a test to check this. The test binary is unfortunately a dSYM as I didn't want to try to wrangle clang or gcc to cross compile and spit out a DWARF.

If/When support for relative addresses are added in, this test _might_ need to be updated to adapt to that. I'm not sure how to make this as painless as possible for the person who has to do this in the future; Thoughts would be appreciated here.

### Extra 

Something to note is that allowing the `entry_pc` to determine the uniqueness of functions also has a drawback: the symcache writer by default assigns a fixed address to all inlined functions. If the "original" copy of the inlined function is also present in the debug file, then there exists at least two copies of that function in the symcache: the original copy, and the inlined version of it. Symcaches are a bit bloated because of this.


symbolic-symcache/tests/test_writer.rs
```diff
@@ -127,6 +127,25 @@ fn test_write_functions_macos() -> Result<(), Error> {
     Ok(())
 }
 
+// Tests that functions with identical names, compilation directories, and languages but different
+// entry_pcs have separate, distinct entries in the symcache. The specific use case generating this
+// is two identically-named static C functions nestled in two different files sharing a common
+// compilation directory.
```
Swatinem, line 133: would be nice to also commit the source files, and instructions to reproduce the fixture (as in: compile flags, etc)

Swatinem (approved):
nice!

Also, dsym is actually dwarf ;-)

symbolic-symcache/tests/test_writer.rs
relaxolotl: done 🎉  it's an aggressively simple example so the build process isn't too complex, thankfully. cross-compiling might be another problem, but i'm just going to close my eyes to that problem for the time being.

relaxolotl:
> Also, dsym is actually dwarf ;-)

fair enough 🥲 iirc it does have some special apple-flavoured segments too - i think clang just runs the binary through dsymutil to strip the dwarf bits out if told to extract the debug info because of that flavour

PR #584, opened 2022-05-20
Upgrade uuid, debugid, proguard

jhpratt:


Swatinem:
I recently landed these changes in https://github.com/getsentry/symbolic/pull/577. However they are right now only present in a [`breaking-release`](https://github.com/getsentry/symbolic/tree/breaking-release) feature branch, as we have a ton more breaking changes queued up for bigger release to come.
We are also working on a new release of the `pdb` crate which as well will update its `uuid` dependency. With all these dependencies, there is a long tail of releases to track and wait for I’m afraid.

I do very much appreciate your contributions though! ;-)

jhpratt:
Ah, I didn't realize that! No problem.

PR #588, opened 2022-05-30
fix: Only skip one fn when encountering unknown Unwind Codes on Win-x64

Swatinem:
Actually there was a comment above the match explaining the intent, but it did not match the actual code.


loewenheim (approved):
Nice catch!

PR #598, opened 2022-06-15
ref: Add a new object type for il2cpp

Swatinem:
I just noticed I forgot to push this change, as I thought it would be part of sentry-cli, but it actually lives in symbolic.
That said, the enum is not extensible right now, and it might be worth handling these things in sentry-cli rather than symbolic, as we only want to add this so the sentry-cli output tells us what the purpose/type of the uploaded file is, nothing more nothing less.


symbolic-debuginfo/src/base.rs
```diff
@@ -56,6 +56,9 @@ pub enum ObjectKind {
     /// information corresponding to the original object file.
     Sources,
 
+    /// An Il2Cpp LineMapping json file.
+    Il2Cpp,
```
flub, line 60: this is maybe a bit a generic name for a very specific file format?

kamilogorek: We already use this naming in all other places, so it should be rather specific to what it relates to.

ashwoods (approved):

kamilogorek:
> Didn’t I open this as a draft? 

image

👀

PR #600, opened 2022-06-20
Breaking release

ashwoods:


Swatinem (approved):
The important thing here is to "merge with a merge commit", so we don’t just squash all the individual items and their history.

ashwoods:
Besides an update to the `changelog`,  don't we want to bump the version and it's references now?

PR #627, opened 2022-07-22
ref: Move FunctionStack into its own module

Swatinem:


kamilogorek (approved):

symbolic-debuginfo/src/pdb.rs
```diff
@@ -670,6 +671,7 @@ impl<'s> Unit<'s> {
                 line: line_info.line_start.into(),
             });
         }
+        lines.sort_by_key(|line| line.address);
```
mstange, line 674: Ah, perfect, I was thinking about submitting a similar patch (also with the first() / last() change below). Thanks!

PR #631, opened 2022-07-25
Revert "feat: Use pdb_addr2line::TypeFormatter (#426)"

Swatinem:
This reverts commit ca9c64bc355cb0f4c508b7c3cf94f89ec92ec756.

The reason for the revert is that the new code throws errors when processing the `electron.exe.pdb` from this archive:
https://github.com/electron/electron/releases/download/v20.0.0-beta.11/electron-v20.0.0-beta.11-win32-x64-pdb.zip

It might be possible to safely skip over this error instead, but I rather revert for now and figure out a more error tolerant way later.

CC @mstange you probably know your way around that code better to figure out a proper solution for this.

My example errors out with: `Type 1660 not found`.
I was invoking the `symcache_debug` util like so:

> cargo run -p symcache_debug --release -- -d testcases/electron20/electron.exe.pdb -w --report

### DIFF
--- a/symbolic-debuginfo/Cargo.toml
+++ b/symbolic-debuginfo/Cargo.toml
@@ -48,13 +48,12 @@ macho = [
 ]
 # PDB/PE processing
 ms = [
-    "elsa",
     "goblin/pe32",
     "goblin/pe64",
     "goblin/std",
     "lazycell",
     "parking_lot",
-    "pdb-addr2line",
+    "pdb",
     "scroll",
     "smallvec",
 ]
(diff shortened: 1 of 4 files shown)


mstange:
I'm taking a look.

mstange:
I've released pdb-addr2line 0.10.1 with the fix. The offending function now gets stringified as `std::Cr::codecvt<char16_t,char8_t,_Mbstatet>::do_out(_Mbstatet&, char16_t const*, char16_t const*, char16_t const*&, <missing type 0x67c>, <missing type 0x67c>, <missing type 0x67c>) const`

The electron.exe.pdb now successfully converts to a symcache.

In addition, we can add a fallback to this call to `formatter.format_function`:

https://github.com/getsentry/symbolic/blob/99750d304d265bf5c685d175322cce19843afefe/symbolic-debuginfo/src/pdb.rs#L700-L704

For example, we could turn it into the following:

```rust
        let name = name.to_string();
        let name = Name::new(
            formatter.format_function(&name, self.module_index, type_index).unwrap_or(name),
            NameMangling::Unmangled,
            Language::Unknown,
        );
```

Swatinem:
thanks for looking into this! I think falling back to the original name is also a sensible thing to do

Swatinem:
lets roll forward: #634

PR #632, opened 2022-07-25
ref: Simplify FunctionStack and PDB handling

Swatinem:
PDB inlinees and line records are being sorted now, and line records are being cleaned:
- zero-size line records are removed
- consecutive line records are merged

### DIFF
--- a/symbolic-debuginfo/src/pdb.rs
+++ b/symbolic-debuginfo/src/pdb.rs
@@ -662,16 +662,39 @@ impl<'s> Unit<'s> {
                 None => continue,
             };
 
+            // skip 0-sized line infos
+            let size = line_info.length.map(u64::from);
+            if size == Some(0) {
+                continue;
+            }
+
             let file_info = program.get_file_info(line_info.file_index)?;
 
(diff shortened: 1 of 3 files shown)


kamilogorek (approved):

symbolic-debuginfo/src/pdb.rs
```diff
@@ -662,17 +662,40 @@ impl<'s> Unit<'s> {
...
             });
         }
         lines.sort_by_key(|line| line.address);
 
+        // Merge line infos that only differ in their `column` information, which we don't
+        // care about. We only want to output line infos that differ in their file/line.
+        lines.dedup_by(|current, prev| {
+            // the records need to be consecutive to be able to merge
+            let first_end = prev.size.and_then(|size| prev.address.checked_add(size));
+            let is_consecutive = first_end == Some(current.address);
```
kamilogorek, line 687: Just to double-check. This should be the same address, not `first_end + 1`?

Swatinem: Yes indeed

PR #644, opened 2022-07-29
meta: Write a changelog for the changes since 9.0.0.

mstange:


Swatinem (approved):
Thanks! You could also give yourself credit as external contributor ;-)

Issue #638, opened 2022-07-28
Investigate Windows Unwinding involving `RtlDispatchException`


Swatinem:
We have a customer (with retained minidumps and usable debug files; sentry employees should know which one I mean) that reports incomplete / truncated stack traces for Windows minidumps that involve `RtlDispatchException`.

We recently improved our handling of [`UWOP_SET_FPREG`](https://github.com/getsentry/symbolic/issues/546) which is used by this function.
I have also verified again that a previously broken minimal reproduction in https://github.com/getsentry/sentry-native/issues/582 now processes correctly.

The investigation should focus on the following points:
- `UWOP_SET_FPREG` relies on `$rbp` being available. Is it? If not, why? And can we get it from somewhere?
- NOTE: `$rbp` is a preserved / non-volatile / callee saves register. So if CFI does not explicitly state how to restore it, we should be able to assume it is not modified by the function.
- Maybe some other frames in the stack use `UWOP_PUSH_MACHFRAME`, and we are running into https://github.com/rust-minidump/rust-minidump/issues/637 ?

---

<details>
<summary>`RtlDispatchException` unwind info</summary>

You need a [patched llvm-objdump](https://reviews.llvm.org/D107655) for this.

```
Function Table:
  Start Address: 0x51210
  End Address: 0x51578
  Unwind Info Address: 0x146454
    Version: 1
    Flags: 3 UNW_ExceptionHandler UNW_TerminateHandler
    Size of prolog: 48
    Number of Codes: 12
    Frame register: RBP
    Frame offset: 64
    Unwind Codes:
      0x1f: UOP_SaveNonVol RBX [0x0220]
      0x18: UOP_SetFPReg 
      0x13: UOP_AllocLarge 58
      0x0c: UOP_PushNonVol R15
      0x0a: UOP_PushNonVol R14
      0x08: UOP_PushNonVol R13
      0x06: UOP_PushNonVol R12
      0x04: UOP_PushNonVol RDI
      0x03: UOP_PushNonVol RSI
      0x02: UOP_PushNonVol RBP

---

STACK CFI INIT 51210 368
  .cfa: $rbp 464 +
  $rbp: .cfa 16 - ^
  $rsi: .cfa 24 - ^
  $rdi: .cfa 32 - ^
  $r12: .cfa 40 - ^
  $r13: .cfa 48 - ^
  $r14: .cfa 56 - ^
  $r15: .cfa 64 - ^
  $rbx: $rbp 480 + ^
[...]

PR #645, opened 2022-08-02
fix: Allow Windows CFI to underflow

Swatinem:
We have seen some SaveNonVolatile unwind operations underflow the current stack frame, so they are saving/loading registers into the callers stack frame.
For example `VCRUNTIME140.dll!__C_specific_handler` can do this.

Its full Unwind Operations look like this:

```
Function Table:
  Start Address: 0xbff0
  End Address: 0xc1eb
  Unwind Info Address: 0x10138
    Version: 1
    Flags: 0
    Size of prolog: 28
    Number of Codes: 12
    No frame pointer used
    Unwind Codes:
      0x1c: UOP_SaveNonVol RSI [0x0080]
      0x1c: UOP_SaveNonVol RBP [0x0078]
      0x1c: UOP_SaveNonVol RBX [0x0070]
      0x1c: UOP_AllocSmall 64
      0x18: UOP_PushNonVol R15
      0x16: UOP_PushNonVol R14
      0x14: UOP_PushNonVol R13
      0x12: UOP_PushNonVol R12
      0x10: UOP_PushNonVol RDI
```

We previously used saturating_sub to clamp these offsets to 0, meaning things are loaded from the current stack frame instead of from the callers frame. Now we correctly generate ASCII CFI for these Unwind Codes.

This fixes one pre-condition of #638, namely that it depends on `$rbp` which is loaded from the above unwind codes, thus outside its own stack frame.

### DIFF
--- a/symbolic-cfi/src/lib.rs
+++ b/symbolic-cfi/src/lib.rs
@@ -1018,7 +1018,13 @@ impl<W: Write> AsciiCfiWriter<W> {
             }
 
             // The minimal stack size is 8 for RIP
-            let mut stack_size: u32 = 8;
+            // Also, we are using signed math here everywhere, as some unwind operations, such as
+            // `SaveNonVolatile` can actually point out of the current stack frame, which appears
+            // to be perfectly fine.
+            // When this happens, the resulting ASCII CFI can also contain weird things like
+            // `.cfa -8 - ^` (that means, take the value 8 *above* the stack frame) which is
+            // perfectly fine as well.
+            let mut stack_size: i32 = 8;
             // Special handling for machine frames
             let mut machine_frame_offset = 0;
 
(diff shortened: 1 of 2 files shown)


symbolic-cfi/src/lib.rs
```diff
@@ -1074,7 +1080,8 @@ impl<W: Write> AsciiCfiWriter<W> {
                                         " {}: {} {} + ^",
                                         reg.name(),
                                         unwind_info.frame_register.name(),
-                                        offset.saturating_sub(unwind_info.frame_register_offset)
+                                        offset.wrapping_sub(unwind_info.frame_register_offset)
+                                            as i32
```
kamilogorek, line 1084: Ugh, thats ugly wrapping :3

kamilogorek (approved):
Snapshots + nit, but 💯

PR #649, opened 2022-08-03
fix: Undecorate Windows Symbols in SymCache

Swatinem:
https://github.com/getsentry/symbolic/pull/642 removed stripping the underscores from C-decorated Windows symbols.

The `symbolic-debuginfo` abstraction now outputs raw names (except for mach-o, where it still strips underscores).

As a followup to this change, this will now undecorate these symbols when writing a symcache.

### DIFF
--- a/symbolic-symcache/src/writer.rs
+++ b/symbolic-symcache/src/writer.rs
@@ -699,7 +699,7 @@ fn undecorate_win_symbol(name: &str) -> &str {
             }
         }
     }
-    
+
     name
 }
 
--- a/symbolic-symcache/src/writer.rs
+++ b/symbolic-symcache/src/writer.rs
@@ -679,7 +679,7 @@ fn undecorate_win_symbol(name: &str) -> &str {
 
     // Parse __vectorcall.
     if let Some((name, param_size)) = name.rsplit_once("@@") {
-        if let Ok(_param_size) = param_size.parse::<u32>() {
+        if param_size.parse::<u32>().is_ok() {
             return name;
         }
     }
@@ -688,7 +688,7 @@ fn undecorate_win_symbol(name: &str) -> &str {
     if !name.is_empty() {
         if let ("@" | "_", rest) = name.split_at(1) {
             if let Some((name, param_size)) = rest.rsplit_once('@') {
-                if let Ok(_param_size) = param_size.parse::<u32>() {
+                if param_size.parse::<u32>().is_ok() {
                     // __stdcall or __fastcall
                     return name;
                 }
@@ -699,6 +699,7 @@ fn undecorate_win_symbol(name: &str) -> &str {
             }
         }
     }
+    
     name
 }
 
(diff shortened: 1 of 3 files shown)


symbolic-symcache/src/writer.rs
```diff
@@ -639,6 +659,49 @@ impl<'a> SymCacheConverter<'a> {
...
+/// This code is adapted from `dump_syms`:
+/// See https://github.com/mozilla/dump_syms/blob/325cf2c61b2cacc55a7f1af74081b57237c7f9de/src/symbol.rs#L169-L216
+fn undecorate_win_symbol(name: &str) -> &str {
+    if name.starts_with('?') || name.contains(&[':', '(', '<']) {
+        return name;
+    }
+
+    // Parse __vectorcall.
+    if let Some((name, param_size)) = name.rsplit_once("@@") {
+        if let Ok(_param_size) = param_size.parse::<u32>() {
```
kamilogorek, line 682:
```suggestion
        if param_size.parse::<u32>().is_ok() {
```

we dont use parsed value anyway.

symbolic-symcache/src/writer.rs
```diff
@@ -639,6 +659,49 @@ impl<'a> SymCacheConverter<'a> {
...
+                    return name;
+                }
+            }
+            if let Some(name) = name.strip_prefix('_') {
+                // __cdecl
+                return name;
+            }
+        }
+    }
+    name
```
kamilogorek, line 702:
```suggestion
    }
    
    name
```

symbolic-symcache/src/writer.rs
```diff
@@ -639,6 +659,49 @@ impl<'a> SymCacheConverter<'a> {
...
+        if let Ok(_param_size) = param_size.parse::<u32>() {
+            return name;
+        }
+    }
+
+    // Parse the other three.
+    if !name.is_empty() {
+        if let ("@" | "_", rest) = name.split_at(1) {
+            if let Some((name, param_size)) = rest.rsplit_once('@') {
+                if let Ok(_param_size) = param_size.parse::<u32>() {
```
kamilogorek, line 691:
```suggestion
                if param_size.parse::<u32>().is_ok() {
```

symbolic-symcache/src/writer.rs
```diff
@@ -140,6 +146,8 @@ impl<'a> SymCacheConverter<'a> {
             self.process_symbolic_symbol(&symbol);
         }
 
+        self.is_windows_object = false;
+
```
kamilogorek, line 150:
```suggestion
```

Should line 137 already cover this case? `matches!` will just return `false` for us.

kamilogorek (changes requested):

symbolic-symcache/src/writer.rs
Swatinem: This was intentional. I only want to scope this to the call. Which is a bit leaky anywhere, as there is a bunch of `?` in that function that early-return and bypass this.

Though the `process_symbolic_function` is public so anyone could just process a random function as well, for which we don’t know where it came from.

kamilogorek (approved):
wow, really cargo, really?

image

PR #650, opened 2022-08-03
ref: Remove all experimental usym and il2cpp metadata code

Swatinem:
The `-il2cpp` subcrate has always had a note that it was experimental and things could disappear at any time, so lets make that happen.

This should hopefully slightly improve build times as we don’t build a bunch of unused code anymore.

If anyone feels nostalgic about any of this, it is still preserved in our commit history.


kamilogorek (approved):
bye.

PR #653, opened 2022-08-03
ref: Use CFI for modules that only have a CodeId

Swatinem:
We are dealing with a PE file that does not have a CodeView record / DebugId,
in which case we should still be able to use it based on its CodeId.

### DIFF
--- a/examples/minidump_stackwalk/src/main.rs
+++ b/examples/minidump_stackwalk/src/main.rs
@@ -17,14 +17,14 @@ use thiserror::Error;
 use walkdir::WalkDir;
 
 use symbolic::cfi::CfiCache;
-use symbolic::common::{Arch, ByteView, DebugId, InstructionInfo, SelfCell};
+use symbolic::common::{Arch, ByteView, CodeId, DebugId, InstructionInfo, SelfCell};
 use symbolic::debuginfo::{Archive, FileFormat};
 use symbolic::demangle::{Demangle, DemangleOptions};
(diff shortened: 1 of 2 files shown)


Swatinem:
BTW, we should port this also to symbolicator, as in https://getsentry.atlassian.net/browse/ISSUE-1559 we are able to fetch the file from an external symbol server via the CodeId, though we don’t use it for CFI/stackwalking.

examples/minidump_stackwalk/src/main.rs
```diff
@@ -108,26 +114,50 @@ impl<'a> LocalSymbolProvider<'a> {
...
+                    let by_debug_id = object_db
+                        .by_debug_id
+                        .entry(debug_id)
+                        .or_insert_with(Vec::new);
+                    by_debug_id.push(object_meta.clone());
+                }
+
+                if let Some(code_id) = object.code_id() {
+                    let by_code_id = object_db.by_code_id.entry(code_id).or_insert_with(Vec::new);
+                    by_code_id.push(object_meta);
```
kamilogorek, line 144: I wonder why we need to clone in `debug_id` case, but not here?

kamilogorek (approved):

examples/minidump_stackwalk/src/main.rs
Swatinem: Because we don’t need the original anymore after this call, so it is moved ;-)

PR #655, opened 2022-08-08
fix: Slightly lower demangling recursion limit

Swatinem:
This brings us in line with the symbolic changes from the long running `fix/demangle-fixes`
branch. Tests are only reformatted.

We should be able to merge this into symbolic, and rather pull in our patched `cpp_demangle` via a cargo `[patch]` section in symbolicator.


kamilogorek (approved):
CI failures will be fixed once https://github.com/getsentry/symbolic/pull/656 is merged

kamilogorek:
Fixed CI, rebased, merged.

PR #658, opened 2022-08-10
fix: Correctly read the debug_id of Deterministic PE files

Swatinem:
Deterministic PE files have a slightly different way of storing their debug_id, which
should match the corresponding ID inside a portable pdb `#Pdb` stream.

See https://github.com/dotnet/runtime/blob/main/docs/design/specs/PE-COFF.md#codeview-debug-directory-entry-type-2


symbolic-debuginfo/src/pe.rs
```diff
@@ -116,18 +116,31 @@ impl<'data> PeObject<'data> {
...
+            })
+            .and_then(|(debug_directory, cv_record)| {
+                let guid = &cv_record.signature;
+
+                // Deterministic PE files have a different debug_id format:
+                //
+                // > Version Major=any, Minor=0x504d of the data format has the same structure as above.
+                // > The Age shall be 1. The format of the .pdb file that this PE/COFF file was built with is Portable PDB.
+                // > The Major version specified in the entry indicates the version of the Portable PDB format.
+                // > Together 16B of the Guid concatenated with 4B of the TimeDateStamp field of the entry form a PDB ID that should be used to match the PE/COFF image with the associated PDB (instead of Guid and Age).
```
mstange, line 133: "concatenated with 4B of the TimeDateStamp field"

Do you know if this means "including leading zeros" or "excluding leading zeros"? The `from_guid_age` method excludes leading zeros of the age.

kamilogorek (approved):

symbolic-debuginfo/src/pe.rs
Swatinem: I think internally that is just all bytes, its only when you pretty print it that leading zeros will be stripped.

So I would say this does not really matter. 🤷🏻‍♂️

mstange: I see. Well, it matters for the pretty printers, if you ever need to compare the pretty printed forms for equality.
For example, doesn't it matter for the symbol server paths?

Swatinem: Good point. I have no idea if we can/should try to download portable pdbs from symbol servers.
Maybe @bruno-garcia can shed some light here? I think mono was rather looking up the PDBs by the module version id, which is a different thing entirely.

we need to eventually parse the metadata to read the module version id, https://github.com/getsentry/symbolic/pull/659#issue-1334757961 is some preliminary work to get that going.

loewenheim (approved):

symbolic-debuginfo/src/pe.rs
mattjohnsonpint: We are currently having an issue with this just in that some of our tests in the sentry-dotnet repo are flaky because we generate debug IDs without the leading zero on this part.  So sometimes the last part isn't 8 full digits.

Since yes - it's all bytes internally - then it only matters for string comparisons. But I'd say go ahead and format it as 8 digits always (with leading zeros as needed).

I'm going to make a similar change in the .NET SDK code to resolve the flaky test, so it should match.

mattjohnsonpint: See https://github.com/getsentry/sentry-dotnet/pull/2114

Swatinem: To close this discussion:
- Internally we have consistent pretty-printing of these debug ids, so sentry-cli uploading to sentry, and symbolicator querying from sentry. It does not really matter if it is with or without leading zero, at lost as it is consistent.
- For accessing external symbol servers, the spec requires us to mask out the last bytes with `FF`, probably to punt on this very question :-D So it does not matter for external servers either.
- The SDK PR that @mattjohnsonpint mentioned is there to improve test flakyness, as the tests are masking the debug id, but previously had a 1-in-16 chance to fail because the masked string changed in length. However when the SDK sends the event to sentry and to symbolicator, it is being parsed as a `DebugId` and the above two points apply, so everything is fine.

It would be nice to have explicit tests for this edge case though. So if someone wants to try the "bitcoin-like" hashing game and give it a 1-in-16 chance to generate a file with a leading-0 id, be my guest :-)

PR #663, opened 2022-08-16
ref: Skip parsing Minified Source when dealing with Metro/Hermes

Swatinem:
The Metro/Hermes SourceMaps have scopes embedded in them which we now prefer.
In that case, we can completely skip parsing and extracting scopes from the minified JS.

This depends on having https://github.com/getsentry/rust-sourcemap/pull/48 in a published release.


kamilogorek (approved):
Niceee

PR #680, opened 2022-09-13
ref: Clean up public SourceMapCache APIs

Swatinem:


kamilogorek (approved):

Swatinem:
Also related to the `_internal_not_stable_` module:

Can we move some of these parts out of the symbolic monorepo, so that we can independently evolve and break the API of the smaller moving parts there.

Symbolic itself is really just concerned with baking this into a binary format that you can mmap and do quick lookups.

PR #685, opened 2022-09-16
fix: Be more lenient parsing Abbreviations

Swatinem:
The recent fix for Abbreviations parsing might have been too overeager and may cause EOF errors in case there are no Abbreviations (at offset 0) at all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,17 @@
 # Changelog
 
+## Unreleased
+
+**Fixes**:
+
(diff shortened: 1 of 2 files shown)


ashwoods (approved):

loewenheim (approved):

Swatinem:
For visibility: I think the underlying problem might have caused an increase in https://sentry.my.sentry.io/organizations/sentry/issues/311736/?project=5

PR #688, opened 2022-09-23
feat: Add new `sourcemapcache` subcrate and python bindings

Swatinem:
The new subcrate offers functionality to build a fast lookup index for resolving minified JS line/column to their original line/column/file and function name using information extracted from a provided minified source and its corresponding sourcemap.

This functionality is also exposed via the symbolic python API to be used by the sentry monolith, see https://github.com/getsentry/sentry/pull/38551


kamilogorek (approved):

symbolic-sourcemapcache/src/lookup.rs
```diff
@@ -0,0 +1,353 @@
...
+        }
+
+        let (_, buf) = align_to(buf, 8).ok_or(Error::SourcePositions)?;
+        let num_mappings = header.num_mappings as usize;
+        let (min_source_positions, buf) =
+            raw::MinifiedSourcePosition::slice_from_prefix(buf, num_mappings)
+                .ok_or(Error::SourcePositions)?;
+        let (orig_source_locations, buf) =
+            raw::OriginalSourceLocation::slice_from_prefix(buf, num_mappings)
+                .ok_or(Error::SourceLocations)?;
```
Swatinem, line 107: the writer does an `align(8)` in between these two tables, the reader should reflect that.

symbolic-cabi/src/sourcemapcache.rs
```diff
@@ -0,0 +1,197 @@
...
+        let sourcemap_slice = slice::from_raw_parts(sourcemap_content as *const _, sourcemap_len);
+
+        let writer = SourceMapCacheWriter::new(
+            String::from_utf8_lossy(source_slice).deref(),
+            String::from_utf8_lossy(sourcemap_slice).deref()
+        )?;
+        let mut buffer = Vec::new();
+        writer.serialize(&mut buffer)?;
+
+        let byteview = ByteView::from_vec(buffer);
```
Swatinem, line 97: This holds the sourcemapcache buffer in memory. Not sure its worth putting in the API to write it to disk, and to read/map it from disk right now. But we should think about that rather sooner than later.

symbolic-cabi/src/sourcemapcache.rs
```diff
@@ -0,0 +1,197 @@
...
+
+        let post_context = SymbolicStrVec::from_vec(post_context);
+
+        (pre_context, post_context)
+    } else {
+        (
+            SymbolicStrVec {
+                strs: ptr::null_mut(),
+                len: 0,
+            },
```
Swatinem, line 147: How about deriving `Default`? Or is that not possible for raw pointers?

Swatinem: yup, its not possible for raw pointers, so scratch that ;-)

PR #699, opened 2022-10-17
ref: Cross-compile aarch64 python pkg

Swatinem:
This ports https://github.com/getsentry/relay/pull/1438 to symbolic. It is using a x86_64 docker container configured for cross compiling that builds a aarch64 symbolic binary which is then re-used when building the aarch64 wheel. Thus it avoids running the whole cargo build in aarch64 qemu which can be extremely slow.

---

Currently the complete build for the binary python package runs in an arm64 qemu, which is super slow. The idea would be to cross-compile the symbolic-cabi library and then move that into the arm64 qemu image. That way, the cross-compile would be slow, and only a tiny part of packaging up the python bits is run in qemu.

The difficulty here is to find the right way to cross-compile the library, where the C/C++ bits (mostly related to demangling) need a C++ cross-compiling setup, and the image that does the cross compilation is "sufficiently old" to have a build-time glibc dependency that matches the one required by the python target.

As I believe python `manylinux2014` is already outdated itself, we might jump to the next higher version (`manylinux_2_24` I believe?) which has newer glibc requirements, requiring a "less old" image / C++ compiler to build.


loewenheim (approved):

Swatinem:
> * How hard is the requirement on `manylinux2014`? Which would be acceptable alternatives?

I believe we can bump quite a bit higher than that. I will defer to @asottile-sentry to give an authoritative answer and to also chime in on the other questions.

Another alternative would be to just stop pulling in the C++ (swift-demangle) dependency into the CABI crate.
We are not using that directly ourselves. Another user of our python bindings is Mozilla. We were discussing deprecations already some time ago in https://github.com/getsentry/symbolic/issues/573#issuecomment-1129234530 where it did not seem that they were using the demangling part from python CC @gabrielesvelto @willkg

We might as well just remove the need to build the C++ code alltogether. We might need to do an API-breaking release soon anyway due to #729.

Swatinem:
@loewenheim noticed that we have a `SourceLocation.function_name` getter that does demangling in the symcache bindings:

https://github.com/getsentry/symbolic/blob/0b39652fbbc230bb4fdbcf6e4bab657664dd5ab0/py/symbolic/symcache.py#L41-L44

@gabrielesvelto can I assume that you are never using that? In which case we can just safely remove it with the next breaking change.

willkg:
Mozilla Symbolication Service is not using `function_name` in symcache directly.

asottile-sentry:
for manylinux, sentry only cares about 2_28+ so go wild

PR #703, opened 2022-10-31
feat(smcache): Expose token name via Python bindings

kamilogorek:
Followup to https://github.com/getsentry/symbolic/pull/698

### DIFF
--- a/symbolic-cabi/src/sourcemapcache.rs
+++ b/symbolic-cabi/src/sourcemapcache.rs
@@ -65,6 +65,8 @@ pub struct SymbolicSmTokenMatch {
     pub col: u32,
     /// The path to the original source.
     pub src: SymbolicStr,
+    /// The name of the source location as it is defined in the SourceMap.
+    pub name: SymbolicStr,
     /// The name of the function containing the token.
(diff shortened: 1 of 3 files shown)


symbolic-cabi/src/sourcemapcache.rs
```diff
@@ -156,6 +158,7 @@ fn make_token_match(token: SourceLocation, context_lines: u32) -> *mut SymbolicS
         line: token.line() + 1,
         col: token.column() + 1,
         src: SymbolicStr::new(token.file_name().unwrap_or_default()),
+        name: SymbolicStr::new(token.name().unwrap_or_default()),
```
Swatinem, line 161: The `unwrap_or_default` means we will return an empty string in case there is no name. is that really what we want?

Swatinem (approved):

symbolic-cabi/src/sourcemapcache.rs
kamilogorek: Yes, because `decode_str` will then return `''`, which then will fallback to `None` in `sourcemapcache.py`

PR #707, opened 2022-11-09
Bump goblin to 0.6

Porges:
Bump `goblin` dependency across breaking version 0.5→0.6. There are [several bug fixes](https://github.com/m4b/goblin/blob/master/CHANGELOG.md#060---2022-10-23) included.

The version number update is tentative; it’s possible there are breaking changes exposed (in particular [the MachO types changed](https://github.com/m4b/goblin/compare/0.5.4...0.6.0#diff-a22eb400c0763d872ba8fd4c9e1a563d0e0cf957c36fb3bf6796318ac2278827)), in which case this should be a major version bump?

### DIFF
--- a/symbolic-debuginfo/Cargo.toml
+++ b/symbolic-debuginfo/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "symbolic-debuginfo"
-version = "10.1.1"
+version = "10.1.2"
 license = "MIT"
 authors = [
     "Armin Ronacher <armin.ronacher@active-4.com>",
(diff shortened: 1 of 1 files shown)


symbolic-debuginfo/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "symbolic-debuginfo"
-version = "10.1.1"
+version = "10.1.2"
```
Swatinem, line 3: no need to do this, we have our own release scripts for this

Swatinem (approved):

Swatinem:
I wend ahead and bumped the dep along with others in #709

PR #724, opened 2022-11-29
ref: Update gimli

Swatinem:
Reverts #683 as upstream gimli now has an `AbbreviationsCache` that serves the same purpose.


loewenheim (approved):
Looks like there are some nice simplifications!

PR #738, opened 2022-12-21
Use cross-compilation for aarch64 release-build

supervacuus:
This is a replacement draft for #699. Differences are:

* it doesn't change the existing workflows for `i686` and `x86_64` 
* it introduces a minor conditional on the path from the GHA to the `aarch64`-specific build-script
* the `aarch64`-specific build script is run in a minimal cross-compilation container which provides
  * a `manylinux2014` compatible cross-compilation toolchain (which can be easily switched to manylinux_2_28 if needed)
  * a `pypa manylinux` environment
* the build script differs from the normal build script in that
  * it must build `libffi` with the cross-compilation toolchain
  * sets up a python `crossenv` so we don't have to load and start yet another container to build the` aarch64` wheel

#skip-changelog

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,7 +6,7 @@ on:
       - "release/**"
       
 env:
-  MANYLINUX_VERSION: manylinux2014
+  MANYLINUX_VERSION: manylinux_2_28
 
 jobs:
   python-wheel-mac:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,13 +54,13 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3
-
+      
       - if: matrix.build-arch == 'manylinux2014_aarch64'
-        uses: docker/setup-qemu-action@v1
-        with:
-          platforms: arm64
-
-      - name: Build in Docker
+        name: Build in Docker (cross-compile host:amd64, target:aarch64)
+        run: make wheel-manylinux-aarch64 IMAGE=messense/manylinux2014-cross:aarch64-amd64
+      
+      - if: matrix.build-arch != 'manylinux2014_aarch64'
+        name: Build in Docker
         run: make wheel-manylinux IMAGE=quay.io/pypa/${{ matrix.build-arch }}
 
       - uses: actions/upload-artifact@v3.1.1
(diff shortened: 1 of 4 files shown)


Swatinem (approved):
lgtm, and build times are a lot better according to the CI run!

As @asottile-sentry mentioned in https://github.com/getsentry/symbolic/pull/699#issuecomment-1337659763 you can just straight to manylinux 2_28.
that should hopefully give us more time before we have to touch any of this again.

Swatinem (approved):

supervacuus:
> lgtm, and build times are a lot better according to the CI run!

Yeah, it should be in the same ballpark as the `x86_64` and `i686` ones. The docker image is tiny (relative to what it provides), and the build is now native. I also removed `qemu`, so there is no `aarch64` binary running anywhere, not even the wheel-building `pypa` container.

> As @asottile-sentry mentioned in [#699 (comment)](https://github.com/getsentry/symbolic/pull/699#issuecomment-1337659763) you can just straight to manylinux 2_28. that should hopefully give us more time before we have to touch any of this again.

I only used `manylinux2014` because all other `symbolic` Linux downloads also use it. We can switch to `manylinux_2_28` for all Linux wheels or only the `aarch64` one. I leave this up to you. @asottile-sentry clearly said to go wild :-).

supervacuus:
Hmm, turns out there is no pypa image hosted on [quay.io](https://quay.io/organization/pypa) for `manylinux_2_28` and `i686`. Should we use `manylinux_2_24` for this architecture?

Swatinem:
> there is no pypa image hosted on [quay.io](https://quay.io/organization/pypa) for `manylinux_2_28` and `i686`. Should we use `manylinux_2_24` for this architecture?

Yes, we can do that for now. @asottile-sentry can probably enlighten us if we even need to support 32-bit platforms at all.

supervacuus:
Ugh, `manylinux_2_24` is based on Debian 9. They're also [discussing the deprecation of the image](https://github.com/pypa/manylinux#manylinux_2_24-debian-9-based) for other reasons. I will revert `i686` to `manylinux2014`.

asottile-sentry:
I would just drop 32 bit -- it's 2023

PR #744, opened 2023-01-16
feat: Support DWARF debug info in PE files

### LINKED ISSUE #352: Support DWARF in PE
Potential downstream consumer of this is https://github.com/mozilla/dump_syms/issues/224

Older Rust compiler versions also used to output DWARF sections in PE.

casept:
I'm not quite sure how I should implement tests for this. All the other formats seem to use `breakpad` binaries as the test case, but `breakpad_client` doesn't build under MinGW. Is it relevant that the test cases match? Or can I just use some random project for this?

Also, is support for platforms with non-4K page size relevant? Goblin doesn't seem to expose page size, so this metadata would have to be kept up-to-date in `symbolic`.

Closes #352.


casept:
I've snapshotted a binary of the sqlite3 shell I compiled for the test cases now.

loewenheim (approved):
Looks good to me, thank you very much for this!

symbolic-debuginfo/src/pe.rs
```diff
@@ -209,14 +212,30 @@ impl<'data> PeObject<'data> {
...
+        DwarfDebugSession::parse(self, symbols, self.load_address() as i64, self.kind())
     }
 
     /// Determines whether this object contains stack unwinding information.
     pub fn has_unwind_info(&self) -> bool {
+        // Sections don't exist in "vanilla" PE, only one w/ DWARF
+        // FIXME: Is this change needed / what uses this functionality?
         !self.is_stub && self.exception_data().map_or(false, |e| !e.is_empty())
+            || self.has_section("eh_frame")
+            || self.has_section("debug_frame")
```
Swatinem, line 238: Our CFI converter is using this. I believe we can skip checking for `eh_frame` and `debug_frame` as long as we cannot actually consume that data, which I believe we do not at this point.

symbolic-debuginfo/src/pe.rs
```diff
@@ -229,6 +248,17 @@ impl<'data> PeObject<'data> {
         &self.pe.sections
     }
 
+    /// Returns the `SectionTable` for the section with this name, if present.
+    pub fn section(&self, name: &str) -> Option<SectionTable> {
```
Swatinem, line 252: can we leave this as private, or does it need to be `pub`?

symbolic-debuginfo/src/pe.rs
```diff
@@ -357,54 +387,26 @@ impl<'data, 'object> Iterator for PeSymbolIterator<'data, 'object> {
...
+        // the only known platform running PE's with big-endian code is the Xbox360. Probably not worth handling.
+        return RunTimeEndian::Little;
+    }
+
+    fn raw_section(&self, name: &str) -> Option<DwarfSection<'data>> {
+        // Name is given without leading "."
+        let sect = self.section(&format!(".{}", name))?;
+        // TODO: Proper error handling if conversion fails
+        let dwarf_data: &'data [u8] = &self.data[sect.pointer_to_raw_data.try_into().unwrap()..]
+            [..sect.virtual_size.try_into().unwrap()];
```
Swatinem, line 402: I believe we can just use `as` casting for the number values here.
However I would very much prefer the use of `.get()` to be safe against out of bounds access. Slice syntax panics if one of these is out of bounds. As we use "untrusted" values coming from the section table, we should be defensive here.

symbolic-debuginfo/src/object.rs
```diff
@@ -446,7 +446,6 @@ pub enum ObjectDebugSession<'d> {
     Breakpad(BreakpadDebugSession<'d>),
     Dwarf(DwarfDebugSession<'d>),
     Pdb(PdbDebugSession<'d>),
-    Pe(PeDebugSession<'d>),
```
Swatinem, line 449: Just a NOTE: removing these types is considered a breaking change.

As https://github.com/getsentry/publish/issues/1700 has been stuck and noone has tried to fix this yet, we can squeeze this breaking change into the next release still.

Please make sure to document the changes in the changelog though.

Swatinem:
Also, 3+M for the testcase is a bit hefty. All you would need is a main fn with a `printf`, just enough to get some kind of debug info out of it.

symbolic-debuginfo/src/pe.rs
casept: I didn't end up using this function. Removed.

symbolic-debuginfo/src/pe.rs
casept: Alright, removed this for now.

symbolic-debuginfo/src/pe.rs
casept: Makes sense, I've implemented your suggestions.

casept:
> Also, 3+M for the testcase is a bit hefty. All you would need is a main fn with a `printf`, just enough to get some kind of debug info out of it.

The reason why I didn't do this is because the other test fixtures looked more like "real world" programs. The MacOS example is also over 1M in size.

Anyways, the test cases are replaced now.

symbolic-debuginfo/src/object.rs
casept: I've updated the changelog.

Swatinem (approved):

Swatinem:
Clippy still complains about unneeded `return` statements. After that I would say its good to go :-)

casept:
> Clippy still complains about unneeded `return` statements. After that I would say its good to go :-)

Fixed!

PR #752, opened 2023-02-01
feat: Access embedded PortablePDB in DLL

vaind:
replaces & closes #659

PE (a.k.a a .dll) may contain embedded Portable PDB (when built with `<DebugType>embedded</DebugType>`) as shown in the following screenshot from dotPeek
image

We need to access that but it seems it's currently impossible with the used PE reader, see the tracking issue: https://github.com/m4b/goblin/issues/314


loewenheim (dismissed):

symbolic-ppdb/tests/test_ppdb.rs
```diff
@@ -73,3 +86,77 @@ fn test_embedded_sources_with_metadata_broken() {
...
+    let pe_buf = std::fs::read(fixture(
+        "windows/Sentry.Samples.Console.Basic-embedded-ppdb.dll",
+    ))
+    .unwrap();
+    let pe = PeObject::parse(&pe_buf).unwrap();
+
+    let embedded_ppdb = pe.embedded_ppdb().unwrap().unwrap();
+
+    let mut ppdb_buf = vec![0; embedded_ppdb.get_size()];
+    embedded_ppdb.decompress(&mut ppdb_buf).unwrap();
```
Swatinem, line 114: It does not matter for the unit test, but `Vec::with_capacity` allocates, but does not zero-initialize.

Other than that, writing into an `&mut` out paramater might be useful in theory, but I’m wondering if that is a case of YAGNI here, as I can’t imagine any other use-case than doing this with a fresh `Vec`. If the param would be a `impl Write`, that would be different, as you could write directly to a file. But this way, it might just be simpler to return `Vec<u8>` from that fn.

symbolic-ppdb/tests/test_ppdb.rs
```diff
@@ -73,3 +86,77 @@ fn test_embedded_sources_with_metadata_broken() {
...
+    let ppdb = PortablePdb::parse(&ppdb_buf).unwrap();
+
+    assert!(ppdb.pdb_id().is_some());
```
Swatinem, line 117: instead of just `is_some`, can we check it is the same `debug_id` as reported by the PE?

Swatinem (commented):
This is awesome work! Reminds me that I have https://github.com/m4b/goblin/pull/319 just laying there since half a year :-D

symbolic-debuginfo/src/pe.rs
```diff
@@ -276,6 +278,60 @@ impl<'data> PeObject<'data> {
...
+            if idd.data_type == 17 {
+                // ImageDebugDirectory.pointer_to_raw_data stores a raw offset -- not a virtual offset -- which we can use directly
+                let mut offset: usize = match parse_options.resolve_rva {
+                    true => idd.pointer_to_raw_data as usize,
+                    false => idd.address_of_raw_data as usize,
+                };
+
+                // See data specification:
+                // https://github.com/dotnet/runtime/blob/97ddb55e3adde20ceac579d935cef83cfe996169/docs/design/specs/PE-COFF.md#embedded-portable-pdb-debug-directory-entry-type-17
+                offset += 4; // skip `signature`
```
Swatinem, line 321: You should still check that it matches though :-)

Swatinem:
As for the Symbolicator integration, I believe there are two ways to approach this.

1. As you mentioned, we will not use the PE itself anymore, so we can decide to "unpack" the Portable PDB when fetching the object here: https://github.com/getsentry/symbolicator/blob/9014ed70d127fb33d051b37efa781b01e7e598e2/crates/symbolicator-service/src/services/objects/data_cache.rs#L135-L172
That way, we can use it "cheaply" for source lookups, and directly create a ppdb cache from it.
The advantage here is that the source lookups and ppdb cache creation are unchanged, and fast (no decompression needed anymore). The disadvantage is that we throw away the "outer" PE.
2. You could do the `get_embedded_ppdb` dance both in source lookup, and in ppdb conversion.
This is a bit more work, and would still have to do decompression on every source lookup. But it has the advantage of keeping the original PE intact.

Given that we (at this time) are not using the outer PE for anything, I would recommend to just go with 1 and unpack directly when downloading.

symbolic-ppdb/tests/test_ppdb.rs
Swatinem: Oh, thats why its zero-initialized I guess? Well even more reason to return a `Vec<u8>` and remove that footgun.

symbolic-debuginfo/src/pe.rs
```diff
@@ -276,6 +278,60 @@ impl<'data> PeObject<'data> {
...
         }
     }
+
+    /// Returns the raw buffer of Embedded Portable PDB Debug directory entry, if any.
+    pub fn embedded_ppdb(&self) -> Result<Option<PeEmbeddedPortablePDB<'data>>, PeError> {
+        // Note: This is currently not supported by goblin, see https://github.com/m4b/goblin/issues/314
+        let opt_header = match self.pe.header.optional_header {
+            Some(v) => v,
+            None => return Ok(None),
+        };
```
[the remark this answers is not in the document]
vaind: ok, changed manually also for the following cases

symbolic-ppdb/tests/test_ppdb.rs
vaind: ok changed

symbolic-debuginfo/src/pe.rs
vaind: that it matches _what_? It doesn't match the pe.header.signature...

vaind:
> As for the Symbolicator integration, I believe there are two ways to approach this.
> 
> 1. As you mentioned, we will not use the PE itself anymore, so we can decide to "unpack" the Portable PDB when fetching the object here: [getsentry/symbolicator@`9014ed7`/crates/symbolicator-service/src/services/objects/data_cache.rs#L135-L172](https://github.com/getsentry/symbolicator/blob/9014ed70d127fb33d051b37efa781b01e7e598e2/crates/symbolicator-service/src/services/objects/data_cache.rs#L135-L172)
>    That way, we can use it "cheaply" for source lookups, and directly create a ppdb cache from it.
>    The advantage here is that the source lookups and ppdb cache creation are unchanged, and fast (no decompression needed anymore). The disadvantage is that we throw away the "outer" PE.
> 2. You could do the `get_embedded_ppdb` dance both in source lookup, and in ppdb conversion.
>    This is a bit more work, and would still have to do decompression on every source lookup. But it has the advantage of keeping the original PE intact.
> 
> Given that we (at this time) are not using the outer PE for anything, I would recommend to just go with 1 and unpack directly when downloading.

I was thinking of a combination of both. The first point for actual symbolication, and the second one (or just a part of it) to display the correct information in the UI (that it has sources)

symbolic-debuginfo/src/pe.rs
```diff
@@ -318,14 +309,33 @@ impl<'data> PeObject<'data> {
 
                 // See data specification:
                 // https://github.com/dotnet/runtime/blob/97ddb55e3adde20ceac579d935cef83cfe996169/docs/design/specs/PE-COFF.md#embedded-portable-pdb-debug-directory-entry-type-17
-                offset += 4; // skip `signature`
+                // let signature = &self.data[offset..(offset + 4)];
+                let mut signature: [u8; 4] = [0; 4];
+                self.data
+                    .gread_inout(&mut offset, &mut signature)
+                    .map_err(PeError::new)?;
+                if signature != "MPDB".as_bytes() {
```
Swatinem, line 317:
```suggestion
                let signature: [u8; 4] = self.data
                    .gread(&mut offset)
                    .map_err(PeError::new)?;
                if signature != b"MPDB" {
```
First time I saw `gread_inout`.

symbolic-debuginfo/src/pe.rs
```diff
@@ -318,14 +309,33 @@ impl<'data> PeObject<'data> {
...
+                    )));
+                }
                 let uncompressed_size: u32 = self
                     .data
                     .gread_with(&mut offset, LE)
                     .map_err(PeError::new)?;
-                let compressed_size = idd.size_of_data - 8; // 8 = the number bytes we have just read.
+
+                // 8 == the number bytes we have just read.
+                let compressed_size = idd.size_of_data as usize - 8;
```
Swatinem, line 328: Actually, this can underflow if `size_of_data` (which is untrusted input) is <8.
It then overflows in `offset + compressed_size`, causing a panic because end < start.

```suggestion
                let Some(compressed_size) = (idd.size_of_data as usize).checked_sub(8) else { ... };
```

I know I’m very pedantic about this, just want to be sure we are as safe as possible. Also reminds me that we should maybe try to fuzz all our portable pdb work at some point just in case.
I have fuzzed the rest of symbolic quite some time ago, fixing some of these issues.

Swatinem (commented):
almost there :-)

symbolic-debuginfo/src/pe.rs
vaind: not the same thing - `gread` doesn't work with slices

vaind:
```
error[E0277]: the trait bound `[u8; 4]: TryFromCtx<'_, _>` is not satisfied
   --> symbolic-debuginfo\src\pe.rs:313:42
    |
313 |                 let signature: [u8; 4] = self.data.gread(&mut offset).map_err(PeError::new)?;
    |                                          ^^^^^^^^^ ----- required by a bound introduced by this call
    |                                          |
    |                                          the trait `TryFromCtx<'_, _>` is not implemented for `[u8; 4]`
    |
    = help: the trait `TryFromCtx<'a, usize>` is implemented for `&'a [u8]`
note: required by a bound in `gread`
   --> C:\Users\dlugo\.cargo\registry\src\github.com-1ecc6299db9ec823\scroll-0.11.0\src\pread.rs:88:21
    |
88  |     fn gread<'a, N: TryFromCtx<'a, Ctx, Self, Error = E>>(
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ required by this bound in `gread`
    ```

symbolic-debuginfo/src/pe.rs
vaind: added an explicit check above

Swatinem (approved):

loewenheim (approved):

PR #755, opened 2023-02-02
ref(smcache): Add context lines and source printing to our example

kamilogorek:
Mostly to ease debugging with our support team.

#skip-changelog


loewenheim (approved):

Swatinem (commented):
IMO a view like:
```
1: ...
2: ...
>: ...
3: ...
4: ...
```
might be a good idea. I always find it very confusing to have the pre-context after the context.
We will only have `line_contents` anyway if we also have a `file`.
So you can always show either the full contents or shortened context, or a "No source context available" message.

Issue #750, opened 2023-01-30
Embedded symbols/sources not found


mattjohnsonpint:
### Environment

Sentry-CLI 2.12.0

### Steps to Reproduce

1. Using .NET 6 or .NET 7 SDK, create a new class library
  ```sh
  dotnet new classlib -n MyLib
  cd MyLib
  ```

2. Modify `MyLib.csproj` and add the following to the existing property group, which tells the compiler to embed debug info into the dll/exe itself.  [Reference here.](https://learn.microsoft.com/dotnet/csharp/language-reference/compiler-options/code-generation#debugtype)
  ```xml
  <DebugType>embedded</DebugType>
  ```

3. Build in release configuration:
  ```sh
  dotnet build -c Release
  ```

4. Check the resulting file with Sentry CLI:
  ```sh
  sentry-cli dif check bin/Release/net7.0/MyLib.dll
  ```

### Expected Result

Should have found debug symbols and sources.  This works if we build without embedded sources and check the pdb file, but not when we use embedded sources and check the dll.

### Actual Result

Nothing found in the dll.

```
Debug Info File Check
  Type: pe library
  Contained debug identifiers:
    > Debug ID: 940225b8-316b-46cd-b6c5-4e6a5a6613d7-1
      Code ID:  8be87e768000
      Arch:     x86
  Contained debug information:
    > none
  Usable: no (missing debug or unwind information)
```

Doesn't work with `dif upload` either.

mattjohnsonpint:
@Swatinem

mattjohnsonpint:
I presume this is a symbolic issue, not specifically sentry-cli.  But let me know.  Thanks.

vaind:
Duplicate of https://github.com/getsentry/sentry-dotnet/issues/2073 though that's in the dotnet repo

Swatinem:
There is a bunch of PRs that I never finished to read the "correct" deterministic debug id from the dll, as well as provide access to the underlying .NET metadata from the dll: https://github.com/getsentry/symbolic/pull/658 and https://github.com/getsentry/symbolic/pull/659
I believe if we finish those up, we should be able to use the `dll` directly, instead of requiring a `pdb`. Might be some work though.
Can you provide us with a small fixture to test against?

vaind:
[...]

PR #757, opened 2023-02-07
feat: PE embedded PPDB decompression to a custom writer

vaind:
Part of #750

### DIFF
--- a/symbolic-debuginfo/src/pe.rs
+++ b/symbolic-debuginfo/src/pe.rs
@@ -3,9 +3,7 @@
 use std::borrow::Cow;
 use std::error::Error;
 use std::fmt;
-use std::fs::File;
-use std::io::BufWriter;
-use std::io::Read;
+use std::io::Write;
 
 use gimli::RunTimeEndian;
 use goblin::pe;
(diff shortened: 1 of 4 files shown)


symbolic-debuginfo/src/pe.rs
```diff
@@ -516,4 +517,19 @@ impl<'data, 'object> PeEmbeddedPortablePDB<'data> {
         }
         Ok(output)
     }
+
+    /// Reads the Portable PDB contents into the provided file.
+    pub fn decompress_to_file(&self, output: File) -> Result<File, PeError> {
+        use std::io::prelude::*;
+        let buf_writer = BufWriter::new(output);
+        let mut decoder = flate2::write::DeflateDecoder::new(buf_writer);
```
Swatinem, line 525: Wow, that is confusing, the same name in different modules.

But this looks interesting. We might as well just make this generic and have a `decompress_to<W>(...) where W: Write`
And you can just give that a `BufWriter<File>` or a `Vec`, both should be `Write`.

Swatinem (approved):
I  would say its up to you if you want to keep the convenience methods.

Usually its the callers responsibility what kind of `Write` they want to provide, if they want to use a `BufWriter` at all, etc.

symbolic-debuginfo/src/pe.rs
vaind: Agreed, I've made the change

PR #776, opened 2023-03-27
fix: il2cpp source mapping without boundary

vaind:


symbolic-il2cpp/src/line_mapping/mod.rs
```diff
@@ -116,20 +119,170 @@ mod tests {
...
+        //<source_info:/Scripts/AdditionalButtons.cs:44>                        | /Scripts/AdditionalButtons.cs:43
+        // }                                                                    | /Scripts/AdditionalButtons.cs:43
+        return;                                                                 | /Scripts/AdditionalButtons.cs:44
+    }                                                                           | /Scripts/AdditionalButtons.cs:44
+}                                                                               | /Scripts/AdditionalButtons.cs:44
+        "#.trim();
+        let cpp_lines = cpp_source.split('\n').collect::<Vec<_>>();
+        let clean_cpp_lines = cpp_lines
+            .iter()
+            .map(|line| line.split_once('|').unwrap().0.trim_end())
```
Swatinem, line 265: there is also `rsplit_once` here and below.

symbolic-il2cpp/src/line_mapping/mod.rs
```diff
@@ -116,20 +119,170 @@ mod tests {
...
+        ",
+        );
+        assert_eq!(parsed_mapping.lookup("main.cpp", 0), None);
+        assert_eq!(parsed_mapping.lookup("main.cpp", 1), None);
         assert_eq!(parsed_mapping.lookup("main.cpp", 2), None);
+        assert_eq!(parsed_mapping.lookup("main.cpp", 3), None);
+        assert_eq!(parsed_mapping.lookup("main.cpp", 4), None);
         assert_eq!(parsed_mapping.lookup("main.cpp", 5), Some(("main.cs", 17)));
-        assert_eq!(parsed_mapping.lookup("main.cpp", 12), None);
+        assert_eq!(parsed_mapping.lookup("main.cpp", 6), Some(("main.cs", 17)));
```
Swatinem, line 130: this might be a bit overkill here to enumerate every line. the important one I guess is that `12` now has a mapping.

Swatinem (approved):

symbolic-il2cpp/src/line_mapping/mod.rs
vaind: it may be but it also makes the expectation clear (and removes the guesswork) so if you don't mind I'd keep it there for each line

PR #778, opened 2023-03-28
Make SourceBundle writer deterministic

Swatinem:
The default ZipWriter/FileOptions is using the current timestamp depending on feature flags. Now we explicitly default to a deterministic timestamp for files.

I verified locally that this fixes https://github.com/getsentry/sentry-cli/issues/831 🎉


kamilogorek (approved):
🥲

PR #779, opened 2023-03-30
build(unreal): Remove outdated `time` dependency

jan-auer:
This removes default features from `chrono`, which had a dependency on an old,
vulnerable version of `time` that was no longer used. Additionally, this bumps
`anylog` for the same reason.

### DIFF
--- a/symbolic-unreal/Cargo.toml
+++ b/symbolic-unreal/Cargo.toml
@@ -26,10 +26,9 @@ all-features = true
 serde = ["dep:serde", "chrono/serde"]
 
 [dependencies]
-anylog = "0.6.3"
+anylog = "0.6.4"
 bytes = "1.4.0"
-# this is still used for compatibility with `anylog`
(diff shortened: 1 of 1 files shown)


symbolic-unreal/Cargo.toml
```diff
@@ -26,10 +26,9 @@ all-features = true
 serde = ["dep:serde", "chrono/serde"]
 
 [dependencies]
-anylog = "0.6.3"
+anylog = "0.6.4"
 bytes = "1.4.0"
-# this is still used for compatibility with `anylog`
```
Swatinem, line 31: do you want to keep this comment?

Swatinem (approved):
looks good. so the problem is only with `time`, and not with `chrono` itself? I thought `chrono` itself was also flagged by RUSTSEC?

jan-auer:
Chrono is fine, it's just the outdated `time` dependency that was the issue.

PR #780, opened 2023-04-12
Fix the problem that some symbols are not parsed

xuezhulian:
debug_line example:
0x0000000102704a68      0     13      3   0             0 
0x0000000102704a6c    313     12      3   0             0 
0x0000000102704a78      0     12      3   0             0 
0x0000000102704a84    314      1      3   0             0  is_stmt
0x0000000108cc1bfc    314      1      3   0             0  is_stmt end_sequence
0x0000000102704e40     53      0      3   0             0  is_stmt
0x0000000102704e4c     54      6      3   0             0  is_stmt prologue_end
0x0000000102704e58      0      6      3   0             0 
0x0000000102704e5c     54      6      3   0             0 
0x0000000102704e68      0      6      3   0             0 
0x0000000102704e6c     54      5      3   0             0 
0x0000000102704e70      0      5      3   0             0 
0x0000000102704e78     54      5      3   0             0 
0x0000000102704e7c     54      5      3   0             0  end_sequence
0x000000010274c3c0    389      0      3   0             0  is_stmt
0x000000010274c3cc    390      6      3   0             0  is_stmt prologue_end
0x000000010274c3d8      0      6      3   0             0 
0x000000010274c3dc    390      6      3   0             0 
0x000000010274c3e8      0      6      3   0             0 
0x000000010274c3ec    390      5      3   0             0 
0x000000010274c3f0      0      5      3   0             0 
0x000000010274c3f8    390      5      3   0             0 
0x000000010274c3fc    390      5      3   0             0  end_sequence
0x00000001072b37b0     49      0      3   0             0  is_stmt
0x00000001072b37bc     50      6      3   0             0  is_stmt prologue_end

pc start with 0x0000000107 will hit the seq  with the range 0x0000000102704a68 0x0000000108cc1bfc. the symbols for these pc will not to be resolved

### DIFF
--- a/symbolic-debuginfo/src/dwarf.rs
+++ b/symbolic-debuginfo/src/dwarf.rs
@@ -347,7 +347,12 @@ impl<'d> DwarfLineProgram<'d> {
             let from = match seq.rows.binary_search_by_key(&range.begin, |x| x.address) {
                 Ok(idx) => idx,
                 Err(0) => continue,
-                Err(next_idx) => next_idx - 1,
+                Err(next_idx) => {
+                    if next_idx == seq.rows.len() {
+                        continue;
+                    }
+                    next_idx - 1
+                }
             };
 
             let len = seq.rows[from..]


xuezhulian:
Thank you for your reply. 0x0000000108cc1bfc is an abnormal but not fixed value, perhaps related to a packet size reduction technique called segment migration. This value appears more than 13425 times in our dwarf.  Segment migration is universal around me, and Sentry is becoming popular, maybe you should know this and take some measures. Thank you again.

xuezhulian:
`if segment.name().ok() != Some("__TEXT")` Problems also arise here. Segment migration will cause all functions in symtab to be filtered.

Swatinem:
> is an abnormal but not fixed value

Well if it is indeed abnormal, and appears multiple times in the file, it does look like a bug in the compiler / linker that is generating these line programs.

Can you say which compiler / linker and which flags you are using?

That might also clear up the confusion as I have never heard about "packet size reduction technique called segment migration", which might be a translation problem :-)

Swatinem:
So if all this is doing is a `-rename_section` linker flag, I would say that the linker does something wrong here. It is supposed to patch all the line program addresses correctly, but it looks like it does not do that?

xuezhulian:
I have some new discoveries, and I'll discuss it with you when I've figured it out

ashwoods:
Closing this due to inactivity. @xuezhulian Please feel free to reopen when anything new comes up.

PR #788, opened 2023-05-15
Update macos builder to latest

Swatinem:
Well, looks like the 10.15 CI host/job does not start in any reasonable time, so lets update

#skip-changelog


jan-auer (commented):
Can you take the approach from Relay instead, so we do not have to deprecate support for 10.15 at this point?
https://github.com/getsentry/relay/pull/1865

Swatinem:
Sure, building with the latest runner, but building the same `py_platform` as before.

PR #790, opened 2023-05-24
Remove the bitvec crate.

afranchuk:
It is used only in one place, and an in-crate definition is simple enough.

We are using `symbolic` and are auditing our third-party crates. To reduce the audit surface of `symbolic`, this removes `bitvec` and replaces it with an implementation that should be as performant. `bitvec` is a huge crate, so it is preferable to remove it since it is barely used.

If this is not desirable, we can also patch `bitvec` with an API-equivalent interface on our side, but that is an uglier solution and may be more burdensome in the future.

### DIFF
--- a/symbolic-debuginfo/Cargo.toml
+++ b/symbolic-debuginfo/Cargo.toml
@@ -76,10 +76,9 @@ sourcebundle = [
 # JavaScript stuff
 js = []
 # WASM processing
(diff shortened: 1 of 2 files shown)


loewenheim (approved):
LGTM, thank you for the contribution!

kamilogorek:
It's an internal change, so we can skip changelog for it.

PR #797, opened 2023-06-24
ref: add some mypy typing to symbolic

asottile-sentry:


Swatinem (approved):
thanks for doing this!

I’m wondering how much of this we would get for free if we used something like pyo3 or similar…

PR #819, opened 2023-10-24
Add experimental symbolic-py pyo3/maturin crate

Swatinem:
This is an attempt to replace the manual cabi/py wrapper with one built using pyo3 and maturin.


daddycocoaman:
Hi,

Just curious but does this effort mean that `symbolic` is expected to maintain Python/C API support with all the necessary functions exposed?

Swatinem:
No, on the contrary.

This is just an early experiment, which eventually feeds into https://github.com/getsentry/rfcs/pull/119.
The goal would be to completely deprecate the python bindings to symbolic directly, and rather have sentry specific python bindings that have no SemVer guarantees whatsoever.

The python bindings in symbolic itself are:
- completely outdated (binding-wise)
- expose only a subset of symbolic functionality on the one hand and
- expose more than the symbolic rust crates (things like proguard)
- are a nightmare to maintain, and cause frequent headaches to release to pypi

All that being said, I would consider the symbolic python bindings as deprecated, and we will completely remove them at one point or another in a future semver-major release.

Swatinem:
As this PR has shown, creating python bindings with maturin/pyo3 is really straight forward, so you can just pick whatever functionality want from the symbolic Rust crate and create your own bindings directly if you really want that.

Swatinem:
This was now released as https://github.com/getsentry/ophio

PR #821, opened 2023-11-17
fix empty parameters list issue

viglia:
Use a specific boolean flag to decide whether to deobfuscate based on parameters list or line numbers.

Empty parameters list should not be used to make this discrimination, but at the same Time `Option<T>` type is not FFI safe therefore we cannot use it for the Rust bindings.

### DIFF
--- a/py/symbolic/proguard.py
+++ b/py/symbolic/proguard.py
@@ -81,7 +81,7 @@ class ProguardMapper(RustObject):
         return output if len(output[0]) > 0 and len(output[1]) > 0 else None
 
     def remap_frame(
-        self, klass: str, method: str, line: int, parameters: str = ""
+        self, klass: str, method: str, line: int, parameters: str = "", use_parameters: bool = False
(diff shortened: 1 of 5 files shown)


py/symbolic/proguard.py
```diff
@@ -81,7 +81,12 @@ def remap_method(self, klass: str, method: str) -> Tuple[str, str] | None:
         return output if len(output[0]) > 0 and len(output[1]) > 0 else None
 
     def remap_frame(
-        self, klass: str, method: str, line: int, parameters: str = ""
+        self,
+        klass: str,
+        method: str,
+        line: int,
+        parameters: str = "",
+        use_parameters: bool = False,
```
Swatinem, line 89: you can still use `Optional` here, but then internally split that into a bool + str

viglia: good point!! 👍🏻

Swatinem (approved):

PR #823, opened 2023-11-30
enhance: make mapping by params initialization optional (android deobfuscation)

viglia:
This follow the changes in https://github.com/getsentry/rust-proguard/pull/34

This will allow us to make `mappings_by_params` initialization optional so that we don't incur in unnecessary memory increase in those part of sentry services where deobfuscation based on parameters list instead of line numbers is not necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+**Internal**
+
+**Fixes**
+
+**Features**
+
+- Make mapping by params initialization optional ([#823](https://github.com/getsentry/symbolic/pull/823))
+
 ## 12.7.1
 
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+**Internal**
```
Swatinem, line 3: You have to prefix this with `## Unreleased`, otherwise the auto-changelog will not recognize this.

I recently fixed the auto-changelog after the fact: https://github.com/getsentry/symbolic/commit/6970909942af44aaadd2fc4d8c91133d6b3dd405

Swatinem (approved):

PR #833, opened 2024-03-26
chore: update tokio to latest version

anonrig:
Fixes https://github.com/getsentry/symbolic/issues/832


Swatinem (approved):
you actually have to update the `Cargo.lock`, otherwise this doesn’t do anything

PR #866, opened 2024-09-04
Revert "all-repos: update actions/upload-artifact to v4"

szokeasaurusrex:
Reverts getsentry/symbolic#858. `upload-artifacts` v4 is causing our release builds to fail as it is incompatible with our release build process.

We needed to [revert this change in Sentry CLI](https://github.com/getsentry/sentry-cli/pull/2119) for the same reason.

#skip-changelog


loewenheim (approved):

joshuarli:
ref: https://github.com/getsentry/craft/issues/552

PR #895, opened 2025-02-03
fix(macho): Parse `debug_str_offs` section

jjbayer:
For Mach-O files, we need to map the section header name for debug_str_offsets:

https://shadow.github.io/docs/rust/object/read/trait.Object.html#provided-methods

> • if “.debug_str_offsets” is requested for a Mach-O object file, then the actual section name that is searched for is “__debug_str_offs”.

Fixes: https://github.com/getsentry/sentry/issues/79268


symbolic-debuginfo/src/dwarf.rs
```diff
@@ -1487,3 +1487,23 @@ impl<'s> Iterator for DwarfFunctionIterator<'s> {
...
+    #[test]
+    fn test_loads_debug_str_offsets() {
+        // File generated using dsymutil
+
+        let data = std::fs::read("tests/fixtures/helloworld").unwrap();
+
+        let obj = MachObject::parse(&data).unwrap();
+
+        let sections = DwarfSections::from_dwarf(&obj);
+        assert_eq!(sections.debug_str_offsets.data.len(), 48);
```
jjbayer, line 1507: This assertion fails without `map_section_name`.

loewenheim (approved):

PR #896, opened 2025-02-04
build: Use upload-artifacts v4

jjbayer:
Cargo-culted this PR off https://github.com/getsentry/ophio/pull/69, I don't know if it will actually work.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,10 +36,13 @@ jobs:
           # consumed by cargo and setup.py to obtain the target dir
           CARGO_BUILD_TARGET: ${{ matrix.target }}
 
-      - uses: actions/upload-artifact@v3.1.1
+      - uses: actions/upload-artifact@v4
         with:
-          name: ${{ github.sha }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -63,10 +66,13 @@ jobs:
         name: Build in Docker (x86_64)
         run: make wheel-manylinux IMAGE=quay.io/pypa/"$MANYLINUX_VERSION"_x86_64
 
-      - uses: actions/upload-artifact@v3.1.1
+      - uses: actions/upload-artifact@v4
         with:
-          name: ${{ github.sha }}
+          name: artifact-linux-${{ matrix.build-arch }}
```
jjbayer, line 71: Not sure if the name matters here. `matrix.target` seems to be unavailable so I used `matrix.build-arch` instead.

szokeasaurusrex: I'm not sure either

loewenheim (approved):
Looks reasonable to me

PR #903, opened 2025-03-12
feat(minidump-stackwalk): Allow specifying multiple sources

Dav1dde:
Currently using this to debug a minidump where I want some debug info loaded from the binary but others from system libraries.

The patch makes it possible to specify multiple sources.


loewenheim (approved):
Good idea! Changelog entry wouldn't hurt, though.

PR #907, opened 2025-04-02
feat(unreal-engine-crash): Add extraction command to cli utility

Dav1dde:
Make it possible to extract individual files from an unreal crash report via the commandline utility.


loewenheim (approved):
LGTM, what is the impetus for this feature?

Dav1dde:
> LGTM, what is the impetus for this feature?

Needed to analyze a minidump contained in an unreal crash report, so I had to get it out somehow.

PR #926, opened 2025-06-18
semgrep: exclude example and test directories

mdtro:
A lot of `symbolic`'s findings are in these example and test directories. Using this ignore file will auto-ignore findings here and reduce the noise.

### DIFF
--- a/.semgrepignore
+++ b/.semgrepignore
@@ -1,0 +1,4 @@
+:include .gitignore
+
+examples/
+**/tests/


mdtro:
> Is there no way to turn it off for tests on a code level?
> 
> Unit tests are in the same files as the production code, but they are in clearly separated test modules.

It's a bit difficult to do globally, but we're working on fixing the detection rules for Rust to be aware of the `#[cfg(test)]` and `#[test]` blocks. We can also leverage Semgrep's "memories" AI feature to teach it out test code.

PR #929, opened 2025-07-18
build: switch to python3.11 for releases builds

olksdr:
Python 3.6 was removed from the image, and is not longer available. 
Switching to Python 3.11 for build releases.


olksdr:
> Did you try to build this locally, in Relay we had issues with (build-)dependencies missing.

Fixed dependencies same way as in Relay.

PR #935, opened 2025-09-10
feat: add support for dynamic symbols when DYNAMIC segment is stripped

nsavoire:
If dynamic symbols cannot be parsed from the DYNAMIC segment, try to parse them from the section headers.

This allows to support "debug info" files that do not have a regular symbol table nor debug information but only a dynamic symbol table obtained with: `llvm-objcopy --only-keep-debug --set-section-type .dynstr=3 --set-section-type .dynsym=11 <file> <file>-dynsyms_only`


nsavoire:
> Would you mind adding a changelog entry like
> 
> ```
> ## Unreleased
> 
> - feat(elf): Added support for dynamic symbols when DYNAMIC segment is missing ([#935](https://github.com/getsentry/symbolic/pull/935))
> ```

Done !

PR #947, opened 2025-12-10
ref(symcache): Make parsing backwards-compatible

loewenheim:
This puts in some indirection to make it so that old symcache versions remain parseable, even if the on-disk format changes. For the two currently supported versions, 7 and 8, this is massive overkill because they only differ in how string lengths are serialized to the string table. The intention is to enable us to add data to the format without having to throw out every old symcache immediately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## Unreleased
+
+**Improvements**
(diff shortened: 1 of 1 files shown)


loewenheim:
@jjbayer I implemented your suggestion about deduplication and it's a substantial improvement.

PR #964, opened 2026-03-27
fix(goblin): disable `PE` import table parser

supervacuus:
This is a follow-up to #960 and depends on an upstream change in `goblin` that has not yet been accepted (https://github.com/m4b/goblin/pull/522).

It makes the integration of PE+DWARF debug companions into the DIF pipeline practical (see https://github.com/getsentry/sentry-cli/pull/3240)


supervacuus:
Just FYI, I proposed an alternative PR upstream that fixes the underlying issue instead of adding an import parser flag: https://github.com/m4b/goblin/pull/527

supervacuus:
The option was merged upstream (https://github.com/m4b/goblin/pull/522) and released with [0.10.6](https://github.com/m4b/goblin/blob/HEAD/CHANGELOG.md#0106---2026-5-25).

`0.10.6` also contains the fix (https://github.com/m4b/goblin/pull/527) for the performance regression that became apparent when reading PE+DWARF debug companions with permissive PR parsing enabled. But disabling the import parser also prevents us from running into it.

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+**Fixes**
+
+- disable `PE` import table parser ([#964](https://github.com/getsentry/symbolic/pull/964))
```
Dav1dde: Maybe something like:

```suggestion
- Disable `PE` import table parser and be more permissive during parsing. ([#964](https://github.com/getsentry/symbolic/pull/964))
```

Dav1dde (approved):

supervacuus:
Before I merge this, to verify, since I created this PR initially from a different premise:

* When I merged https://github.com/getsentry/symbolic/pull/960, it was accidentally released as a minor/patch bump, which broke clients.
* Then it was reverted (https://github.com/getsentry/symbolic/pull/966) and deferred until the next major to keep `12.x` clean from breakage.
* Then https://github.com/getsentry/symbolic/pull/976 bumped `gimli` again for `13.x` to the `0.10.5` from #960, but _did not re-enable_ the permissive parsing mode for PE.
* So, this PR reintroduces this behavior, which was thankfully noted in the change log by @Dav1dde.

However, the tests (+snapshots and fixtures) I introduced in #960 are now missing, right? That should still be done, right? Otherwise, a sensible summary? Regarding release plans and potential bumps, are we good?

Dav1dde:
> However, the tests (+snapshots and fixtures) I introduced in https://github.com/getsentry/symbolic/pull/960 are now missing, right? That should still be done, right?

Good catch, yes, the tests are gone and it would be really useful to have them back!

> Regarding release plans and potential bumps, are we good?

Yep, that looks all good to me. Let's wait for @loewenheim as well, he has a lot more context.

loewenheim (approved):

loewenheim:
Everything seems in order now, thanks for picking this up again @supervacuus!

PR #967, opened 2026-04-13
feat: Add Unreal4File::into_bytes

elramen:
Ref: https://linear.app/getsentry/issue/SYMBOL-20/expose-bytes-instance-from-unreal4file

### DIFF
--- a/symbolic-unreal/src/container.rs
+++ b/symbolic-unreal/src/container.rs
@@ -355,6 +355,11 @@ impl Unreal4File {
         &self.bytes
     }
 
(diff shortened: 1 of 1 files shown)


loewenheim (approved):

loewenheim:
I believe a changelog entry is actually warranted here (it's adding a new public method).

elramen:
> I believe a changelog entry is actually warranted here (it's adding a new public method).

Merged at the same time this comment popped up, will create a separate PR

PR #974, opened 2026-04-20
ref(js): Allow magic comments to appear anywhere

### LINKED ISSUE #969: discover_sourcemaps_location misses directives not at column 0 (line.starts_with is too strict)
## Summary

`discover_sourcemaps_location` in [`symbolic-debuginfo/src/js.rs`](https://github.com/getsentry/symbolic/blob/master/symbolic-debuginfo/src/js.rs) uses `line.starts_with("//# sourceMappingURL=")` to detect the directive. This silently misses the directive when it appears at the end of an existing line — a valid and common output of minifiers.

## The bug

```rust
for line in contents.lines().rev() {
    if line.starts_with("//# sourceMappingURL=") || line.starts_with("//@ sourceMappingURL=") {
```

When a minifier appends the directive directly to the last statement without inserting a newline first, the final line of the file looks like:

```
...minified_code...//# sourceMappingURL=app.js.map
```

`line.starts_with("//# sourceMappingURL=")` returns `false` here, so the source map is never found and Sentry silently fails to load it.

## Why this is a spec bug, not a "wrong input" bug

The [ECMAScript Source Map specification (rev3)](https://github.com/tc39/source-map-spec/blob/main/source-map-rev3.md#linking-generated-code-to-source-maps) says:

> The generated code **may include a line** at the end of the source, with the following form:
> `//# sourceMappingURL=<url>`

The spec describes the *form of the annotation*, not the column at which it must begin. It does not say the directive must occupy an entire line by itself or that it must start at column 0. All major JavaScript runtimes (V8/Chrome, SpiderMonkey/Firefox, JavaScriptCore/Safari, Node.js) search for the directive using `indexOf`/`contains`, not `startsWith`, confirming the permissive interpretation is correct.

## Fix

Replace `starts_with` with `contains` (or `find`/`rfind`) when searching for the directive within a line:

```rust
for line in contents.lines().rev() {
    let marker = if let Some(pos) = line.find("//# sourceMappingURL=") {
        Some(("//# sourceMappingURL=".len(), pos))
    } else if let Some(pos) = line.find("//@ sourceMappingURL=") {
        Some(("//@ sourceMappingURL=".len(), pos))
    } else {
        None
    };

    if let Some((prefix_len, pos)) = marker {
        let url = line[pos + prefix_len..].trim();
        let without_query = url.split_once('?').map(|x| x.0).unwrap_or(url);
        let without_fragment = without_query.split_once('#').map(|x| x.0).unwrap_or(without_query);
        return Some(without_fragment);
    }
}
```

## References

- ECMAScript Source Map spec (rev3): https://github.com/tc39/source-map-spec/blob/main/source-map-rev3.md#linking-generated-code-to-source-maps
- TC39 ECMA-426 (living spec): https://tc39.es/source-map/#linking-generated-code-to-source-maps

Dav1dde:
Updates the JS magic comment parsing to not only consider magic comments which are at the start of a line but anywhere in the source. The spec is not 100% clear of whether the magic comment must be at the beginning of a line, but major vendors have implementations allowing the comment to appear anywhere in the source.

The code will use the last occurrence of the magic comment in the JS source.

This will now also be more lenient with whitespaces, but consider the value of the magic comment to be separated by a whitespace. This follows the pattern defined in V8 `//[#@]\s<name>=\s*<value>\s*.*`: https://github.com/v8/v8/blob/4532c6bf8ac0838c9b3e85d6ae6078a1bc1dcb0c/src/parsing/scanner.cc#L279-L281

Fixes: #969

### DIFF
--- a/symbolic-debuginfo/src/js.rs
+++ b/symbolic-debuginfo/src/js.rs
@@ -12,20 +12,18 @@ use serde::Deserialize;
 ///
 /// Any query string or fragments the URL might contain will be stripped away.
 pub fn discover_sourcemaps_location(contents: &str) -> Option<&str> {
-    let finder = MagicCommentFinder::source_mapping_url().allow_deprecated_at_variant();
+    let url = MagicCommentFinder::source_mapping_url()
+        .allow_deprecated_at_variant()
+        .find(contents)?;
 
-    if let Some(url) = finder.find(contents) {
-        // The URL might contain a query string or fragment. Strip those away before recording the URL.
-        let without_query = url.split_once('?').map(|x| x.0).unwrap_or(url);
-        let without_fragment = without_query
-            .split_once('#')
-            .map(|x| x.0)
-            .unwrap_or(without_query);
+    // The URL might contain a query string or fragment. Strip those away before recording the URL.
+    let without_query = url.split_once('?').map(|x| x.0).unwrap_or(url);
+    let without_fragment = without_query
+        .split_once('#')
+        .map(|x| x.0)
+        .unwrap_or(without_query);
 
-        return Some(without_fragment);
-    }
-
-    None
+    Some(without_fragment)
 }
 
 /// Quickly reads the embedded `debug_id` key from a source map.
@@ -57,7 +55,7 @@ pub fn discover_debug_id(contents: &str) -> Option<DebugId> {
 
 /// A helper utility which allows searching for magic comments in JavaScript sources.
 ///
-/// The finder will optionally consider the `//#` variant as well as the deprecated `//@` variant.
+/// The finder will optionally consider the deprecated `//@` variant in addition to the official `//#` variant.
 ///
 /// Generally considered a magic comment is a comment following the pattern: `//[#@]\s<name>=\s*<value>\s*.*`
 struct MagicCommentFinder<'a> {
--- a/symbolic-debuginfo/src/js.rs
+++ b/symbolic-debuginfo/src/js.rs
@@ -109,11 +109,7 @@ impl<'a> MagicCommentFinder<'a> {
         // Trim whitespaces after the `=`.
         let value = value.trim_ascii_start();
         // Split until the next whitespace:
-        let value = value
-            .split(u8::is_ascii_whitespace)
-            .next()
-            // If there is no whitespace, assume until end.
-            .unwrap_or(value);
+        let value = value.split(u8::is_ascii_whitespace).next()?;
 
         // This should never fail, the input was a valid string, we only trimmed characters in the
         // ascii range.
--- a/symbolic-debuginfo/src/js.rs
+++ b/symbolic-debuginfo/src/js.rs
@@ -119,7 +119,7 @@ impl<'a> MagicCommentFinder<'a> {
 
         // This should never fail, the input was a valid string, we only trimmed characters in the
         // ascii range.
-        str::from_utf8(value).ok()
+        std::str::from_utf8(value).ok()
     }
 }
 
--- a/symbolic-debuginfo/src/js.rs
+++ b/symbolic-debuginfo/src/js.rs
@@ -5,26 +5,26 @@
 //! with source maps this module is insufficient.
 
 use debugid::DebugId;
+use memchr::memmem::FinderRev;
 use serde::Deserialize;
 
 /// Parses a sourceMappingURL comment in a file to discover a sourcemap reference.
 ///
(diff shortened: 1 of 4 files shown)


symbolic-debuginfo/src/js.rs
```diff
@@ -50,19 +50,275 @@ pub fn discover_sourcemap_embedded_debug_id(contents: &str) -> Option<DebugId> {
...
+        // Split until the next whitespace:
+        let value = value
+            .split(u8::is_ascii_whitespace)
+            .next()
+            // If there is no whitespace, assume until end.
+            .unwrap_or(value);
+
+        // This should never fail, the input was a valid string, we only trimmed characters in the
+        // ascii range.
+        str::from_utf8(value).ok()
```
Dav1dde: I am pretty confident that this holds true, but I am not sure if I am missing something. If we're sure this is 100% correct, we could use `unsafe from_utf8_unchecked`.

symbolic-debuginfo/src/js.rs
```diff
@@ -50,19 +50,275 @@ pub fn discover_sourcemap_embedded_debug_id(contents: &str) -> Option<DebugId> {
...
+        let haystack = haystack.as_bytes();
+        let mut matches = self.finder.rfind_iter(haystack);
+
+        let value = loop {
+            let pos = matches.next()?;
+            let prefix = haystack.get(pos.checked_sub(4)?..pos)?;
+
+            // The match starts at needle, e.g. `sourceMappingURL=`, check if the characters
+            // before, are what we expect.
+            let is_match = (prefix == b"//# ") || (self.allow_at && prefix == b"//@ ");
```
Dav1dde, line 105: We could be a bit more lenient here to allow any whitespace (e.g. including `\t`), but I am not sure if we actually want that?

symbolic-debuginfo/src/js.rs
```diff
@@ -50,19 +50,277 @@ pub fn discover_sourcemap_embedded_debug_id(contents: &str) -> Option<DebugId> {
...
+        // Split until the next whitespace:
+        let value = value
+            .split(u8::is_ascii_whitespace)
+            .next()
+            // If there is no whitespace, assume until end.
+            .unwrap_or(value);
+
+        // This should never fail, the input was a valid string, we only trimmed characters in the
+        // ascii range.
+        std::str::from_utf8(value).ok()
```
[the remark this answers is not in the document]
Dav1dde: Yes, I talk about that in the issue, which probably fed into this context and no source to back it up -_-

symbolic-debuginfo/src/js.rs
```diff
@@ -5,26 +5,26 @@
...
+        let without_query = url.split_once('?').map(|x| x.0).unwrap_or(url);
+        let without_fragment = without_query
+            .split_once('#')
+            .map(|x| x.0)
+            .unwrap_or(without_query);
+
+        return Some(without_fragment);
     }
+
     None
```
loewenheim, line 28: This should be simpler with `?`:

```suggestion
    let url = finder.find(contents)?;
    // The URL might contain a query string or fragment. Strip those away before recording the URL.
    let without_query = url.split_once('?').map(|x| x.0).unwrap_or(url);
    let without_fragment = without_query
        .split_once('#')
        .map(|x| x.0)
        .unwrap_or(without_query);

    Some(without_fragment)
```

symbolic-debuginfo/src/js.rs
```diff
@@ -50,19 +50,277 @@ pub fn discover_sourcemap_embedded_debug_id(contents: &str) -> Option<DebugId> {
...
-        if let Some(rest) = line.strip_prefix("//# debugId=") {
-            return rest.trim().parse().ok();
+    MagicCommentFinder::debug_id()
+        .find(contents)
+        .and_then(|s| s.parse().ok())
+}
+
+/// A helper utility which allows searching for magic comments in JavaScript sources.
+///
+/// The finder will optionally consider the `//#` variant as well as the deprecated `//@` variant.
```
loewenheim, line 60: This sounds like `//#` is optional, instead of `//@`.
```suggestion
/// The finder will optionally consider the deprecated `//@` variant in addition to the official `//#` variant.
```

symbolic-debuginfo/src/js.rs
loewenheim: I think this is totally fine as it is.

loewenheim (changes requested):

loewenheim (approved):

PR #976, opened 2026-05-04
dep: Bump publicly exposed dependencies

loewenheim:
This updates some dependencies which have a semver impact because some of their types are publicly exposed:
- `gimli`: 0.32.3 -> 0.33.0
- `goblin`: 0.8.0 -> 0.10.5
- `scroll`: 0.12.0 -> 0.13.0
- `uuid`: 1.3.0 -> 1.23.0

The `goblin` part of this is just the depedency bump and minor adjustment from #960, without the use of the new feature.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,7 +28,7 @@ gimli = { version = "0.32.3", default-features = false, features = [
   "std",
   "fallible-iterator",
 ] }
-goblin = { version = "0.8.0", default-features = false }
+goblin = { version = "0.10.5", default-features = false }
 indexmap = "2.0.0"
 insta = { version = "1.28.0", features = ["yaml"] }
 itertools = "0.14.0"
@@ -49,7 +49,7 @@ proptest = "1.6.0"
 regex = "1.7.1"
 rustc-demangle = "0.1.21"
 # keep this in sync with whatever version `goblin` uses
-scroll = "0.12.0"
+scroll = "0.13.0"
 serde = { version = "1.0.171", features = ["derive"] }
 serde_json = "1.0.102"
 similar-asserts = "1.4.2"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,7 +23,7 @@ fallible-iterator = "0.3.0"
 flate2 = { version = "1.1.8", default-features = false, features = [
   "rust_backend",
 ] }
-gimli = { version = "0.32.3", default-features = false, features = [
+gimli = { version = "0.33.0", default-features = false, features = [
   "read",
   "std",
   "fallible-iterator",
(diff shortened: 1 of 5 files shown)


symbolic-debuginfo/src/dwarf.rs
```diff
@@ -55,11 +55,11 @@ type RangeLists<'a> = gimli::read::RangeLists<Slice<'a>>;
 type Unit<'a> = gimli::read::Unit<Slice<'a>>;
 type DwarfInner<'a> = gimli::read::Dwarf<Slice<'a>>;
 
-type Die<'d, 'u> = gimli::read::DebuggingInformationEntry<'u, 'u, Slice<'d>, usize>;
+type Die<'d, 'u> = gimli::read::DebuggingInformationEntry<Slice<'d>, usize>;
```
Dav1dde:
```suggestion
type Die<'d> = gimli::read::DebuggingInformationEntry<Slice<'d>, usize>;
```

Seems like `'u` is never used.

tobias-wilfert (approved):

symbolic-debuginfo/src/dwarf.rs
```diff
@@ -656,7 +654,7 @@ impl<'d, 'a> DwarfUnit<'d, 'a> {
                                 // are triggered by an inverted range (going high to low).
                                 // See a few more examples of broken ranges here:
                                 // https://github.com/emscripten-core/emscripten/issues/15552
-                                Err(gimli::Error::InvalidAddressRange) => None,
+                                Err(gimli::Error::InvalidCfiSetLoc(_)) => None,
```
loewenheim, line 657: In 0.32, the `InvalidAddressRange` error variant is returned only in the `evaluate` function:
https://github.com/gimli-rs/gimli/blob/71720829ef6f61ea5b0d4e1b9315534d2c8b9264/src/read/cfi.rs#L2325-L2327

However, in 0.33, this variant is removed and `InvalidCfiSetLoc` is now returned at the same location:
https://github.com/gimli-rs/gimli/blob/8e70a8accfef4472dd09dbcd7cdbfdefe4803c6b/src/read/cfi.rs#L2369-L2371

symbolic-debuginfo/src/dwarf.rs
loewenheim: TIL that you're allowed to have unused lifetime parameters in type aliases, but not type definitions (and unused type parameters aren't allowed in either).

Cargo.toml
```diff
@@ -62,7 +62,7 @@ time = { version = "0.3.20", features = ["formatting"] }
 tokio = "1.36.0"
 tracing = "0.1.34"
 tracing-subscriber = "0.3.11"
-uuid = "1.3.0"
```
Dav1dde, line 65: Unlike the other's this isn't a breaking change. Does symbolic require `1.23`? If not, there is an argument to be made to leave it at the minimum bound which is required.

Either way, seems fine.

Dav1dde (approved):

Cargo.toml
loewenheim: You're right, I've removed this update.

PR #980, opened 2026-05-18
feat(demangle): Import libswift demangle 6.3.1

### LINKED ISSUE #979: Update Swift demangler
Swift demangling functionality hasn't been updated in a while. #913 stalled out because I couldn't quite make it work.

Dav1dde:
Fixes: #979
Refs: #913 

- Updates the Python script to add a missing header which can't be inferred from `clang -H`.
- Updates the build and Python script to exclude cpp files which aren't actually needed.
- Updates the patch.
- Updates the readme to first checkout, then run the scripts from the repo
- Added the testcase from #913, potentially adding another testcase for 6.3

There is a failing test case `$s11Supercharge2AXO7ElementPAAE8elements33_35EDDAA799FBB5B74D2F426690B0D99DLL3for2asSayqd__GSo28NSAccessibilityAttributeNamea_qd__mtSo7AXErrorVYKAcDRd__lFAC3AppC_AC6WindowCTgm5` which I removed, it also fails with `swift demangle` on the command line. Best guess right now is, that the original symbol is now mangled differently and the old version can't be de-mangled.
I will do a timeboxed investigation if we can support the old mangled string, but generally we prefer supporting more recent versions over older ones.


loewenheim (approved):

symbolic-demangle/vendor/swift/README.md
Brett-Best: This likely needs updating

Dav1dde: Thanks!

Brett-Best:
Test cases that can be added:
1. `$s12SharedDomain7ServiceC06streamC0_4bodyACXDScSyxG_yxYaYbKYCcts8SendableRzlFZyycAA06StreamC0CYbcfu_yycfu0_Tm`
2. `$s18OrderedCollections0A10DictionaryV20uniqueKeysWithValuesACyxq_Gqd___tcSTRd__x_q_t7ElementRtd__lufC6TaggedAHVy8SportsUI0J23ParticipantRowViewModelVSSG_ALs15LazyMapSequenceVySayALGAM_ALtGTt0g5Tf4g_n`

These currently fail and can be added.

Dav1dde:
@Brett-Best thanks, added the tests and confirmed they fail on master. Marked them as 6.3 even though technically they could be 6.2 or anything in between, but that should be fine.

Also found a solution to fix the breaking test in #981 with an additional patch.

Brett-Best:
Thanks @Dav1dde, appreciate the quick turnaround once I landed in the right place :)

PR #984, opened 2026-06-08
feat(debuginfo): Add object file parsing options

loewenheim:
This adds a new struct `ParseObjectOptions` to `symbolic-debuginfo` which is intended to customize parsing of object files. To this end, `parse_with_options` functions are added to many types that take an argument of type `ParseObjectOptions`.

For now, the only option is `max_decompressed_section_size`, which is intended to limit the uncompressed sizes of compressed sections in ELF files.

There are also two new tests to make sure the limiting of section sizes for ELF files actually works.


symbolic-debuginfo/src/macho/mod.rs
```diff
@@ -70,6 +71,11 @@ impl<'d> MachObject<'d> {
     }
 
     /// Tries to parse a MachO from the given slice.
+    pub fn parse_with_opts(data: &'d [u8], _opts: ParseObjectOptions) -> Result<Self, MachError> {
+        Self::parse(data)
```
Dav1dde: Nit: could also move the body of parse to parse_with_opts.

Also what @jjbayer noted may make sense to just drop a small comment that opts is unused but there for API reasons.

(applies to all objects)

symbolic-debuginfo/src/object.rs
```diff
@@ -124,6 +124,16 @@ impl Error for ObjectError {
...
 }
 
+/// Options for parsing object files.
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy, Default)]
+pub struct ParseObjectOptions {
+    /// Maximum uncompressed size for compressed debug file sections.
+    ///
+    /// This is only relevant to ELF objects.
+    pub max_decompressed_section_size: Option<u64>,
```
Dav1dde: Any specific reason you reached for `u64` rather than `usize`? Practically the allocation can't got beyond `usize`.

```suggestion
    pub max_decompressed_section_size: Option<usize>,
```

symbolic-debuginfo/src/elf.rs
```diff
@@ -612,6 +625,13 @@ impl<'data> ElfObject<'data> {
             (ty, compression.ch_size, compressed)
         };
 
+        if self
+            .max_decompressed_section_size
+            .is_some_and(|max| size > max)
```
Dav1dde: Maybe formats better `if size > self.max_decompressed_section_size.unwrap_or(u64::max)`

symbolic-debuginfo/tests/test_objects.rs
```diff
@@ -424,6 +425,24 @@ fn test_elf_debug_link_compressed() -> Result<(), Error> {
...
+}
+
+#[test]
+fn test_elf_compressed_shf() {
+    let mut opts = ParseObjectOptions::default();
+    opts.max_decompressed_section_size = Some(2 << 30);
+    let view = ByteView::open(fixture("linux/elf_compressed_shf")).unwrap();
+    let object = ElfObject::parse_with_opts(&view, opts).unwrap();
+    let _ = object.debug_session();
+}
```
Dav1dde, line 465: As discussed, we can make a binary with a small size, e.g. 50 bytes and test once with a threshold of 49 and once with 50 bytes and assert different behaviour.

symbolic-debuginfo/src/object.rs
loewenheim: I missed the fact that it made more sense to change the other side of the comparison to `usize` rather than make this a `u64`.

symbolic-debuginfo/src/macho/mod.rs
```diff
@@ -385,8 +394,8 @@ impl<'d> Parse<'d> for MachObject<'d> {
         Self::test(data)
     }
 
-    fn parse(data: &'d [u8]) -> Result<Self, MachError> {
-        Self::parse(data)
+    fn parse_with_opts(data: &'d [u8], opts: ParseObjectOptions) -> Result<Self, Self::Error> {
+        Self::parse_with_opts(data, opts)
```
jjbayer: What if instead of making this a forwarder, we remove `parse_with_opts` from `impl MachObject` and only define it in `impl Parse for MachObject`? Then you can still use it in the macro, but it becomes more obvious why the function has this signature.

loewenheim: Would the `parse_with_opts` on `MachObject` then delegate to `MachObject::parse`?

jjbayer: I guess so, yes.

symbolic-debuginfo/tests/test_objects.rs
```diff
@@ -424,6 +425,32 @@ fn test_elf_debug_link_compressed() -> Result<(), Error> {
     check_debug_info("elf_with_compressed_debuglink", "debug_info.txt")
 }
 
+#[test]
+fn test_elf_compressed_gnu() {
+    let view = ByteView::open(fixture("regression/elf_compressed_gnu")).unwrap();
+    let mut opts = ParseObjectOptions::default();
+
+    let object = ElfObject::parse_with_opts(&view, opts).unwrap();
+    assert!(object.debug_session().is_err());
```
Dav1dde: Does this end in an error because the section is invalid but correctly decompressed?

Not sure why there is an error asserted without the size limit, a comment would help here explaining why.

Dav1dde (approved):

PR #985, opened 2026-06-17
fix(macho): Update offset for macho debug info process

rodolfoBee:
This PR is a result of this issue from the Gimli library: https://github.com/gimli-rs/gimli/issues/890

According to the maintainer, the issue is caused by the file size and the 32bit offset limitation. This pull follows the recommendation from the maintainer. 

I tested it locally with two affected files, and the original error was no longer reproducible. 

There is also a new PR in the Gimli repo that might also fix the issue without having to merge changes to Symbolic: https://github.com/gimli-rs/object/pull/923


symbolic-debuginfo/src/macho/mod.rs
```diff
@@ -475,11 +475,32 @@ impl<'data> Dwarf<'data> for MachObject<'data> {
...
+                                let start = usize::try_from(file_offset).ok()?;
+                                let len = usize::try_from(header.size).ok()?;
+                                let end = start.checked_add(len)?;
+                                Some((self.data.get(start..end)?, file_offset))
+                            }) {
+                            Some((data, file_offset)) => (data, file_offset),
+                            // Fall back to the data goblin provided if the segment-relative
+                            // computation is out of bounds or overflows.
+                            None => (data, u64::from(header.offset)),
+                        };
```
loewenheim: The logic looks sound, but since it's only relevant for 64bit files, you can gate it behind the `is_64` flag:

```suggestion
                        let (data, offset) = if self.macho.is_64 {
                            match header
                            […]
                        } else {
                            (data, u64::from(header.offset))
                        }
```

Also, a comment linking to the alternative LLVM solution in https://github.com/llvm/llvm-project/blob/5dc8ac225fa7965518fb49cce90516b46c745e0f/llvm/lib/Object/MachOObjectFile.cpp#L1992-L2016 would be welcome.

loewenheim (changes requested):

Dav1dde (approved):

loewenheim (approved):

PR #986, opened 2026-06-18
feat: WebAssembly bindings + @sentry/symbolic npm package

BYK:
## Summary

Adds WebAssembly bindings for `symbolic-debuginfo` and publishes them to npm as **`@sentry/symbolic`**, so JavaScript/TypeScript tools can parse debug information files (Mach-O/dSYM, ELF, PE/PDB, Portable PDB, WebAssembly, Breakpad, source bundles) without a native binary.

Motivation: the [TypeScript sentry-cli rewrite](https://github.com/getsentry/cli) needs to parse DIFs for `debug-files` commands. There is no JS/WASM build of symbolic and no pure-JS parser for these formats (especially PDB/DWARF), so this compiles `symbolic-debuginfo` to `wasm32-unknown-unknown`.

## What's here

**1. Make `symbolic-debuginfo` wasm32-compatible (backward-compatible)**
- `symbolic-common`: `cfg`-gate `memmap2` out for `wasm32` (no `mmap` syscall there). `ByteView`'s mmap constructors are unavailable on wasm; callers build from in-memory bytes (`from_slice`/`from_vec`), and `open` reads the whole file.
- `symbolic-debuginfo`: split the `zstd` C library out of the `elf` feature into a new **`elf-zstd`** feature (it doesn't build for wasm32). `elf-zstd` is included in `default`, so **native builds are unchanged and still link the C zstd**.
- For wasm, a new **`elf-zstd-pure`** feature decompresses zstd-compressed ELF debug sections with the pure-Rust [`ruzstd`](https://crates.io/crates/ruzstd) decoder. When both features are on (e.g. `--all-features`), the C `zstd` implementation wins. So zstd is supported everywhere — C on native, pure-Rust on wasm.

**2. `symbolic-wasm` crate** — a small `wasm-bindgen` wrapper exposing:
- `parse_debug_file(bytes)` → archive metadata (debug id, code id, arch, kind, feature flags per object)
- `peek_format(bytes)` → detected format name

`publish = false` (npm-only, not crates.io).

**3. `@sentry/symbolic` npm package** (`symbolic-wasm/npm/`)
- `make npm` / `symbolic-wasm/build-npm.sh`: builds the wasm (size-optimized via `--config` overrides since a non-root `[profile.release]` is ignored), runs `wasm-bindgen` (web target, with a version-matched CLI auto-installed), `wasm-opt -Oz`, and `npm pack`.
- Ships the glue + `.wasm` (≈**1.14 MB raw / 456 KB gz**, ~467 KB tarball). Runs in Node and browsers; the host passes file bytes in.

**4. Release integration**
- `.github/workflows/build.yml`: new `npm-package` job builds the tarball and uploads `artifact-npm`, merged into the sha-named release artifact.
- `.craft.yml`: `npm` target (`access: public`) + `requireNames` for `sentry-symbolic-*.tgz`.
- `scripts/bump-version`: keeps `@sentry/symbolic` version in lockstep with the crates.

## Validation

- Native `cargo check`/`clippy --all-features` and `cargo fmt` pass; native builds still link the C zstd.
- `symbolic-wasm` compiles to `wasm32-unknown-unknown` (incl. `ruzstd`); the packed tarball parses real ELF and Breakpad inputs correctly (debug id matches `readelf`). `ruzstd` decode validated against zstd-compressed input.


symbolic-wasm/src/lib.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    let mut objects = Vec::new();
+    for object in archive.objects() {
+        let object = object.map_err(|e| JsError::new(&e.to_string()))?;
+        objects.push(ObjectInfo {
+            debug_id: object.debug_id().to_string(),
+            code_id: object.code_id().map(|c| c.to_string()),
+            arch: object.arch().to_string(),
+            file_format: format!("{:?}", object.file_format()),
+            kind: format!("{:?}", object.kind()),
```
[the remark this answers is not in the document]
BYK: Good catch — fixed in 0f18ef0. `file_format`, `kind`, and `arch` now use the canonical `.name()` strings (`elf`, `lib`, `x86_64`, …) to match `symbolic-cabi`/Python instead of Rust `Debug` (`Elf`, `Executable`, …). `peek_format` too, and the README/doc examples were updated.

BYK:
## Native build impact — reproducibility check

To verify this PR doesn't change native builds, I compiled the two touched crates at the base commit (`3581ad0`) and at the PR head with identical settings and compared the artifacts.

**Method:** `rustc 1.92.0`, `--release`, default features, `CARGO_INCREMENTAL=0`, `RUSTFLAGS="-Cdebuginfo=0 --remap-path-prefix=<src>=/s --remap-path-prefix=~/.cargo=/cargo"`, then `ar x` each `.rlib` and compared members.

**`symbolic-common`** — metadata hash identical (`47995d3f…`):
- All **5/5 compiled object units (`*.rcgu.o`) are byte-identical** ✅
- Only `lib.rmeta` differs (252504 → 252520 bytes)

→ The `memmap2`/`ByteView` wasm-gating is a **true no-op for native codegen**. The tiny `rmeta` delta is just source-span offsets shifted by the added `#[cfg]` attribute lines (compile-time metadata, not machine code).

**`symbolic-debuginfo`** — not byte-identical:
- Metadata-disambiguator hash changed (`e52449e3…` → `bb63e760…`), which propagates through symbol mangling across all CGUs.

This is **by design, not a behavior change**: supporting wasm requires splitting `zstd` out of the `elf` feature into `elf-zstd` (a Cargo feature can't conditionally enable a dependency per-target, and the C `zstd` crate can't build for wasm32). `elf-zstd` stays in `default`, so the **enabled dependencies are unchanged** (native default still links the C `zstd`) and the **cfg-expanded native code is unchanged** (under `default`, the active zstd match arm is the identical C-`zstd` one). But changing the default feature *set* changes Cargo's `-Cmetadata` hash → different symbol mangling → different bytes. There's no way to avoid this while supporting wasm.

**Conclusion:** native behavior is unchanged — `symbolic-common`'s object code is byte-identical, and `symbolic-debuginfo`'s only native delta is the cosmetic metadata/symbol-hash rename intrinsic to the `elf-zstd` split. Combined with the green test suite, native consumers are unaffected.

betegon (approved):

symbolic-debuginfo/src/elf.rs
```diff
@@ -637,7 +637,29 @@ impl<'data> ElfObject<'data> {
...
+                let decoder = ruzstd::StreamingDecoder::new(compressed).ok()?;
+                let mut decompressed = Vec::with_capacity(size);
+                // Bound the output to the declared decompressed size, matching the
+                // zlib and C-`zstd` arms (and honoring `max_decompressed_section_size`,
+                // which is checked against `size` above). `read_to_end` would
+                // otherwise grow unboundedly for a crafted high-ratio stream.
+                decoder
+                    .take(size as u64)
+                    .read_to_end(&mut decompressed)
+                    .ok()?;
```
Dav1dde, line 656: This creates partially decompressed parts, unlike `zstd::bulk::decompress` which returns an error if the passed `size` is too small.

symbolic-debuginfo/Cargo.toml
```diff
@@ -45,8 +46,15 @@ elf = [
...
-    "zstd",
 ]
+# Optional zstd support for zstd-compressed ELF debug sections.
+# Split out from `elf` because the `zstd` crate is a C library that does not
+# build for wasm32. Included in `default` to preserve existing behavior.
+elf-zstd = ["elf", "dep:zstd"]
+# Pure-Rust zstd support (via `ruzstd`) for targets where the `zstd` C library
+# is unavailable, e.g. wasm32. Used by the `symbolic-wasm` crate. When both
+# `elf-zstd` and `elf-zstd-pure` are enabled, the C `zstd` implementation wins.
+elf-zstd-pure = ["elf", "dep:ruzstd"]
```
Dav1dde, line 57: This is a breaking or at least silently breaking change, `default-features = false, features = ["elf"]` now behaves differently.

As discussed with `mmap`, a way to avoid this is to use `cfg`, which may be better suited here than a feature as it rarely makes sense to disable support for zstd here. Might need some slight finessing to get cargo to do what we want.

symbolic-wasm/npm/.gitignore
```diff
@@ -0,0 +1,7 @@
+# Generated by build-npm.sh (wasm-bindgen output + packed tarball).
```
Dav1dde, line 1: Prefer to have these top level and if you want to scope it, you can just specify paths with wildcards.

symbolic-wasm/build-npm.sh
```diff
@@ -0,0 +1,70 @@
...
+  --omit-default-module-path \
+  "$WASM_IN"
+
+BG_WASM="$NPM_DIR/symbolic_bg.wasm"
+
+if command -v wasm-opt >/dev/null 2>&1; then
+  echo "Optimizing with wasm-opt -Oz..."
+  wasm-opt -Oz -o "$BG_WASM" "$BG_WASM"
+else
+  echo "wasm-opt not found — skipping size optimization" >&2
```
Dav1dde, line 62: Maybe worth throwing an error here so we don't end up publishing an unoptimized wasm by accident.

symbolic-wasm/Cargo.toml
```diff
@@ -0,0 +1,45 @@
...
+# Published to npm (@sentry/symbolic), not crates.io.
+publish = false
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[dependencies]
+wasm-bindgen = "0.2"
+serde = { version = "1", features = ["derive"] }
+serde-wasm-bindgen = "0.6"
```
Dav1dde, line 26: These should be workspace dependencies

Cargo.lock
```diff
@@ -1976,6 +1976,15 @@ dependencies = [
  "wait-timeout",
 ]
 
+[[package]]
+name = "ruzstd"
+version = "0.7.3"
```
Dav1dde, line 1981: Latest version is `0.8.3`

symbolic-common/src/byteview.rs
```diff
@@ -315,7 +331,7 @@ pub enum AccessPattern {
 }
 
 impl AccessPattern {
-    #[cfg(unix)]
+    #[cfg(all(unix, not(target_arch = "wasm32")))]
```
Dav1dde, line 334: `unix` and `wasm32` should be mutually exclusive already?

symbolic-common/src/byteview.rs
```diff
@@ -11,6 +11,7 @@ use std::ops::Deref;
 use std::path::Path;
 use std::sync::Arc;
 
+#[cfg(not(target_arch = "wasm32"))]
```
Dav1dde, line 14: I think it makes sense to have `mmap` as a feature instead, anyone may want to toggle it on/off not just wasm, as it is not an essential feature for the crate to work.

Unfortunately this is a breaking change, so the `cfg` is a decent middleground for now.

Dav1dde (changes requested):
Right now `symbolic-wasm` is only useful for the sentry cli, unlike the Python bindings which offer almost full support for symbolic's API.

If the only usecase and motiviation for creating this wasm crate/npm package, then it may be better placed in the sentry cli, this also enables faster iteration in the CLI with full access to the symbolic Rust API.

Right now, without plans for API parity, this is not the right place for the package. Unlike the name suggests, it doesn't offer access to symbolic's API from WASM.

.github/workflows/build.yml
```diff
@@ -97,10 +97,34 @@ jobs:
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3
+
+      - run: rustup toolchain install stable --profile minimal --target wasm32-unknown-unknown --no-self-update
+
+      # Node/npm are preinstalled on ubuntu runners (used for `npm pack`).
+      # Provides wasm-opt for size optimization (build-npm.sh skips it if absent).
+      - run: npm install -g binaryen
```
Dav1dde, line 111: Should pin a version here, especially with the recent surge of supply chain attacks.

PR #988, opened 2026-06-18
feat(debuginfo): filesystem-free source bundles via a source provider

BYK:
Adds the library primitive for building source bundles **without filesystem access** — the building block for the WebAssembly source-bundle API.

> **Rescoped.** This now contains only the `symbolic-debuginfo` change. The `@sentry/symbolic` (wasm) surface — `Archive`/`Object`, source enumeration, `createSourceBundle` — moves to @Dav1dde's foundational wasm PR (the `self_cell`-based ownership model). This PR is the filesystem-free primitive that sits underneath it, independent of the wasm side. Rebased onto `master` (post-#990).

## What

`SourceBundleWriter::write_object_with_source_provider` obtains each source file's contents from a caller-supplied `provider` returning any `impl std::io::Read`, instead of `std::fs`. The host (e.g. a wasm embedder) supplies the bytes.

`write_object_with_filter` now **delegates** to a shared private `write_object_with_filter_and_provider` (passing `File::open`), so there's a single copy of the collection loop. The filesystem path's output is **byte-for-byte unchanged** (verified: `write_object` produces an identical bundle at base vs HEAD). Because the provider returns `Read`, each referenced source is requested **at most once** — now a documented guarantee rather than a correctness requirement for destructive providers.

## Tests

`tests/test_source_bundle.rs`: round-trips a bundle built via a provider (re-parses it and confirms the supplied source is embedded, and that an unreferenced provided path is not), an empty-provider case, and a zstd-compressed-DWARF fixture.


symbolic-wasm/src/lib.rs
```diff
@@ -71,3 +98,111 @@ pub fn parse_debug_file(data: &[u8]) -> Result<JsValue, JsError> {
...
+        }
+    }
+    let object =
+        target.ok_or_else(|| JsError::new(&format!("no object with debug id {debug_id}")))?;
+
+    let sink = SharedCursor::new();
+    let writer = SourceBundleWriter::start(sink.clone()).map_err(to_js)?;
+    writer
+        .write_object_with_source_provider(&object, object_name, |path| contents.remove(path))
+        .map_err(to_js)?;
```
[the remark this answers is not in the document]
BYK: Thanks — investigated, and this is safe: `write_object_resolved` shares a single `files_handled` set across both loops. The main loop inserts every `session.files()` path into `files_handled` unconditionally (even when the provider returns `None`), and the il2cpp `referenced_files` loop skips any path already in `files_handled`. So no path is ever passed to `provider` more than once — including a file that is both a normal source and an il2cpp-referenced file. A destructive provider (`HashMap::remove`) is therefore safe.

To make this explicit and guard against regressions:
- Documented the guarantee on `write_object_with_source_provider` ("each referenced source path is passed to `provider` at most once …").
- Added a regression test (`test_write_object_with_source_provider_il2cpp_destructive`) that reproduces your exact scenario — `collect_il2cpp_sources(true)` + a destructive `|path| contents.remove(path)` provider, where a main source carries a `//<source_info:…>` reference — and asserts the il2cpp-referenced file is bundled, not dropped.

Done in d2fa0b3.

BYK:
## Final review + local byte-identity verification

### 1. Original code provably unchanged (purely additive)
`git diff 15e05ed..HEAD` is **417 insertions, 0 deletions** — no existing line is removed or modified anywhere in the PR. Verified that in `symbolic-debuginfo/src/sourcebundle/mod.rs`, the existing `write_object`, `write_object_with_filter`, the il2cpp `referenced_files` loop, `collect_il2cpp_sources`, and the `use std::io::{…}` imports are **byte-for-byte identical** to base; the only change is the new standalone `write_object_with_source_provider` method. Likewise `symbolic-wasm`'s existing `parse_debug_file`/`peek_format` are unchanged. The new method is a separate implementation (not a refactor of the filesystem path) precisely to guarantee this.

### 2. Local byte-identity test (e2e)
Built a small harness that calls the **existing** `SourceBundleWriter::write_object` and ran it against `symbolic-debuginfo` at **base `15e05ed`** vs the **PR HEAD**, on the same ELF+DWARF input (with the referenced sources present on disk):

| Build | bundle bytes | sha256 |
|-------|-------------:|--------|
| base `15e05ed` | 7517 | `6ec7aac4ac7cea51484dcdf0b3c18760626a1b6e10bad66923907d885612197f` |
| PR HEAD | 7517 | `6ec7aac4ac7cea51484dcdf0b3c18760626a1b6e10bad66923907d885612197f` |

**`cmp` → identical.** The native filesystem source-bundle output is byte-for-byte unchanged by this PR.

### 3. New-code correctness (critical review)
- `write_object_with_source_provider` mirrors the existing collection logic, substituting the provider for `std::fs::read`. The `files_handled`/`referenced_files` dedup is intact, so each path is requested **at most once** — a destructive provider (`HashMap::remove`) is safe even with il2cpp. Covered by a dedicated regression test.
- `SharedCursor` finish-then-read is sound (single-threaded wasm; reads the full backing `Vec` after the ZIP central directory is flushed). `create_source_bundle` returns `undefined` when nothing is bundled.
- Tests: round-trip (with unreferenced-path exclusion), empty-provider, zstd-DWARF, and il2cpp + destructive-provider — 4/4 pass.

### Checks
`cargo fmt --check`, `cargo clippy --all-features -D warnings`, `cargo test` (4/4), and `wasm32-unknown-unknown` release build all pass. CI green: Rust tests (macOS/Linux/Windows), doc, Style/Linting, Danger, Cursor Bugbot, and Seer Code Review.

symbolic-debuginfo/src/sourcebundle/mod.rs
```diff
@@ -1409,6 +1409,106 @@ where
...
+    ///
+    /// This finishes the source bundle and flushes the underlying writer.
+    ///
+    /// [`write_object_with_filter`]: Self::write_object_with_filter
+    pub fn write_object_with_source_provider<'data, 'object, O, E, P>(
+        mut self,
+        object: &'object O,
+        object_name: &str,
+        mut provider: P,
+    ) -> Result<bool, SourceBundleError>
```
Dav1dde: This makes sense, we shouldn't have a hard filesystem dependency when the underlying writer is generic.

But implementation wise, it'd be great if the existing methods delegate to the more generic one, like currently `write_object` delegates to `write_object_with_filter`. Following this pattern we can rewrite `write_object_with_filter` to delegate to `write_object_with_source_provider`/`write_object_with_filter_and_source_provider` (ungodly names, we should fix in a breaking change release and probably cleanup the API surface a bit).

That means we don't have to maintain 2 copies of the same code.

symbolic-debuginfo/tests/test_source_bundle.rs
```diff
@@ -0,0 +1,161 @@
...
+        paths.push(path);
+    }
+    Ok(paths)
+}
+
+/// `write_object_with_source_provider` bundles source content supplied by the
+/// caller instead of reading from the filesystem, so it works even when none of
+/// the original build-time source files exist on disk.
+#[test]
+fn test_write_object_with_source_provider() -> Result<(), Error> {
```
Dav1dde: Tests should not return results, they only make debugging harder, just `unwrap()` and you end up with the same error message but also a stack trace pointing you to the location where the assert failed.

symbolic-debuginfo/tests/test_source_bundle.rs
```diff
@@ -0,0 +1,161 @@
...
+
+use std::collections::{HashMap, HashSet};
+use std::io::Cursor;
+
+use symbolic_common::ByteView;
+use symbolic_debuginfo::sourcebundle::SourceBundleWriter;
+use symbolic_debuginfo::Object;
+use symbolic_testutils::fixture;
+
+type Error = Box<dyn std::error::Error>;
```
Dav1dde: Usually what you do is:


```suggestion
type Result<T, E = Box<dyn std::error::Error>> = std::result::Result<T, E>;
```

Then you also don't need to type error anymore at all.

symbolic-debuginfo/src/sourcebundle/mod.rs
```diff
@@ -1409,6 +1409,106 @@ where
...
+    /// references, are skipped), so a destructive provider — e.g. one that moves
+    /// bytes out of a map — is safe.
+    ///
+    /// This is a filesystem-free analogue of [`write_object_with_filter`]; the
+    /// two are kept as separate implementations so this addition does not alter
+    /// the existing filesystem code path.
+    ///
+    /// This finishes the source bundle and flushes the underlying writer.
+    ///
+    /// [`write_object_with_filter`]: Self::write_object_with_filter
```
Dav1dde: This is a lot of words to just say what `write_object_with_filter` says + "you can bring your own filesystem". A lot of unnecessary noise about `il2cpp` a caller doesn't need to care about.

The upside of not requesting the same source twice, is a general improvement, which can be documented as a behavior and guarantee but then also all other implementations should give that same behavior.

In general I would not document it, and leave it up to the implementation, from a caller perspective it shouldn't make much of a difference if we change the signatures to not require unnecessary ownership of the data.

symbolic-debuginfo/src/sourcebundle/mod.rs
```diff
@@ -1409,6 +1409,106 @@ where
...
+    pub fn write_object_with_source_provider<'data, 'object, O, E, P>(
+        mut self,
+        object: &'object O,
+        object_name: &str,
+        mut provider: P,
+    ) -> Result<bool, SourceBundleError>
+    where
+        O: ObjectLike<'data, 'object, Error = E>,
+        E: std::error::Error + Send + Sync + 'static,
+        P: FnMut(&str) -> Option<Vec<u8>>,
```
Dav1dde: None of the function does need ownership of the data, the `Vec<u8>` is not required, this can be a `&[u8]` slice or even better anything that implements `std::fs::Read` as that is actually the bound which is required.

Then all this destructive stuff isn't relevant at all anymore.

symbolic-debuginfo/tests/test_source_bundle.rs
```diff
@@ -0,0 +1,161 @@
...
+
+    Ok(())
+}
+
+/// Regression test for the il2cpp + destructive-provider case: a source file
+/// referenced via an il2cpp `//<source_info:...>` comment is collected in a
+/// second pass. Because each path is requested at most once, a destructive
+/// provider (`HashMap::remove`) must still bundle it (not silently drop it).
+#[test]
+fn test_write_object_with_source_provider_il2cpp_destructive() -> Result<(), Error> {
```
Dav1dde: No longer necessary if we have to guarantee this behavior, all the destructiveness is just a nice to have if the provider can just return `std::io::Read`.

symbolic-wasm/src/lib.rs
```diff
@@ -71,3 +82,129 @@ pub fn parse_debug_file(data: &[u8]) -> Result<JsValue, JsError> {
...
+/// Returns the source bundle bytes, or `undefined` if no source files were
+/// bundled (so the caller can skip writing an empty bundle).
+#[wasm_bindgen]
+pub fn create_source_bundle(
+    data: &[u8],
+    debug_id: &str,
+    object_name: &str,
+    sources: JsValue,
+) -> Result<Option<Vec<u8>>, JsError> {
+    let pairs: Vec<(String, ByteBuf)> = serde_wasm_bindgen::from_value(sources).map_err(to_js)?;
```
Dav1dde: This is quite an expensive thing to do, have you considered to provide an API which passes a callback to retrieve the data instead?

symbolic-wasm/src/lib.rs
```diff
@@ -71,3 +82,129 @@ pub fn parse_debug_file(data: &[u8]) -> Result<JsValue, JsError> {
...
+    let object =
+        target.ok_or_else(|| JsError::new(&format!("no object with debug id {debug_id}")))?;
+
+    let sink = SharedCursor::new();
+    let writer = SourceBundleWriter::start(sink.clone()).map_err(to_js)?;
+    let written = writer
+        .write_object_with_source_provider(&object, object_name, |path| contents.remove(path))
+        .map_err(to_js)?;
+
+    Ok(written.then(|| sink.bytes()))
```
Dav1dde:
```suggestion
    let mut sink = Cursor::new(Vec::new());
    let writer = SourceBundleWriter::start(&mut sink).map_err(to_js)?;
    let written = writer
        .write_object_with_source_provider(&object, object_name, |path| contents.remove(path))
        .map_err(to_js)?;

    Ok(written.then(|| sink.into_inner()))
```

This also makes that intense type `SharedCursor(Rc<RefCell<Cursor<Vec<u8>>>>)` obsolete.

Dav1dde (commented):
As mentioned in the previous PR, I think all the `symbolic-wasm` logic which is only required for Sentry cli, should not live in this repo but be moved to the new cli repo, similar to how the "old" sentry-cli also has these functions implement in its own code base and actually uses symbolic just as a dependency.

The direction of making `symbolic` not require the filesystem, is just a great change and direction to take.

symbolic-debuginfo/src/sourcebundle/mod.rs
BYK: Done — `write_object_with_filter` now delegates to a shared private `write_object_with_filter_and_provider` (passing `File::open`), so there's a single copy of the loop. Verified the filesystem path is byte-for-byte unchanged: `write_object` output has identical sha256 at base vs HEAD (7584 bytes). (5ed9e330)

symbolic-debuginfo/src/sourcebundle/mod.rs
BYK: Done — the provider now returns any `impl std::io::Read`. The wasm side hands it a `&[u8]` borrowed from the source map (non-destructive), so no ownership transfer is needed.

symbolic-debuginfo/src/sourcebundle/mod.rs
BYK: Trimmed — the doc no longer mentions il2cpp. "Each referenced source path is requested at most once" is now documented as a plain guarantee on the method.

symbolic-debuginfo/tests/test_source_bundle.rs
BYK: Removed — with the `Read` provider there's no destructiveness, so this regression test is no longer meaningful.

symbolic-debuginfo/tests/test_source_bundle.rs
BYK: Done — tests no longer return `Result`; they `unwrap()` so panics point at the failing assertion.

symbolic-debuginfo/tests/test_source_bundle.rs
BYK: Dropped the `type Error` alias entirely (tests no longer return `Result`).

symbolic-wasm/src/lib.rs
BYK: Done — replaced `SharedCursor(Rc<RefCell<Cursor>>)` with a plain `Cursor::new(Vec::new())` + `into_inner()`, as suggested.

symbolic-wasm/src/lib.rs
BYK: Made the provider non-destructive for now (it borrows `&[u8]` from the map), and `list_source_files` already bounds the map to referenced files so it isn't materializing the whole tree. A callback-style retrieval is a natural fit for the general-purpose API-base restructure (below); I'd prefer to land it there, but happy to do it here if you'd rather.

BYK:
Thanks for the thorough review — addressed it all in 5ed9e330:

- **Delegation + generic `Read` provider.** `write_object_with_filter` now delegates to a shared private `write_object_with_filter_and_provider`; the provider returns any `impl std::io::Read`. One copy of the loop, and the filesystem path is **byte-for-byte unchanged** (`write_object` output: identical sha256 at base vs HEAD).
- **Docs/tests cleanup.** Trimmed the il2cpp noise; tests `unwrap()` instead of returning `Result`; dropped the destructive-provider test (moot with `Read`).
- **wasm simplification.** Plain `Cursor` + `into_inner()` instead of `SharedCursor`; non-destructive `&[u8]` provider.

**On the bigger direction** — agreed (per our offline chat): keep `symbolic-wasm` here but grow it into a **general-purpose API base** (Archive/Object classes mirroring the Python bindings) rather than CLI-shaped helpers. This PR keeps the surface minimal but on that path; I'll follow up with the class-based API (and can fold the data-retrieval callback into that).

**Smoke test (your idea — love it).** Added `symbolic-wasm/smoke-test.mjs`: it loads the **built** wasm and runs every export (`peek_format`, `parse_debug_file`, `list_source_files`, `create_source_bundle`) against a real fixture, wired into `build-npm.sh` so a broken JS/wasm boundary fails the build (and CI). This closes the gap where the bindings were only compile-checked.

BYK:
Rescoped to just the `symbolic-debuginfo` primitive (`write_object_with_source_provider` + the `write_object_with_filter` delegation) and rebased onto `master` (post-#990). The `@sentry/symbolic` wasm surface — `Archive`/`Object`, source enumeration, `createSourceBundle` — is moving to @Dav1dde's foundational wasm PR (the `self_cell` ownership model), so I dropped it here; this PR is the filesystem-free building block underneath it. The earlier inline threads about the wasm code are therefore moot.

Dav1dde:
Nice, gonna base my WASM PR on this, then get it merged tomorrow!

Dav1dde (approved):

PR #989, opened 2026-06-19
fix(debuginfo): address review feedback from #986 (zstd via cfg, ruzstd 0.8, packaging)

BYK:
Follow-up to #986 addressing @Dav1dde's review comments.

## Changes

**zstd: `cfg(target_arch)` instead of features** (the main one)
- Removes the `elf-zstd` / `elf-zstd-pure` features. `#986` split `zstd` out of `elf` into `elf-zstd`, which **silently changed** `default-features = false, features = ["elf"]` (it lost zstd support). Now `zstd` (C, native) and `ruzstd` (pure-Rust, wasm32) are selected by `cfg(target_arch)`, so `features = ["elf"]` keeps zstd support everywhere it can — native behavior matches pre-#986.

**ruzstd: 0.8.3 + reject size mismatches**
- Bumped to 0.8.3 (and added to `[workspace.dependencies]`).
- The previous `take(size).read_to_end(...)` could silently truncate a stream that decompresses to more than `size`. Now it reads `size`+1 bytes and requires the result to be exactly the declared `size`, returning `None` on mismatch — matching `zstd::bulk::decompress`, which errors when the size is wrong. Verified: exact size → ok; too-small → rejected (no partial output); too-large → rejected.

**symbolic-common**
- `cfg(all(unix, not(target_arch = "wasm32")))` → `cfg(unix)` (mutually exclusive already).

**symbolic-wasm / packaging**
- Use `[workspace.dependencies]` for `wasm-bindgen` / `serde` / `serde-wasm-bindgen`; drop the `elf-zstd-pure` feature.
- Move the generated npm artifacts into the **top-level** `.gitignore` (scoped paths) instead of `symbolic-wasm/npm/.gitignore`.
- `build-npm.sh`: make `wasm-opt` **required** — error out instead of packaging an unoptimized wasm by accident.
- CI: **pin** `binaryen@130.0.0` (supply-chain).

## Verification
- `cargo check -p symbolic-debuginfo --all-features` (native, links `zstd`) and `cargo build -p symbolic-wasm --target wasm32-unknown-unknown` (links `ruzstd`) both pass.
- `cargo fmt --check`, `cargo clippy --all-features -D clippy::all`, and `cargo test -p symbolic-debuginfo --all-features` (incl. the zstd fixtures) all pass.


Dav1dde (approved):
Thanks!

symbolic-debuginfo/src/elf.rs
```diff
@@ -637,29 +637,30 @@ impl<'data> ElfObject<'data> {
...
+                // be exactly `size`, matching `zstd::bulk::decompress` which errors on
+                // a size mismatch.
                 let mut decompressed = Vec::with_capacity(size);
-                // Bound the output to the declared decompressed size, matching the
-                // zlib and C-`zstd` arms (and honoring `max_decompressed_section_size`,
-                // which is checked against `size` above). `read_to_end` would
-                // otherwise grow unboundedly for a crafted high-ratio stream.
                 decoder
-                    .take(size as u64)
+                    .take(size as u64 + 1)
```
loewenheim, line 656: I don't think this fixes the problem—if the decompressed stream is, say, actually `size + 2` bytes long it would still truncate. Can you think of a better way to produce an error here @Dav1dde , and is it even really necessary?

Dav1dde: @loewenheim did you see the check right below? That should address it, no?

> and is it even really necessary?

Probably not, but I'd prefer to actually fail here instead of leave half-finished data around, which will lead to a different kind of failure downstream.

loewenheim: Oh, duh. I did miss it. Looks good in that case!

PR #990, opened 2026-06-19
feat(debuginfo): use C zstd on wasm too, drop ruzstd

BYK:
Follow-up to #989. Per the discussion about why we shipped two zstd implementations, it turns out we can use the **C `zstd` library on wasm too**, so we don't need the pure-Rust `ruzstd` fallback at all.

## How

`zstd-sys` ships a `wasm-shim/` whose headers `#define malloc`/`memcpy`/`qsort`/… to `rust_zstd_wasm_shim_*` functions implemented in Rust (`#[no_mangle]`, backed by Rust's allocator). Its `build.rs` auto-enables the shim for `wasm32-*` targets, so the C zstd compiles to `wasm32-unknown-unknown` with **no libc/sysroot** — it just needs **clang** (which has a native WebAssembly backend).

## What changes

- `symbolic-debuginfo`: one `zstd` dependency for **all** targets (no `[target.'cfg(...)']` split), and a single `zstd::bulk::decompress` arm in `elf.rs` (drops the wasm-only `ruzstd` arm and its manual size-mismatch guard).
- Drops the `ruzstd` dependency entirely.
- New `wasm-build` CI job compiles `symbolic-wasm` for `wasm32-unknown-unknown` (installs `clang`/`lld`/`llvm`) so the C-on-wasm path stays covered on PRs.

## Why

- **One implementation everywhere** — answers "why two decoders?". Native keeps full speed; wasm now also gets full C-zstd speed (`ruzstd` is significantly slower per its author), instead of trading correctness for a pure-Rust build.
- Simpler code: no `cfg(target_arch)` split, no `ruzstd` size-check workaround.

## Validation

- Builds: `symbolic-wasm` → `wasm32-unknown-unknown` compiles **and links** (clang 18 + the shim).
- Correctness on wasm: built `@sentry/symbolic` with C-zstd and ran `list_source_files` against `crash.debug-zstd` (zstd-compressed DWARF) → **152 sources, identical to the uncompressed fixture** — i.e. the wasm decompressed the zstd sections correctly. (`list_source_files` lands in #988; verified locally on that branch.)
- Native: `cargo test -p symbolic-debuginfo --all-features` (incl. the zstd fixtures) passes; `fmt` + `clippy` clean.

## Cost

- **clang** is required to build for wasm. It's preinstalled on GitHub `ubuntu-latest` runners (the new job also installs it explicitly to be safe); local contributors running `make npm` need it too.
- Bundle size: **~+3%** (`symbolic_bg.wasm` 2349 → 2428 KB raw, 902 → 926 KB gzip).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,23 @@ jobs:
 
       - run: cargo doc --workspace --all-features --document-private-items --no-deps
 
+  wasm-build:
+    name: WASM build
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3
+
+      - run: rustup toolchain install stable --profile minimal --target wasm32-unknown-unknown --no-self-update
+
+      - uses: Swatinem/rust-cache@42dc69e1aa15d09112580998cf2ef0119e2e91ae # v2
+
+      # The C `zstd` library is compiled to wasm by clang (zstd-sys's wasm-shim);
+      # llvm provides llvm-ar to archive the wasm objects. clang is normally
+      # preinstalled on the runner; install explicitly so the build is reproducible.
+      - run: sudo apt-get update && sudo apt-get install -y clang lld llvm
+
+      - run: cargo build -p symbolic-wasm --target wasm32-unknown-unknown
+
   test-rust:
     strategy:
       fail-fast: false
--- a/symbolic-debuginfo/Cargo.toml
+++ b/symbolic-debuginfo/Cargo.toml
@@ -110,9 +110,6 @@ thiserror = { workspace = true }
 wasmparser = { workspace = true, optional = true }
 zip = { workspace = true, optional = true }
 
-# zstd-compressed ELF debug sections. The C `zstd` library also compiles to
-# wasm32 (zstd-sys ships a wasm-shim that maps malloc/memcpy/... onto Rust),
-# so a single dependency covers every target — no pure-Rust fallback needed.
 zstd = { workspace = true }
 
 [dev-dependencies]
--- a/symbolic-debuginfo/Cargo.toml
+++ b/symbolic-debuginfo/Cargo.toml
@@ -110,15 +110,10 @@ thiserror = { workspace = true }
 wasmparser = { workspace = true, optional = true }
 zip = { workspace = true, optional = true }
 
-# zstd-compressed ELF debug sections are decompressed with the C `zstd` library
-# on native targets, and the pure-Rust `ruzstd` decoder on wasm32 (where the
-# `zstd` C library does not build). Selected by `cfg(target_arch)` rather than a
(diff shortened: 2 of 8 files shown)


symbolic-debuginfo/Cargo.toml
```diff
@@ -110,16 +110,11 @@ thiserror = { workspace = true }
...
 zip = { workspace = true, optional = true }
 
-# zstd-compressed ELF debug sections are decompressed with the C `zstd` library
-# on native targets, and the pure-Rust `ruzstd` decoder on wasm32 (where the
-# `zstd` C library does not build). Selected by `cfg(target_arch)` rather than a
-# feature so `features = ["elf"]` keeps zstd support everywhere it can.
-[target.'cfg(not(target_arch = "wasm32"))'.dependencies]
+# zstd-compressed ELF debug sections. The C `zstd` library also compiles to
+# wasm32 (zstd-sys ships a wasm-shim that maps malloc/memcpy/... onto Rust),
+# so a single dependency covers every target — no pure-Rust fallback needed.
```
Dav1dde: This is an unnecessary comment.

symbolic-debuginfo/Cargo.toml
```diff
@@ -110,16 +110,11 @@ thiserror = { workspace = true }
 wasmparser = { workspace = true, optional = true }
 zip = { workspace = true, optional = true }
```
Dav1dde, line 112:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -50,6 +50,23 @@ jobs:
...
+      - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3
+
+      - run: rustup toolchain install stable --profile minimal --target wasm32-unknown-unknown --no-self-update
+
+      - uses: Swatinem/rust-cache@42dc69e1aa15d09112580998cf2ef0119e2e91ae # v2
+
+      # The C `zstd` library is compiled to wasm by clang (zstd-sys's wasm-shim);
+      # llvm provides llvm-ar to archive the wasm objects. clang is normally
+      # preinstalled on the runner; install explicitly so the build is reproducible.
+      - run: sudo apt-get update && sudo apt-get install -y clang lld llvm
```
Dav1dde, line 66: Isn't this then now also missing in the `image.yml`?

Dav1dde (approved):

symbolic-debuginfo/Cargo.toml
BYK: Removed in f4ffd445.

.github/workflows/ci.yml
BYK: Good catch. There's no `image.yml` in this repo — the other place that builds the wasm is the release `npm-package` job in `build.yml` (it runs `make npm`, which now compiles the C zstd to wasm). Added `clang lld llvm` there too in f4ffd445. (It would otherwise rely on the runner's preinstalled clang; better to be explicit.)

PR #993, opened 2026-06-22
test(wasm): packaged-artifact smoke test (+ fix Object/global shadowing)

BYK:
Follow-up to #992 (now on master). Adds the "does the published package actually work?" smoke-test layer we discussed, and fixes a ship-blocking bug it caught.

## The bug it caught (now live on master)

The class exported as `Object` shadows the JS global `Object` for the whole module, so the generated `--target web` glue breaks where it uses the global:
- `Object.create(Object.prototype)` — used by `archive.objects()`
- `Object.getPrototypeOf(module)` — used by `initSync`

So a real consumer doing `import * as symbolic from "@sentry/symbolic"; symbolic.initSync({ module })` gets `TypeError: Object.getPrototypeOf is not a function`. `wasm-pack test` (the `test-wasm` job) passes because it builds its own glue and never loads the shipped `--target web` artifact.

Fixed by exporting the class as **`ObjectFile`** via `#[wasm_bindgen(js_name = ObjectFile)]` (`objects(): ObjectFile[]`). The JS name is bikesheddable — anything that isn't a JS global works.

## The smoke test

`npm pack`s the package, installs the tarball into a throwaway project, and runs it through `node:test` as a real consumer would — bare `import "@sentry/symbolic"` (the `.` export), `require.resolve("@sentry/symbolic/symbolic_bg.wasm")` (the `./symbolic_bg.wasm` export + `files[]`), and `initSync` of the shipped glue.

- `symbolic-wasm/package-smoke.test.mjs` — the `node:test` assertions.
- `symbolic-wasm/smoke-test.mjs` — pack + install + run orchestrator.
- `build-npm.sh` — runs it after packing (so `make npm` covers it locally + in the release build).
- `ci.yml` — a `wasm-package` job runs `make npm` on PRs.

Complements `wasm-pack test --node` (behavior, node/browser) with the packaging layer it can't cover.

## Validation

`make npm` builds and the smoke test passes; without the rename it fails at `initSync`, which is the point.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,23 +50,6 @@ jobs:
 
       - run: cargo doc --workspace --all-features --document-private-items --no-deps
 
-  wasm-build:
-    name: WASM build
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3
-
-      - run: rustup toolchain install stable --profile minimal --target wasm32-unknown-unknown --no-self-update
-
-      - uses: Swatinem/rust-cache@42dc69e1aa15d09112580998cf2ef0119e2e91ae # v2
-
-      # The C `zstd` library is compiled to wasm by clang (zstd-sys's wasm-shim);
-      # llvm provides llvm-ar to archive the wasm objects. clang is normally
-      # preinstalled on the runner; install explicitly so the build is reproducible.
-      - run: sudo apt-get update && sudo apt-get install -y clang lld llvm
-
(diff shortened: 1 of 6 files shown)


BYK:
Updated per @Dav1dde's feedback: the smoke test now runs as `npm test` from `symbolic-wasm/npm/`. The test files (`smoke-test.mjs`, `package-smoke.test.mjs`) live in the package dir but are excluded from `files[]` (it's an allowlist), so `cd symbolic-wasm/npm && npm test` works and nothing extra is published. Kept it as pack+install so it still exercises the real consumer path (exports map + resolution), not just a local file load.

Dav1dde (approved):

PR #996, opened 2026-06-22
build(deps): Bump memmap2 floor to 0.9.11 (RUSTSEC-2026-0186)

mingley:
## Summary

Raises the workspace `memmap2` floor from `^0.9.0` to `0.9.11`, excluding `0.9.0..=0.9.10` — the range affected by [RUSTSEC-2026-0186](https://rustsec.org/advisories/RUSTSEC-2026-0186.html) (unchecked pointer offset in `advise_range`/`flush_range`). The committed `Cargo.lock` resolves to `0.9.5` (affected); this updates it to `0.9.11`.

- `symbolic-common`'s `byteview` uses only `Mmap::map` and whole-map `Mmap::advise` ([byteview.rs](https://github.com/getsentry/symbolic/blob/master/symbolic-common/src/byteview.rs)), not the offset-based `advise_range`/`flush_range` — floor-raise for hygiene, not an exploitable fix.
- MSRV-safe: memmap2 0.9.11 requires Rust 1.65 (symbolic MSRV 1.83).

## Testing Done

- `cargo check -p symbolic-common` — clean.
- `cargo test -p symbolic-common` — 16 unit + 60 doc tests pass (incl. `byteview` mmap path).


Dav1dde:
@mingley hey! What's your motiviation for the change, I think symbolic doesn't use the affected APIs directly and the used version already allows users to use 0.9.11?

mingley:
> @mingley hey! What's your motiviation for the change, I think symbolic doesn't use the affected APIs directly and the used version already allows users to use 0.9.11?

Just doing ecosystem cleanup on some transitive deps that can easily make the upgrade. 

Related: https://github.com/tikv/pprof-rs/pull/293

At work, doing this floor increase is simpler than having the constant churn of different teams pinging asking if they can add a new ignore line to their deny.toml files, as the floor increase makes the state unrepresentable and the ignore not needed. I own an internal profiling lib that a bunch of services use and we transitively pull this lib in. 

If you don't want to bump the patch, it's fine, the main one I care about is https://github.com/tikv/pprof-rs/pull/293

Dav1dde (approved):

Dav1dde:
@mingley thanks, no reason not to do this if it is useful for you.

PR #998, opened 2026-06-24
feat(msvc): Vendor `msvc-demangler`

tobias-wilfert:
Part 1 of fixing:

Customers have been experiencing stack overflows when demangeling msvc symbols. Since the library that is used for demangeling (https://github.com/mstange/msvc-demangler-rust) does not allow to set a max recursion limit, vendor it and add that limit.

Ref: INGEST-950


symbolic-demangle/vendor/msvc-demangler/lib.rs
```diff
@@ -1440,6 +1493,10 @@ impl<'a> ParserState<'a> {
     }
 
     fn read_nested_array(&mut self, dimension: i32) -> Result<(Type<'a>, StorageClass)> {
+        self.bounded_recursion(|s| s.read_nested_array_inner(dimension))
+    }
+
+    fn read_nested_array_inner(&mut self, dimension: i32) -> Result<(Type<'a>, StorageClass)> {
```
tobias-wilfert: Tried adding this to all cases of excessive recursion that I could find (even if it is not triggered by the one test case below).

symbolic-demangle/src/lib.rs
```diff
@@ -65,6 +65,13 @@ extern "C" {
     fn symbolic_demangle_is_swift_symbol(sym: *const c_char) -> c_int;
 }
 
+/// Maximum parser recursion depth when demangling MSVC symbols.
+///
+/// Since some symbols can cause stack overflows this is used to limit the max recursion depth so
+/// we fail gracefully.
+#[cfg(feature = "msvc")]
+const MSVC_MAX_RECURSION_DEPTH: usize = 128;
```
tobias-wilfert: Value is somewhat chosen arbitrarily.

symbolic-demangle/src/lib.rs
```diff
@@ -33,6 +33,8 @@
 
 #![warn(missing_docs)]
 
+extern crate alloc;
```
Dav1dde: This is only required for `msvc-demangler` because it supports no-std, would either drop the line and change the vendored code to just use `std::` or at least add a comment here why this is necessary.

Dav1dde (changes requested):
Please split the PR into 1) vendoring the library, 2) making the recursion changes, so we have a clear diff of what was changed and what is original code.

symbolic-demangle/vendor/msvc-demangler/lib.rs
tobias-wilfert: Adding this logic was moved to a seperate PR.

symbolic-demangle/src/lib.rs
tobias-wilfert: Adding this logic was moved to a seperate PR.

symbolic-demangle/vendor/msvc-demangler/lib.rs
```diff
@@ -0,0 +1,2401 @@
...
+// Demangler class takes the main role in demangling symbols.
+// It has a set of functions to parse mangled symbols into Type instnaces.
+// It also has a set of functions to cnovert Type instances to strings.
+struct ParserState<'a> {
+    // Mangled symbol. read_* functions shorten this string
+    // as they parse it.
+    remaining: &'a [u8],
+
+    // The original input
+    input: &'a str,
```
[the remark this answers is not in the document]
tobias-wilfert: Yes this will be added in a separate PR.

symbolic-demangle/vendor/msvc-demangler/lib.rs
```diff
@@ -0,0 +1,2401 @@
...
+use core::fmt;
+use core::mem;
+use core::result;
+use core::str;
+use core::str::Utf8Error;
+use std::borrow::Cow;
+use std::boxed::Box;
+use std::string::String;
+use std::string::{FromUtf8Error, ToString};
+use std::vec::Vec;
```
loewenheim: Do you actually need to import `Box`, `String`, and `Vec`?

Dav1dde (approved):

tobias-wilfert:
Will address the issues found by Sentry bot in a separate PR.

PR #999, opened 2026-06-24
feat(debuginfo): Add parsing option for limiting embedded source sizes

loewenheim:
This adds a new field `max_embedded_source_size` to `ParseObjectOptions`. This option is only relevant for object types that can contain embedded source files (or, to be precise, from which `symbolic` can _extract_ embedded source files). Which is to say, source bundles and portable PDB files.

Both of these object types now validate the (claimed) size of embedded sources against this maximum and return an error if it is exceeded. Both file formats have gained new error kinds for this case.

# Alternatives

We could also not make this a part of `ParseObjectOptions`, but rather add versions of `source_by_{path,url,debug_id}` that take a size limit, because that's where the limit actually ends up getting used.

ref: INGEST-966

### DIFF
--- a/symbolic-ppdb/src/format/mod.rs
+++ b/symbolic-ppdb/src/format/mod.rs
@@ -106,6 +106,9 @@ pub enum FormatErrorKind {
     /// Failed to parse Source Link JSON
     #[error("invalid source link JSON")]
     InvalidSourceLinkJson,
+    /// An embedded source file exceeded the configured size limit.
+    #[error("embedded source file size ({0}) exceeds maximum")]
+    EmbeddedSourceFileSizeExceeded(usize),
 }
 
 /// An error encountered while parsing a [`PortablePdb`] file.
@@ -476,7 +479,7 @@ impl<'data, 'object> EmbeddedSource<'data> {
     }
 
     /// Reads the source file contents from the Portable PDB.
-    pub fn get_contents(&self) -> Result<Cow<'data, [u8]>, FormatError> {
+    pub fn get_contents(&self, max_size: Option<usize>) -> Result<Cow<'data, [u8]>, FormatError> {
         // The blob has the following structure: `Blob ::= format content`
         // - format - int32 - Indicates how the content is serialized.
         //     0 = raw bytes, uncompressed.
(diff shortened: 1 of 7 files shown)


symbolic-ppdb/src/format/mod.rs
```diff
@@ -488,9 +491,24 @@ impl<'data, 'object> EmbeddedSource<'data> {
...
         let (format_blob, data_blob) = self.blob.split_at(4);
         let format = u32::from_ne_bytes(format_blob.try_into().unwrap());
+
+        // Per the above comment, the format is really an `i32`.
+        // None-negative values are currently valid, negative ones are reserved.
+        // We can't trivially change this to an `i32`, though, because of
+        // `FormatErrorKind::InvalidBlobFormat`, which expects a `u32`.
+        // Therefore, we manually bound the format here to the largest
+        // positive `i32`.
+        const MAX_VALID_FORMAT: u32 = i32::MAX as u32;
```
loewenheim, line 512: This part is really unfortunate—I don't know why we didn't make `InvalidBlobFormat` take an `i32` to begin with. We could fix it as part of a breaking release.

symbolic-ppdb/src/format/mod.rs
```diff
@@ -488,9 +491,24 @@ impl<'data, 'object> EmbeddedSource<'data> {
...
+
```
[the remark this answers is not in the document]
loewenheim: This is a fair point; one the one hand, for uncompressed sources, we don't have to allocate and just return a slice, so it's not really dangerous. On the other hand, from the name of the option, one wouldn't expect its functionality to depend on whether the source is compressed.

I'm open to suggestions here—do we
1. make the option apply to uncompressed sources as well,
2. change the option's name so it's clear it only applies to compressed sources,
3. or leave the option's name alone and just clarify the semantics in the documentation?

symbolic-ppdb/src/format/mod.rs
```diff
@@ -476,7 +479,7 @@ impl<'data, 'object> EmbeddedSource<'data> {
     }
 
     /// Reads the source file contents from the Portable PDB.
-    pub fn get_contents(&self) -> Result<Cow<'data, [u8]>, FormatError> {
+    pub fn get_contents(&self, max_size: Option<usize>) -> Result<Cow<'data, [u8]>, FormatError> {
```
Dav1dde: That's a breaking change :(

In the API here: https://docs.rs/symbolic-ppdb/latest/symbolic_ppdb/struct.EmbeddedSource.html#method.get_contents

Maybe a new method and adding it to the breaking changes issue?

symbolic-ppdb/src/format/mod.rs
Dav1dde: Add it to the issue please

symbolic-ppdb/src/format/mod.rs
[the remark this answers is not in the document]
Dav1dde: I think we can just document it, that this only applies to decompression for now.

Potentially rename the config to show that this is only scoped for decompression.

Dav1dde (changes requested):

symbolic-ppdb/src/format/mod.rs
loewenheim: Dang, missed that this is public (which it obviously is, otherwise we couldn't call it from `symbolic-debuginfo` 🙃 )

symbolic-ppdb/src/format/mod.rs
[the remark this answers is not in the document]
loewenheim: I've renamed it to be explicit about this.

Dav1dde (approved):

PR #1000, opened 2026-06-24
ci(ci): Add cargo semver checks

Dav1dde:
Experiment to add cargo semver checks, which hopefully catches accidental semver changes before things are merged.


loewenheim:
I can't tell if it's working.

Dav1dde:
> I can't tell if it's working.

it's not :(

PR #1002, opened 2026-06-25
fix(dwarf): Fix range to addr/line matrix mapping

Dav1dde:
We observed in WASM emscripten builds that the function range is determined by the `low pc` and `high pc` of the function, this range happened to start before a valid row in the debug line matrix. For example a function range may go from `0x07` to `0x30` but the function range goes from `0x06` to `0x30`, due to the function prologue being included in the function range (1 byte in this case).

This hits a case in the line / row resolution code, where the beginning of the function range does not match any row in the sequence and therefor the row is skipped.
The updated logic essentially clamps the start of the range to the beginning of the row sequence, similarly what was already done for the end of the range.

Also fixes an unrelated off by one comparison on the range/sequence check.

Commit 6ad9b1df4973a2a3d09f51d55e11097fb9a51816 shows the before and after.

Also testable with `addr2line`:

```
└─ % ./run addr2line -e index.wasm -i -f -C 0x1e9
+ cargo run --quiet --package addr2line -- -e index.wasm -i -f -C 0x1e9
level_1()
  at /tmp/wasm-repro/main.cpp:34
```


symbolic-debuginfo/src/dwarf.rs
```diff
@@ -338,21 +338,51 @@ impl<'d> DwarfLineProgram<'d> {
...
 
             let from = match seq.rows.binary_search_by_key(&range.begin, |x| x.address) {
                 Ok(idx) => idx,
-                Err(0) => continue,
+                // If the range begins before the sequence, we clamp it to the beginning of the
+                // sequence as a range determined by the low and high pc may include a function
+                // prologue which is not mapped into the addr/line matrix.
+                //
+                // Since the sequences are ordered by address, this can only happen if the beginning
+                // of the range is not part of another range already.
```
loewenheim:
```suggestion
                // of the range is not part of another sequence already.
```

loewenheim (approved):
Minor correction, otherwise looks good.

PR #1003, opened 2026-06-26
fix(unreal): Handle XML parsing without recursion

klochek:
Sadly, quickxml's serde support does not handle empty tags the way we'd like--even for an optional field, like `foo : Option<String>`, if it sees a `<foo>` tag in the xml, it expects there to be a value in the tag, and it will error if there isn't one, while we'd like to interpret an empty tag (`<foo></foo>`) as just None.  So, do it the hard way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,7 +49,6 @@
 - `symbolic-debuginfo`: Add a new option `max_decompressed_embedded_source_size` to `ParseObjectOptions`. This option limits the sizes of compressed embedded source files
   when they are returned from methods like `source_by_path`. ([#999](https://github.com/getsentry/symbolic/pull/999))
 
-
 ## 13.5.0
 
 **Features**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 - Add limit on inline nesting depth when parsing Breakpad objects. ([#1014](https://github.com/getsentry/symbolic/pull/1014))
 - Fix potential out of bounds panic on invalid section offsets ([#1013](https://github.com/getsentry/symbolic/pull/1013))
 - Fix a potential panic when locating sections in an elf file. ([#1012](https://github.com/getsentry/symbolic/pull/1012))
+- Handle XML parsing without recursion. ([#1003](https://github.com/getsentry/symbolic/pull/1003))
 
 ## 13.8.0
 
@@ -47,6 +48,7 @@
 
 - `symbolic-debuginfo`: Add a new option `max_decompressed_embedded_source_size` to `ParseObjectOptions`. This option limits the sizes of compressed embedded source files
   when they are returned from methods like `source_by_path`. ([#999](https://github.com/getsentry/symbolic/pull/999))
+
 
 ## 13.5.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+**Fixes**
+
+- Handle XML parsing without recursion. ([#1003](https://github.com/getsentry/symbolic/pull/1003))
+
 ## 14.0.0-alpha.1
 
 - No documented changes.
@@ -11,7 +17,6 @@
 - Add limit on inline nesting depth when parsing Breakpad objects. ([#1014](https://github.com/getsentry/symbolic/pull/1014))
 - Fix potential out of bounds panic on invalid section offsets ([#1013](https://github.com/getsentry/symbolic/pull/1013))
 - Fix a potential panic when locating sections in an elf file. ([#1012](https://github.com/getsentry/symbolic/pull/1012))
-- Handle XML parsing without recursion. ([#1003](https://github.com/getsentry/symbolic/pull/1003))
 
 ## 13.8.0
 
--- a/symbolic-unreal/Cargo.toml
+++ b/symbolic-unreal/Cargo.toml
@@ -35,6 +35,7 @@ scroll = { workspace = true, features = ["derive"] }
 serde = { workspace = true, optional = true }
 thiserror = { workspace = true }
 time = { workspace = true }
+
 [dev-dependencies]
 insta = { workspace = true }
 similar-asserts = { workspace = true }
--- a/symbolic-unreal/Cargo.toml
+++ b/symbolic-unreal/Cargo.toml
@@ -28,14 +28,13 @@ serde = ["dep:serde", "chrono/serde"]
 anylog = { workspace = true }
 bytes = { workspace = true }
 chrono = { workspace = true }
-elementtree = { workspace = true }
+quick-xml = { workspace = true, features = ["escape-html"] }
 flate2 = { workspace = true }
 regex = { workspace = true }
 scroll = { workspace = true, features = ["derive"] }
 serde = { workspace = true, optional = true }
 thiserror = { workspace = true }
 time = { workspace = true }
-
 [dev-dependencies]
 insta = { workspace = true }
 similar-asserts = { workspace = true }
--- a/symbolic-unreal/Cargo.toml
+++ b/symbolic-unreal/Cargo.toml
@@ -28,7 +28,7 @@ serde = ["dep:serde", "chrono/serde"]
 anylog = { workspace = true }
 bytes = { workspace = true }
 chrono = { workspace = true }
-quick-xml = { workspace = true, features = ["escape-html"] }
+quick-xml = { workspace = true }
 flate2 = { workspace = true }
 regex = { workspace = true }
 scroll = { workspace = true, features = ["derive"] }
--- a/symbolic-unreal/src/context.rs
+++ b/symbolic-unreal/src/context.rs
@@ -165,7 +165,7 @@ impl Unreal4ContextRuntimeProperties {
 
         let mut rv = Unreal4ContextRuntimeProperties::default();
 
-        if !r.first_instance_of_tag("RuntimeProperties")? {
+        if !r.next_instance_of_tag("RuntimeProperties")? {
             return Ok(None);
         }
 
@@ -258,7 +258,7 @@ impl Unreal4ContextPlatformProperties {
 
         let mut rv = Unreal4ContextPlatformProperties::default();
 
-        if !r.first_instance_of_tag("PlatformProperties")? {
+        if !r.next_instance_of_tag("PlatformProperties")? {
             return Ok(None);
         }
 
@@ -321,7 +321,7 @@ fn load_data_bag(
     let r = quick_xml::Reader::from_reader(data);
     let mut r = XMLReader::new(r);
 
-    if r.first_instance_of_tag(tag)? {
+    if r.next_instance_of_tag(tag)? {
         while let Some(mut child) = r.next_child()? {
             let name = String::from_utf8_lossy(child.tag().name().as_ref()).to_string();
             if let Some(value) = child.value::<String>()? {
--- a/symbolic-unreal/src/context.rs
+++ b/symbolic-unreal/src/context.rs
@@ -1,10 +1,9 @@
 //! Unreal Engine 4 crash context information
 #![warn(missing_docs)]
 
-use elementtree::{Element, QName};
-
 use std::collections::BTreeMap;
 
+use quick_xml::{escape::resolve_xml_entity, events::BytesStart};
 #[cfg(test)]
 use similar_asserts::assert_eq;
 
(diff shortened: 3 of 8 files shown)


symbolic-unreal/src/context.rs
```diff
@@ -389,20 +385,29 @@ const ROOT_WITH_GAME_AND_ENGINE_DATA: &str = r#"<?xml version="1.0" encoding="UT
...
+            .unwrap()
+            .is_none()
+    );
 }
 
 #[test]
 fn test_get_runtime_properties_no_children() {
-    let root = Element::from_reader(ONLY_ROOT_AND_CHILD_NODES.as_bytes()).unwrap();
-    let actual = Unreal4ContextRuntimeProperties::from_xml(&root).expect("default struct");
+    //let root = Element::from_reader().unwrap();
```
loewenheim: Leftover comment.
```suggestion
```

symbolic-unreal/src/context.rs
```diff
@@ -423,10 +428,38 @@ fn test_get_game_and_engine_data() {
...
+    data.push_str("</FGenericCrashContext>");
+
+    let _ = Unreal4Context::parse(data.as_bytes()).unwrap();
+}
+
 #[test]
 fn test_get_platform_properties_no_children() {
-    let root = Element::from_reader(ONLY_ROOT_AND_CHILD_NODES.as_bytes()).unwrap();
-    let actual = Unreal4ContextPlatformProperties::from_xml(&root).expect("default struct");
+    //let root = Element::from_reader().unwrap();
```
loewenheim:
```suggestion
```

symbolic-unreal/src/lib.rs
```diff
@@ -5,6 +5,7 @@ mod container;
 mod context;
 mod error;
 mod logs;
+mod xml;
```
loewenheim, line 8: There's a module added here, but no corresponding file. Possibly left over from an earlier change?

loewenheim (changes requested):

symbolic-unreal/src/xml.rs
```diff
@@ -0,0 +1,194 @@
...
+                    *node_depth += 1;
+                }
+
+                _ => {}
+            };
+
+            if *node_depth < start_depth {
+                break;
+            }
+        }
```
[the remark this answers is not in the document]
Dav1dde: That sounds real, maybe good to just add a test to make sure.

symbolic-unreal/src/xml.rs
```diff
@@ -0,0 +1,197 @@
...
+                // try to be graceful when handling.
+                quick_xml::events::Event::Start(_) => {
+                    *node_depth += 1;
+                }
+                quick_xml::events::Event::Eof => {
+                    break;
+                }
+
+                _ => {}
+            };
```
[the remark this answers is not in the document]
Dav1dde: I think handling `CData` makese sense

symbolic-unreal/src/xml.rs
```diff
@@ -0,0 +1,197 @@
...
+                        if let Ok(decoded) = bytes_text.decode() {
+                            val += &decoded;
+                        }
+                    }
+                }
+
+                quick_xml::events::Event::GeneralRef(bytes_text) => {
+                    if start_depth == *node_depth {
+                        if let Ok(decoded) = bytes_text.decode() {
+                            if let Some(resolved) = resolve_xml_entity(&decoded) {
```
Dav1dde: I think this needs to be `unescape(&decoded)` [docs](https://docs.rs/quick-xml/latest/quick_xml/escape/fn.unescape.html)?

Or some variant of [resolve_char_ref](https://docs.rs/quick-xml/latest/quick_xml/events/struct.BytesRef.html#method.resolve_char_ref).

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+**Fixes**
+
+- Handle XML parsing without recursion ([#1003](https://github.com/getsentry/symbolic/pull/1003))
```
Dav1dde:
```suggestion
- Handle XML parsing without recursion. ([#1003](https://github.com/getsentry/symbolic/pull/1003))
```

symbolic-unreal/src/xml.rs
```diff
@@ -0,0 +1,197 @@
...
+                // It could be the case that we have text interleaved with nodes, so
+                // try to be graceful when handling.
+                quick_xml::events::Event::Start(_) => {
+                    *node_depth += 1;
+                }
+                quick_xml::events::Event::Eof => {
+                    break;
+                }
+
+                _ => {}
```
Dav1dde: I would specifically list the remaining match arms, this makes mistakes more obvious but also means when we update the library we get notified if there are more variants to match.

symbolic-unreal/src/context.rs
```diff
@@ -161,102 +159,82 @@ pub struct Unreal4ContextRuntimeProperties {
 }
 
 impl Unreal4ContextRuntimeProperties {
-    fn from_xml(root: &Element) -> Option<Self> {
-        let list = root.find("RuntimeProperties")?;
+    fn from_xml(root: &[u8]) -> Result<Option<Self>, quick_xml::Error> {
```
Dav1dde, line 162: I think invalid XML like `"foobar"` will now return `Ok(None)` instead of failing on invalid XML, but I guess that's fine?

symbolic-unreal/src/xml.rs
```diff
@@ -0,0 +1,197 @@
...
+            reader,
+            node_state: NodeState::None,
+        }
+    }
+
+    /// Returns the text value, parsed, from the node at which the reader is
+    /// located.  If there is no such node or value, or the parse fails, this will
+    /// return None.  A best-effort is made to deal with text spread between xml nodes.
+    /// XML entities are resolved (so '&qout;' is resolved to '"', and the like.)
+    fn value<T: std::str::FromStr>(&mut self) -> Result<Option<T>, quick_xml::Error> {
```
Dav1dde, line 53: I think this behaves slightly differently than the previous `text()` calls, but that can be okay.

`<ErrorMessage>before<x>inner</x>after</ErrorMessage>` -> `beforeafter` where before it would only yield the first.

symbolic-unreal/Cargo.toml
```diff
@@ -28,14 +28,13 @@ serde = ["dep:serde", "chrono/serde"]
 anylog = { workspace = true }
 bytes = { workspace = true }
 chrono = { workspace = true }
-elementtree = { workspace = true }
+quick-xml = { workspace = true, features = ["escape-html"] }
```
Dav1dde: Why add the feature here, I don't think you're using it.

symbolic-unreal/src/context.rs
```diff
@@ -330,27 +313,39 @@ pub struct Unreal4Context {
...
 
 impl Unreal4Context {
     /// Parses the unreal context XML file.
     pub fn parse(data: &[u8]) -> Result<Self, Unreal4Error> {
-        let root = Element::from_reader(data)?;
+        let mut engine_data = BTreeMap::default();
+        let mut game_data = BTreeMap::default();
+
+        load_data_bag(data, "EngineData", &mut engine_data)?;
+        load_data_bag(data, "GameData", &mut game_data)?;
```
Dav1dde, line 342: This basically parses the data twice, and then later down 2 more times, so we end up parsing the same XML 4 times, is there a better way to do this?

symbolic-unreal/src/xml.rs
Dav1dde, line 53: Calling this twice leaves a weird internal state. It moves the parser past the next sibling, swallowing it, I think.

Maybe fix-able by remembering more state than just `is_empty`, e.g. whether it was consumed.

symbolic-unreal/src/xml.rs
Dav1dde, line 1: This is a pretty self-contained parser, it'd be really nice to have some unit tests, especially for the cases which came up in discussions. There is a lot of room for accidentally misusing it (e.g. calling value() twice).

symbolic-unreal/src/xml.rs
klochek: The decoded text we get at that exact point looks has been stripped of like 'quot', and not '&quot', so the unescaper doesn't do anything with it.  I gather this is known to be an xml ref at this point, so that's why resolving with resolve_xml_entity correctly works.

symbolic-unreal/src/xml.rs
klochek: Done.

symbolic-unreal/src/context.rs
klochek: It depends on how exactly the xml is invalid :P  `foobar` will indeed return Ok(None), but stuff like `<asdf<<` will throw a quick_xml syntax error; quick_xml is very liberal about what it will accept.  Should be okay, though.

symbolic-unreal/src/xml.rs
klochek: One of the disadvantages of doing streaming parsing--but yeah, this will have to be the new expected behaviour.

symbolic-unreal/Cargo.toml
klochek: Removed.

symbolic-unreal/src/context.rs
klochek: It's a streaming processor, so broadly, we either go and write a full Deserialize implementation, or accept this.  Keep in mind, "parsing" is pretty light, quick_xml isn't doing a lot more than some light syntax checking and string decoding, so the cost shouldn't be high.  I looked into doing our own Deserialize implementation, and it's going to be big (quick_xml Deserialize is 8k lines), so I figured, go with something simple for now, and then if the need eventually calls for it, bite the bullet then.

symbolic-unreal/src/xml.rs
klochek: Done.

symbolic-unreal/src/xml.rs
Dav1dde: Does it work for hex codes like `&#x26`?

symbolic-unreal/src/context.rs
Dav1dde: I was wondering more in the direction if we can setup the streaming parser once.


Pseudo (assuming tags aren't nested):
```
load_data_bags(
    data,
   ("EngineData", &mut engine_data),
   ("GameData", &mut game_data),
   ...
)
```

Now, there is no need to do that now, but maybe you have an idea how to easily do that.

CHANGELOG.md
```diff
@@ -43,6 +44,7 @@
 - `symbolic-debuginfo`: Add a new option `max_decompressed_embedded_source_size` to `ParseObjectOptions`. This option limits the sizes of compressed embedded source files
   when they are returned from methods like `source_by_path`. ([#999](https://github.com/getsentry/symbolic/pull/999))
 
+
```
Dav1dde:
```suggestion
```

symbolic-unreal/Cargo.toml
```diff
@@ -28,14 +28,13 @@ serde = ["dep:serde", "chrono/serde"]
...
 chrono = { workspace = true }
-elementtree = { workspace = true }
+quick-xml = { workspace = true }
 flate2 = { workspace = true }
 regex = { workspace = true }
 scroll = { workspace = true, features = ["derive"] }
 serde = { workspace = true, optional = true }
 thiserror = { workspace = true }
 time = { workspace = true }
-
```
Dav1dde, line 38: Looks like an accident

symbolic-unreal/src/context.rs
klochek: Ah, yeah, I think this would work--assuming no weird nesting, as you point out, but I'd have to add support to more general descendent iteration.  Right now, it's: run to the specified tag, and then iterate over immediate children, returning values (which is basically all we need for parsing this format.)
I think if push comes to shove and we need more (enough to be a general XML parser,) I'll just bite the bullet and copy quick-xml's deserialization implementation, and then make the necessary changes to accommodate our own handling of optional types.

symbolic-unreal/src/xml.rs
klochek: yuck, good catch.  Added single character escaping.

loewenheim (approved):
Disclaimer: I haven't completely thought through `xml.rs`

symbolic-unreal/src/xml.rs
```diff
@@ -0,0 +1,285 @@
...
+
+        Ok(None)
+    }
+}
+#[cfg(test)]
+mod tests {
+    // cdata?
+    // get value a few times
+    // missing tags
+
```
Dav1dde: Did you want to still add these tests?


```suggestion
```

Dav1dde (approved):

PR #1005, opened 2026-06-26
feat(wasm): expose ObjectFile.il2cppLineMapping for Il2cpp line mappings

BYK:
## Summary

Exposes a new WASM API `ObjectFile.il2cppLineMapping(provider)` (npm `@sentry/symbolic`) that extracts a Unity Il2cpp C++→C# line mapping from an object and serializes it as JSON — the same artifact the native CLI produces via `symbolic_il2cpp::ObjectLineMapping`.

Unity's Il2cpp transpiles C# to C++, embedding `//<source_info:File.cs:line>` markers in the generated C++. The new API parses those markers from the object's referenced C++ source files and returns the mapping (or `undefined` when none are found).

## Filesystem-free provider

The existing `ObjectLineMapping::from_object` reads referenced source files from disk via `ByteView::open`, which doesn't work under `wasm32-unknown-unknown`. So `symbolic-il2cpp` gains:

- `ObjectLineMapping::from_object_with_provider(object, provider)` — enumerates the object's referenced source paths and obtains each file's bytes from a caller-supplied `provider: FnMut(&str) -> Option<impl AsRef<[u8]>>` instead of the filesystem.
- `from_object` now **delegates** to it (`|path| ByteView::open(path).ok()`), preserving its filesystem behavior byte-for-byte.

This mirrors the established provider pattern from #988 (`SourceBundleWriter::write_object_with_filter_and_provider`) and the WASM `writeObject` binding.

The WASM binding wires a JS `(path: string) => Uint8Array | null | undefined` callback to that provider and returns the JSON mapping bytes (`undefined` when empty).

## Tests

- **Native integration** (`symbolic-il2cpp/tests/from_object.rs`): drives `from_object_with_provider` with an in-memory provider against a real object; covers the populated and empty cases. Runs in the normal `cargo test` job.
- **`wasm_bindgen_test`** (`symbolic-wasm/tests/debuginfo.rs`): exercises the binding via a JS closure provider under `wasm-pack test --node`.
- **Packaged-artifact smoke test** (`symbolic-wasm/npm/package-smoke.test.mjs`): calls `il2cppLineMapping` on the installed package against real wasm; asserts the JSON shape and the `undefined`-on-empty case.

## Notes

- No new features pulled into the wasm build: `symbolic-debuginfo`'s default feature set equals the curated set already used by `symbolic-wasm`; `symbolic-il2cpp` only adds pure-Rust deps (`indexmap`, `serde_json`).
- `symbolic-testutils` added as a path-only dev-dependency to `symbolic-il2cpp` (same pattern as `symbolic-debuginfo`).


symbolic-wasm/src/debuginfo/mod.rs
```diff
@@ -143,6 +144,41 @@ impl Object {
...
+    /// `provider` is a `(path: string) => Uint8Array | null | undefined` callback:
+    /// it receives a referenced source path and returns the file's bytes, or a
+    /// nullish value to skip it. The callback exists because these bindings have no
+    /// filesystem access under WebAssembly — the host reads the files and hands back
+    /// their bytes.
+    ///
+    /// Returns `undefined` when the object references no Il2cpp `source_info`
+    /// annotations (i.e. the mapping would be empty).
+    #[wasm_bindgen(js_name = il2cppLineMapping)]
+    pub fn il2cpp_line_mapping(&self, provider: &js_sys::Function) -> Result<Option<Vec<u8>>> {
```
Dav1dde: Let's also follow the Rust layout here, we can 1:1 add a `ObjectLineMapping` type or since this is basically just a type with a single operation, a free standing function in an equivalent module would do as well.

BYK: Done — moved it out of `ObjectFile` into a dedicated `il2cpp` module (mirroring the `symbolic-il2cpp` crate). I went with the free-standing-function option you mentioned: `il2cppLineMapping(object, provider)`, since `ObjectLineMapping` is really a one-shot `from_object` → `to_writer`. Added a `pub(crate)` accessor on `Object` so the module can borrow the underlying debuginfo object, and moved the wasm-bindgen test to `tests/il2cpp.rs`.

symbolic-wasm/src/il2cpp.rs
```diff
@@ -0,0 +1,49 @@
...
+    let mapping = ObjectLineMapping::from_object_with_provider(object.as_debuginfo(), |path| {
+        let value = provider
+            .call1(&JsValue::UNDEFINED, &JsValue::from_str(path))
+            .unwrap_throw();
+
+        if value.is_null_or_undefined() {
+            return None;
+        }
+
+        Some(js_sys::Uint8Array::new(&value).to_vec())
```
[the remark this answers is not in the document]
BYK: Good catch — fixed in 02d919ce. Added a shared `provider_bytes` helper that returns `None` for a nullish value, the bytes for a `Uint8Array`, and throws a descriptive error for anything else (instead of letting `Uint8Array::new` turn a number into a zero-filled buffer or raise an opaque `TypeError`). Applied it to both provider call sites — the new il2cpp one and the existing `sourcebundle.rs` writer, which shared the identical idiom — and added a smoke-test assertion that a non-`Uint8Array` return throws.

I left `tests/common/mod.rs` as-is: it's a test fixture that constructs its own `Uint8Array`, not a public callback boundary.

BYK:
@Dav1dde ready for re-review when you get a chance. 🙏

Changes since your last pass:

- Moved the binding out of `ObjectFile` into a dedicated `il2cpp` module as a **free function** `il2cppLineMapping(object, provider)` — the option you blessed — mirroring the standalone `symbolic-il2cpp` crate rather than hanging it off the object.
- In `symbolic-il2cpp`, `ObjectLineMapping::from_object` now delegates to a new generic `from_object_with_provider(object, provider)` (filesystem-free counterpart); `from_object` just supplies `|path| ByteView::open(path).ok()`, so the existing behavior is byte-for-byte preserved.
- Fixed a HIGH Seer flagged: the `provider` return value wasn't validated, so a non-`Uint8Array` return silently mis-coerced (a number → zero-filled buffer). Added a shared `provider_bytes` helper that returns `None` for nullish, the bytes for a `Uint8Array`, and throws a descriptive error otherwise — applied to **both** the new il2cpp site and the identical pre-existing `sourcebundle.rs` provider. Added a smoke-test assertion for the throw path. Seer re-reviewed and passed.
- Just pushed a docs-only fixup removing a duplicated `**Dependencies**` block a rebase had left in the CHANGELOG.

CI is fully green and I ran a separate adversarial review pass — only finding was that CHANGELOG dup, now fixed.

symbolic-wasm/src/utils.rs
```diff
@@ -38,3 +40,29 @@ impl From<Error> for wasm_bindgen::JsValue {
...
+/// values raise an opaque `TypeError`. Validating here keeps a misbehaving callback
+/// from corrupting output or aborting the module with an unhelpful panic.
+pub(crate) fn provider_bytes(value: &wasm_bindgen::JsValue) -> Option<Vec<u8>> {
+    use wasm_bindgen::JsCast;
+
+    if value.is_null_or_undefined() {
+        return None;
+    }
+
+    match value.dyn_ref::<js_sys::Uint8Array>() {
```
Dav1dde, line 75: I guess that makes sense, my initial intention was to support everything which can construct an `Uint8Array` but the strictness may prevent some bugs.

Dav1dde (approved):

PR #1006, opened 2026-06-29
feat(msvc): Add recursion limit

tobias-wilfert:
Follow up to: #998. This now introduces a recursion limit into the vendored library. This is a somewhat naive solution in the sense that a non recursive rewrite of the library would probably be better. The idea is to keep the library vendored till there is a solution upstream.

Ref: INGEST-950


symbolic-demangle/src/lib.rs
```diff
@@ -63,6 +63,13 @@ extern "C" {
     fn symbolic_demangle_is_swift_symbol(sym: *const c_char) -> c_int;
 }
 
+/// Maximum parser recursion depth when demangling MSVC symbols.
+///
+/// Since some symbols can cause stack overflows this is used to limit the max recursion depth so
+/// we fail gracefully.
+#[cfg(feature = "msvc")]
+const MSVC_MAX_RECURSION_DEPTH: usize = 128;
```
tobias-wilfert, line 71: This value is somewhat chosen arbitrarily (based on local testing).

symbolic-demangle/vendor/msvc-demangler/lib.rs
```diff
@@ -1438,6 +1491,10 @@ impl<'a> ParserState<'a> {
     }
 
     fn read_nested_array(&mut self, dimension: i32) -> Result<(Type<'a>, StorageClass)> {
+        self.bounded_recursion(|s| s.read_nested_array_inner(dimension))
+    }
+
+    fn read_nested_array_inner(&mut self, dimension: i32) -> Result<(Type<'a>, StorageClass)> {
```
tobias-wilfert, line 1497: Tried adding this to all cases of excessive recursion that I could find (even if it is not triggered by the one test case below).

loewenheim (approved):

PR #1007, opened 2026-06-29
fix(cfi): Limit unwind info chain traversal

tobias-wilfert:
Fix: INGEST-964


symbolic-cfi/src/lib.rs
```diff
@@ -1195,6 +1202,12 @@ impl<W: Write> AsciiCfiWriter<W> {
                 next_function = unwind_info.chained_info;
             }
 
+            // If we reached the `MAX_UNWIND_CHAIN_LEN` limit treat the input as malformed data and
+            // skip emitting CFI.
+            if next_function.is_some() {
+                continue 'functions;
+            }
```
tobias-wilfert: I think this is the behaviour that we want but not sure.

symbolic-cfi/src/lib.rs
```diff
@@ -55,6 +55,9 @@ pub const CFICACHE_MAGIC: u32 = u32::from_be_bytes(*b"CFIC");
 /// The latest version of the file format.
 pub const CFICACHE_LATEST_VERSION: u32 = 2;
 
+/// Maximum length of a chained `UNWIND_INFO` list to follow for a single function.
+const MAX_UNWIND_CHAIN_LEN: usize = 128;
+
```
tobias-wilfert: Value chosen somewhat arbitrarily.

symbolic-cfi/src/lib.rs
loewenheim: I think I misunderstood what you had in mind before—I thought you wanted to emit an error stating that the file is malformed. Between the alternatives of writing the unwind info up to the maximum and not writing it at all I think I prefer the former. Do you have an opinion @Dav1dde?

Also, it could be helpful to add a "stats" struct to the `AsciiCfiWriter` that counts the lengths of info chains. That way, we could emit a metric in Symbolicator and get an idea of what kind of numbers we are actually dealing with here. That can be done as a follow-up, though.

symbolic-cfi/src/lib.rs
```diff
@@ -1095,7 +1098,11 @@ impl<W: Write> AsciiCfiWriter<W> {
             saved_regs.clear();
 
             let mut next_function = Some(function);
-            while let Some(next) = next_function {
+            for _ in 0..MAX_UNWIND_CHAIN_LEN {
```
Dav1dde: Do we need a maximum here or detect cycles?

symbolic-cfi/src/lib.rs
Dav1dde: Ideally we can really detect cycles and don't need a maximum. If we need a maximum ideally we make that configurable and then return an error.

symbolic-cfi/src/lib.rs
tobias-wilfert: The issue is twofold, there could be a cycle (which we could also detect in a different way) or there could be very long chains. The idea was to go with a max to deal with both. But willing to change it to a cycle detection.

Dav1dde: If we need to guard against both, I think a configurable limit (could be in a follow-up) + error seems like a good solution.

tobias-wilfert:
Latest changes:
- Made the limit now configurable on `AsciiCfiWriter` via the builder patterns similar to how `MagicCommentFinder` and `DemangleOptions` do it already.
- Made the limit an `Option<usize>` similar to `max_decompressed_section_size`. 
- Added a new error variant to return if the limit is exceeded.

If we wanted to overwrite/set the value from Symbolicator we would need to add some more code since it is called via `CfiCache::from_object` but I omitted this for now in the hopes that it is not necessary (can add it though if we think it makes sense).

symbolic-cfi/src/lib.rs
```diff
@@ -372,12 +376,27 @@ fn cfi_register_name(arch: CpuFamily, register: u16) -> Option<&'static str> {
...
 impl<W: Write> AsciiCfiWriter<W> {
+    /// Default maximum length of a chained `UNWIND_INFO` list.
+    pub const MAX_UNWIND_CHAIN_LEN: usize = 128;
+
     /// Creates a new `AsciiCfiWriter` that outputs to a writer.
     pub fn new(inner: W) -> Self {
-        AsciiCfiWriter { inner }
+        AsciiCfiWriter {
+            inner,
+            max_unwind_chain_len: Some(Self::MAX_UNWIND_CHAIN_LEN),
```
loewenheim: I would default this to `None`.

symbolic-cfi/tests/test_cfi.rs
```diff
@@ -132,3 +134,26 @@ fn cfi_from_pe_arm64() -> Result<(), Error> {
...
+    });
+
+    match rx.recv_timeout(Duration::from_secs(10)) {
+        Ok(result) => assert_eq!(
+            result,
+            Err(CfiErrorKind::ExceededUnwindChainLength),
+            "should be rejected once the limit is reached"
+        ),
+        Err(_) => panic!("process_pe did not terminate: loop is unbounded"),
+    }
```
loewenheim, line 161: Why does this test involve a separate thread and a channel? It should be possible to write directly.

tobias-wilfert: So that if it regresses we don't have the inf loop running in the test.

symbolic-cfi/src/lib.rs
tobias-wilfert: Ok and then set it at all the call sides?

symbolic-cfi/tests/test_cfi.rs
loewenheim: Duh. That makes total sense actually 😅

symbolic-cfi/src/lib.rs
loewenheim: Yeah we'll have to pass it in in Symbolicator, but that's not a problem.

tobias-wilfert: Ok, changed the logic to default to None.

symbolic-cfi/src/lib.rs
```diff
@@ -372,12 +376,24 @@ fn cfi_register_name(arch: CpuFamily, register: u16) -> Option<&'static str> {
...
-        AsciiCfiWriter { inner }
+        AsciiCfiWriter {
+            inner,
+            max_unwind_chain_len: None,
+        }
+    }
+
+    /// Sets the maximum length of a chained `UNWIND_INFO` list to follow for a
+    /// single function. Chains longer than this cause [`process`](Self::process)
+    /// to fail with [`CfiErrorKind::ExceededUnwindChainLength`]. Defaults to 128.
```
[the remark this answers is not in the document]
tobias-wilfert: Good point.

loewenheim (approved):

PR #1014, opened 2026-07-07
fix(breakpad): Add inline limit

tobias-wilfert:
Fix: INGEST-979


symbolic-testutils/fixtures/regression/breakpad_deep_inline.sym
Dav1dde, line 1: Looks like this can be easily generated in code with an arbitrary depth if you wanted.

symbolic-debuginfo/src/breakpad.rs
```diff
@@ -1172,8 +1183,8 @@ impl<'data> Parse<'data> for BreakpadObject<'data> {
         Self::test(data)
     }
 
-    fn parse_with_opts(data: &'data [u8], _opts: ParseObjectOptions) -> Result<Self, Self::Error> {
-        Self::parse(data)
+    fn parse_with_opts(data: &'data [u8], opts: ParseObjectOptions) -> Result<Self, Self::Error> {
+        Self::parse(data).map(|o| o.max_inline_depth(opts.max_inline_depth))
```
Dav1dde: As we talked about, `parse` can delegate to `parse_with_opts` using default opts and the constructor here can set the inline depth without requiring the setter/builder.

symbolic-debuginfo/src/object.rs
```diff
@@ -137,6 +137,11 @@ pub struct ParseObjectOptions {
     ///
     /// This is only relevant for source bundles and PPDB objects.
     pub max_decompressed_embedded_source_size: Option<usize>,
+
+    /// The maximum inline nesting depth to process.
+    ///
+    /// This is only relevant for Breakpad objects.
```
loewenheim: I don't think this is literally true. The limit should in principle apply to any object type that allows inlinees. It's just that the example we have involves Breakpad.

We can extend this to other object types as a follow-up.

symbolic-testutils/fixtures/regression/breakpad_deep_inline.sym
tobias-wilfert: Yes, was just not sure which of the 2 to prefer.

tobias-wilfert: Changed it.

symbolic-debuginfo/src/object.rs
tobias-wilfert: Fair enough, you thinking of something like: `Currently this is only relevant for Breakpad objects"?

Dav1dde (approved):

symbolic-debuginfo/src/breakpad.rs
```diff
@@ -1172,8 +1147,36 @@ impl<'data> Parse<'data> for BreakpadObject<'data> {
         Self::test(data)
     }
 
-    fn parse_with_opts(data: &'data [u8], _opts: ParseObjectOptions) -> Result<Self, Self::Error> {
-        Self::parse(data)
+    fn parse_with_opts(data: &'data [u8], opts: ParseObjectOptions) -> Result<Self, Self::Error> {
```
loewenheim, line 1183: Nit: this function should be next to `parse`.

symbolic-debuginfo/src/object.rs
```diff
@@ -137,6 +137,11 @@ pub struct ParseObjectOptions {
     ///
     /// This is only relevant for source bundles and PPDB objects.
     pub max_decompressed_embedded_source_size: Option<usize>,
+
+    /// The maximum inline nesting depth to process.
+    ///
+    /// This is only applies for Breakpad objects (for the time being).
```
loewenheim:
```suggestion
    /// This is only relevant for Breakpad objects (for the time being).
```

loewenheim (changes requested):

symbolic-debuginfo/src/breakpad.rs
tobias-wilfert: Makes sense did it the way that `SourceBundle` does it with a seperate `parse_with_opts`.

loewenheim (approved):

PR #1015, opened 2026-07-09
fix(debuginfo): Add some dwarf debuginfo recursion limits

klochek:
Ref: [INGEST-967](https://linear.app/getsentry/issue/INGEST-967)


symbolic-debuginfo/src/dwarf.rs
```diff
@@ -563,6 +558,12 @@ struct DwarfUnit<'d, 'a> {
 }
 
 impl<'d, 'a> DwarfUnit<'d, 'a> {
+    /// The maximum depth to recurse to in order to resolve a function name.
+    const MAX_RESOLVE_FUNCTION_DEPTH: u8 = 32;
+
+    /// The maximum depth to recurse to when parsing inlined functions.
+    const MAX_PARSE_INLINEE_DEPTH: u32 = 256;
```
loewenheim: For this you could piggyback on the `max_inline_depth` option introduced in https://github.com/getsentry/symbolic/pull/1014.

symbolic-debuginfo/src/macho/mod.rs
```diff
@@ -70,17 +71,6 @@ impl<'d> MachObject<'d> {
         matches!(MachArchive::is_fat(data), Some(false))
     }
 
-    /// Tries to parse a MachO from the given slice.
```
loewenheim, line 73: I would leave this function in place. The removal is breaking for no real reason, IMO.

symbolic-debuginfo/src/wasm/parser.rs
```diff
@@ -48,9 +52,14 @@ impl BitVec {
...
-    /// Tries to parse a WASM from the given slice.
-    pub fn parse(data: &'data [u8]) -> Result<Self, WasmError> {
+impl<'d> Parse<'d> for WasmObject<'d> {
+    type Error = WasmError;
+
+    fn test(data: &[u8]) -> bool {
+        Self::test(data)
+    }
+
+    fn parse_with_opts(data: &'d [u8], popts: ParseObjectOptions) -> Result<Self, Self::Error> {
```
loewenheim: Just for consistency:
```suggestion
    fn parse_with_opts(data: &'d [u8], opts: ParseObjectOptions) -> Result<Self, Self::Error> {
```

symbolic-debuginfo/src/pe.rs
```diff
@@ -82,16 +83,6 @@ impl<'data> PeObject<'data> {
         )
     }
 
-    /// Tries to parse a PE object from the given slice.
-    pub fn parse(data: &'data [u8]) -> Result<Self, PeError> {
```
loewenheim, line 86: Same as for macho.

loewenheim (commented):
LGTM apart from the function removals. I think it's not uncommon in Rust for functions/methods to be defined both on a trait and on a type directly; this lets one use the function even if the trait isn't in scope. I would leave the existing functions in place if there isn't a strong reason to remove them.

loewenheim (approved):

PR #1017, opened 2026-07-10
feat(cfi): decode ARM EHABI (.ARM.exidx/.ARM.extab) unwind tables

aholza:
## Problem

GCC's default (and often *only*) unwind-table format on ARM32 Linux is
EHABI, not DWARF `.eh_frame`/`.debug_frame`. Many ARM32 shared objects ship
an empty or absent `.eh_frame` and rely on `.ARM.exidx`/`.ARM.extab`
exclusively — this is what libgcc's `_Unwind_*` and debuggers like GDB use
on such binaries. `symbolic-cfi`'s ELF path only extracted DWARF CFI, so for
these binaries it produced no CFI at all, forcing consumers (e.g.
Symbolicator) to fall back to stack scanning for the entire unwind.

Concretely: symbolicating a real ARM32/Qt6 minidump crash produced a
6-frame backtrace before hitting `trust: scan` garbage, versus GDB's full
28-frame trace on the same core.

## Solution

Adds a new `arm_exidx` module to `symbolic-cfi` that decodes EHABI unwind
tables and converts them to the same breakpad `STACK CFI` text format the
existing DWARF/compact-unwind paths already produce — no new output format,
same pattern as `process_macho`'s Mach-O compact-unwind conversion. Wired
into `AsciiCfiWriter`'s ELF path alongside DWARF extraction (both are
emitted; EHABI fills in what DWARF is missing).

Two EHABI encodings are handled (per ARM IHI 0038, "Exception Handling
ABI for the ARM Architecture"):

- **Compact model** (personality routines 0/1): the common case, opcodes
  either inlined in the `.ARM.exidx` entry or pointed to in `.ARM.extab`.
- **Generic model** (personality 2+, e.g. `__gxx_personality_v0` /
  C++ exception frames): the personality routine's code is never executed.
  Per LLVM libunwind's `decode_eht_entry()`, every personality routine
  GCC/Clang emit for ARM — standard compact ones and
  `__gxx_personality_v0` alike — is built on
  `__gnu_unwind_frame`/`__gnu_unwind_execute`, which reads the same compact
  opcode stream from the word immediately following the personality
  pointer, just with a 1-byte size-prefix header instead of compact model
  1's 2-byte marker+count header. `interpret()`/`write_record()` are
  unchanged; only opcode-stream extraction differs.

Unsupported and left as a documented gap: VFP/FPA/Intel-Wireless-MMX
opcodes, and truly arbitrary (non-`__gnu_unwind_frame`-based) custom
personality routines, which are out of reach for a static decoder by
definition.

## Testing

- 30 unit tests + doctests in `symbolic-cfi`, several cross-verified
  byte-for-byte against `readelf --unwind` output on a real
  `libQt6Core.so.6`, including a generic-model (`__gxx_personality_v0`)
  case built from the exact bytes of a real function's EHABI entry.
- End-to-end: symbolicated a real ARM32/Qt6 SIGSEGV minidump through
  Symbolicator with this patch applied. The crashed thread now unwinds
  with **zero scan frames** across all 26 frames (`context`/`inline`/`cfi`
  only), matching GDB's backtrace on the same core frame-for-frame.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Unreleased
+
+**Features**
+
+- `symbolic-cfi`: decode ARM EHABI (`.ARM.exidx`/`.ARM.extab`) unwind tables on ELF, covering both the compact model (personality 0/1) and the generic model (personality 2+, e.g. `__gxx_personality_v0`). Many ARM32 Linux binaries have no (or an empty) `.eh_frame`/`.debug_frame` and rely on EHABI exclusively, so this fills in CFI that was previously entirely missing for them. ([#XXXX](https://github.com/getsentry/symbolic/pull/XXXX))
+
 ## 14.0.0-alpha.1
 
 - No documented changes.
--- a/symbolic-cfi/src/arm_exidx.rs
+++ b/symbolic-cfi/src/arm_exidx.rs
@@ -26,6 +26,8 @@
 use std::io::Write;
 
 use symbolic_common::CpuFamily;
+use symbolic_debuginfo::dwarf::gimli::Endianity;
+use symbolic_debuginfo::dwarf::Endian;
 
 use crate::CfiError;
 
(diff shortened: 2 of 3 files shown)


Dav1dde:
Please state your personal motivation for this change.

Dav1dde:
I think a better approach is what we do for macho compact. Expose additional functionality from `symbolic-debuginfo` which parses the unwind ops, then use them in the CFI writer.

Please don't just give this as an instruction to your LLM, this requires significant human steering and proper documentation. Again, see how this is done in `macho/compact.rs`.

Let me know if you have interest on working on this. I will close this for now as this looks more like just a one-shot prompt missing vital information and documentation making this reviewable.

PR #1018, opened 2026-07-16
deps: Update `pdb-addr2line` to 0.12.0

loewenheim:
This change is long past due. By doing this, we move from our unmaintained `pdb` crate to the maintained `pdb2` fork (https://crates.io/crates/pdb2). As https://github.com/afranchuk/pdb/compare/master...getsentry%3Apdb%3Amaster shows, `pdb` contains no significant changes that `pdb2` doesn't have.

### DIFF
--- a/symbolic-debuginfo/src/pdb/mod.rs
+++ b/symbolic-debuginfo/src/pdb/mod.rs
@@ -7,8 +7,8 @@ use std::fmt;
 use std::io::Cursor;
 use std::sync::Arc;
 
-use elsa::FrozenMap;
-use parking_lot::RwLock;
+use elsa::sync::FrozenMap;
+use parking_lot::Mutex;
 use pdb_addr2line::pdb::{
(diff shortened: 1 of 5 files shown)


symbolic-debuginfo/src/pdb/mod.rs
loewenheim, line 1: `RwLock`/`Mutex` changes in this file are down to the fact that `pdb_addr2line::TypeFormatter::new_from_parts` now requires its first parameter to be `Sync`.

loewenheim, line 39: The fact that we're exposing `pdb` here, as well as the hidden `inner` method down below, strictly speaking make this a breaking change.

jjbayer: LOL "remove once abstraction is complete" looks like an old comment.

jjbayer (approved):

PR #1022, opened 2026-07-21
fix(xml): Use `xml10_content` instead of `decode` for EOL handling

klochek:


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+**Fixes**
+
+- Use `xml10_content` instead of `decode` for EOL handling ([#1022](https://github.com/getsentry/symbolic/pull/1022))
```
Dav1dde:
```suggestion
- Use `xml10_content` instead of `decode` for EOL handling. ([#1022](https://github.com/getsentry/symbolic/pull/1022))
```

We should aim to describe in the changelog the effect for the user reading the changelog. For example "Fixed issues when decoding <certain xml elements> unreal reports".

Here specifically, I think it's also fine to just skip the changelog.

Dav1dde (approved):

CHANGELOG.md
loewenheim: Agreed.

PR #1023, opened 2026-07-22
fix(breakpad): Add inlinee nesting limit

klochek:


symbolic-debuginfo/src/breakpad.rs
```diff
@@ -148,6 +148,9 @@ pub enum BreakpadErrorKind {
 
     /// The architecture is invalid.
     InvalidArchitecture,
+
+    /// Too many inlinee nestings were generated.
+    TooManyInlineeNestings,
 }
 
 impl fmt::Display for BreakpadErrorKind {
```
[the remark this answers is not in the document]
loewenheim: Opened #1024 to address this.

symbolic-debuginfo/src/function_builder.rs
```diff
@@ -1,10 +1,40 @@
...
 
 use crate::base::{FileInfo, Function, LineInfo};
 use symbolic_common::Name;
+use thiserror::Error;
+
+/// An error originating in the FunctionBuilder.
+#[derive(Debug, Error)]
+#[error("{kind}")]
+pub struct FunctionBuilderError {
+    pub(crate) kind: FunctionBuilderErrorKind,
```
loewenheim, line 17: Is the `kind` actually accessible in any way? If not, we can keep the enum private and don't need to make it non-exhaustive.

symbolic-debuginfo/src/function_builder.rs
```diff
@@ -322,6 +352,9 @@ impl<'s> FunctionBuilderStack<'s> {
     }
 }
 
+/// Bounds to ensure that the maximum number of inlinees does not get too large.
+const MAX_NESTED_INLINEES: usize = 100000;
```
loewenheim, line 366: Is this meaningfully different from the inlinee depth you're addressing in your other PR?

klochek: Yes--one is a recursion limit (and so we need to be very aggressive about limiting stack depth,) and the other is a vec/allocation limit, which can and must be much higher.  I was thinking about actually making this limit a function of the number of inlinees (maybe 4x? 10x?) but I honestly don't know what good bounds there would look like (I'd need a crash of something interesting, _somewhere_ interesting, to get an idea) so this seemed like a fair compromise.

loewenheim: Ok, what I was missing is that `max_inline_depth` doesn't limit the number of inlinees of a function because they need not all form one stack. So I get now why this limit is necessary. Part of the confusion is the naming, I think: `MAX_NESTED_INLINEES` sounds like a synonym for `max_inline_depth`. If I'm not mistaken, this limit doesn't actually have anything to do with nesting at all, it is simply a hard limit on the inlinees of a function.

There's another problem, though: `MAX_NESTED_INLINEES` is only used in `ensure_proper_nesting`, which converts a heap of inlinees into a vec of the same length. So at that point, the allocation has already happened. I believe moving the check to `add_inlinee` makes more sense.

klochek: The issue this PR is trying to address is that there is _splitting_, arbitrarily, of inlinees, due to a weird compaction representation that we are converting from.  The details are in the linear ticket, but the gist is that you can craft a breakpoint file with a large number of "overlapping" inlinees (N) at a certain depth from a caller (D), which results in O(N*D) non-overlapping inlinees that are created and inserted into that result--_not_ that we're trying to tie down the absolute number of inlinees (which I gather would be caught by straightforward input limits elsewhere in symbolic--unless I assume too much!)

symbolic-debuginfo/src/function_builder.rs
```diff
@@ -1,10 +1,63 @@
...
+#[error("{kind}")]
+pub struct FunctionBuilderError {
+    kind: FunctionBuilderErrorKind,
+}
+
+/// The kind of error origination in the FunctionBuilder.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum FunctionBuilderErrorKind {
+    /// We generated too many nestings (address splits) for an inlinee.
+    TooManyInlineeNestings,
```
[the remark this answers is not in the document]
loewenheim: This seems like a valid concern, but can perhaps be addressed separately.

symbolic-debuginfo/src/function_builder.rs
```diff
@@ -1,10 +1,63 @@
...
+
+impl From<FunctionBuilderError> for DwarfError {
+    fn from(value: FunctionBuilderError) -> Self {
+        match value.kind {
+            FunctionBuilderErrorKind::TooManyInlineeNestings => {
+                DwarfErrorKind::CorruptedData.into()
+            }
+        }
+    }
+}
```
loewenheim: This will create a `DwarfError` with no source, i.e. it won't be possible to tell that the underlying error is a `FunctionBuilderError`. You could do
```suggestion
impl From<FunctionBuilderError> for DwarfError {
    fn from(value: FunctionBuilderError) -> Self {
        Self::new(DwarfErrorKind::CorruptedData, value)
    }
}
```
but it would have to be in `dwarf.rs` because `DwarfError::new` is private.

You also don't have to use `CorruptedData` for this; you can introduce a new `DwarfErrorKind` if none of the existing ones seem apposite.

symbolic-debuginfo/src/function_builder.rs
loewenheim: After actually reading the issue and thinking about it I get the problem now. Sorry for the confusion!

loewenheim (approved):

PR #1033, opened 2026-07-28
ref(symcache): Remove support for v7 and v8 symcache versions

Dav1dde:
Deletes v7 and v8 symcache versions. Symbolic 14 will either only support v9 + v10 or also drop v9 at a later point.

Closes: https://linear.app/getsentry/issue/INGEST-1093/delete-symcache-v7v8


symbolic-testutils/fixtures/symcache/README.md
loewenheim, line 1: Thanks for actually documenting this :bow:

loewenheim (approved):

PR #1035, opened 2026-07-28
ref(symcache): Re-org module structure, simplify multi version implementations

Dav1dde:
Instead of having version files split across two top level modules `raw/{version}.rs` and `{version}/lookup.rs`, raw is also now moved into the top level `{version}` module.

Also:
- Changes some visibility qualifiers from `pub(crate)` to either `pub` or private.
- Implements a macro to more easily manage different symcache versions.
- Instead of referring to names as `FooV9` it uses a fully qualified lookup `v9::Foo`


loewenheim (approved):
Big improvement, thanks for doing this!

PR #1036, opened 2026-07-30
feat(variables): Write variables to symcaches

Dav1dde:
Write variable information into symcaches. I am holding off on writing more docs until we get closer to stability and release of the format, I expect it to still evolve a bit over time.

This change uses parts of the `reserved` header from the original V9 symcache, to optionally extend past the previous end of the symcache with variable information.

It also introduces not just a flag but a version counter of that additional section to allow for iteration on the cache without needing to rebuild existing caches which work fine for symbolication.

This PR does not yet implement any of the reading APIs, it only implements writing the variable information.

```
└─ % ./run symcache_debug --debug-file symbolic-testutils/fixtures/linux/crash.debug --write-cache-file --symcache-file ./test.symc --report
+ cargo run --quiet --package symcache_debug -- --debug-file symbolic-testutils/fixtures/linux/crash.debug --write-cache-file --symcache-file ./test.symc --report
Cache file written to ./test.symc
Cache info:
SymCache {
    version: 9,
    debug_id: DebugId {
        uuid: "c0bcc3f1-9827-fe65-3058-404b2831d9e6",
        appendix: 0,
    },
    arch: Amd64,
    files: 55,
    functions: 697,
    source_locations: 8431,
    ranges: 6975,
    string_bytes: 51929,
    variables: 548,
    variable_locations: 1548,
    types: 12,
}
```

### DIFF
--- a/symbolic-symcache/src/v9/raw.rs
+++ b/symbolic-symcache/src/v9/raw.rs
@@ -2,9 +2,22 @@
 //!
 //! V9 adds source server information support to the File structure.
 
+use core::fmt;
+
 use watto::Pod;
 
 use symbolic_common::DebugId;
+
+/// The current [`Header::variable_header`] version.
+///
+/// On a version mismatch, the sections referenced in the [`VariableHeader`] must not be parsed.
+pub const VARIABLE_HEADER_VERSION: u16 = 1;
+
+/// This [`FunctionVariables`] is a sentinel for functions without variables.
+pub const NO_VARIABLES: FunctionVariables = FunctionVariables {
+    variable_idx: u32::MAX,
+    num_variables: 0,
+};
 
 /// This [`SourceLocation`] is a sentinel value that says that no source location is present here.
 /// This is used to push an "end" range that does not resolve to a valid source location.
@@ -38,9 +51,30 @@ pub struct Header {
     /// Total number of bytes used for string data.
     pub string_bytes: u32,
 
+    /// Version of the optional [`VariableHeader`].
+    ///
+    /// The version is `0` there is no variable header. This allows incompatible changes
+    /// to variable and type information in a symcache without requiring a new symcache version.
+    ///
+    /// Once the format is stable, the variable header will no longer be optional in symcache
+    /// version 10.
+    pub variable_header: u16,
+
     /// Some reserved space in the header for future extensions that would not require a
     /// completely new parsing method.
-    pub _reserved: [u8; 16],
+    pub _reserved: [u8; 14],
+}
+
+/// Optional header appended to the [`Header`] if the symcache contains variable information.
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[repr(C)]
+pub struct VariableHeader {
+    /// Number of included [`Variable`]'s.
+    pub num_variables: u32,
+    /// Number of included [`VariableLocation`]'s.
+    pub num_variable_locations: u32,
+    /// Number of included [`Type`]'s.
+    pub num_types: u32,
 }
 
 /// Serialized Function metadata in the SymCache.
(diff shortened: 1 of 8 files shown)


symbolic-symcache/src/writer.rs
```diff
@@ -408,6 +456,113 @@ impl<'a> SymCacheConverter<'a> {
...
+                    .checked_add(location.size)
+                    .is_none_or(|v| v > u32::MAX as u64)
+                {
+                    continue;
+                }
+
+                let (kind, data) = match location.location {
+                    di::Location::Register { id } => (0, id.into()),
+                    di::Location::FrameOffset { offset } => (1, offset as i32),
+                };
```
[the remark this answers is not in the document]
Dav1dde: There are a lot of assumption of size, this is yet another one.

symbolic-symcache/src/v9/raw.rs
```diff
@@ -38,9 +51,30 @@ pub struct Header {
     /// Total number of bytes used for string data.
     pub string_bytes: u32,
 
+    /// Version of the optional [`VariableHeader`].
+    ///
+    /// The version is `0` there is no variable header. This allows incompatible changes
```
loewenheim: I assume this should either be 
```suggestion
    /// If the version is `0` there is no variable header. This allows incompatible changes
```
or

```suggestion
    /// The version is `0` if there is no variable header. This allows incompatible changes
```

symbolic-symcache/src/v9/raw.rs
```diff
@@ -113,11 +160,181 @@ pub struct SourceLocation {
...
+    pub _reserved: [u8; 2],
+}
+
+/// A representation of a variable location.
+#[derive(Clone, Debug, Hash, PartialEq, Eq)]
+#[repr(C)]
+pub struct VariableLocation {
+    /// First instruction address (pc) where this location is valid.
+    pub start: u32,
+    /// Size of the location, indicating the last instruction address where this location is valid.
```
loewenheim: Nit: 
```suggestion
    /// Size of the location, indicating the first instruction address where this location is no longer valid.
```

symbolic-symcache/src/v9/raw.rs
```diff
@@ -113,11 +160,181 @@ pub struct SourceLocation {
...
+    /// The first location associated with the variable (reference to a [`VariableLocation`]).
+    pub location_idx: u32,
+    /// The amount of locations.
+    pub num_locations: u32,
+    /// The inlined depth of the owning function.
+    ///
+    /// If `0` the variable belongs to the outermost function. This is essentially a way to track
+    /// scopes, limited to function scopes. Each level of depth is a newly nested scope.
+    ///
+    /// The depth can be recovered by tracing [`SourceLocation::inlined_into_idx`] entries.
```
loewenheim: I don't understand this sentence, what does it mean to "recover" the depth?

loewenheim (commented):
Out of interest, why are variables always attributed to the base function?

symbolic-symcache/src/v9/raw.rs
Dav1dde: To select the correct set of variables per inlinee, you need to filter the variables on the depth. The depth isn't stored anywhere, so we'll have to trace the `SourceLocation` backwards via `inlined_into_idx` to know what the function depth at that point is. Once we have the depth, we can find the variables for that inlined frame.

Dav1dde:
> Out of interest, why are variables always attributed to the base function?

I'm still also wrapping my head around this, inlined functions in the dwarf are tied to the "base function", as a function can be inlined multiple times:

```rs
#[inline(always)]
fn foo() {}

#[inline(always)]
fn bar(x: bool) {
    // Recursively inlined.
    if x { bar(false); }
}

fn test1() {
    foo();
    foo();
}

fn test2() {
    foo()
}
```

Since usages/instances of the inlined functions (`foo` and `bar`) are at different locations, the locations (pc ranges) also differ. So each inline location needs different variable locations and these are per base. But we still need a way to recover the depth, so we can show the variables per source location.

symbolic-symcache/src/v9/raw.rs
Dav1dde: I reworded this comment a bit, is that better?

loewenheim:
> Since usages/instances of the inlined functions (foo and bar) are at different locations, the locations (pc ranges) also differ. So each inline location needs different variable locations and these are per base. But we still need a way to recover the depth, so we can show the variables per source location.

Right. Multiple inlined ranges may point to the same function `f`, if `f` got inlined multiple times. If the variables were associated with `f`, then all those inlined versions would share the same variable locations, but they don't necessarily. Therefore the only thing to do is to associate the variables to a function that isn't inlined.

symbolic-symcache/src/v9/raw.rs
loewenheim: Much better, yes.

loewenheim (approved):

PR #1037, opened 2026-08-03
feat(variables): Expose variable information from symcaches

Dav1dde:
Implements the look of variables and types for SymCaches. This is a follow up to #1036 which already serialized the information into the cache.

This will be followed up with a few improvements how to organize the code, for example currently writing and reading are quite far apart in the code and it is easy to for these to drift apart. I have some ideas of how to make that better.

Also includes:
- Some renames for location and variable types, as they are exposed from the root of `symbolic_debuginfo`
- Re-generate the test fixtures to include variable and type information

Closes: INGEST-1092


symbolic-symcache/src/v9/lookup.rs
```diff
@@ -111,13 +152,66 @@ impl<'data> SourceLocation<'data, '_> {
...
+    pub fn variables(&self) -> Variables<'data, 'cache> {
+        let (current, remaining) = self
+            .cache
+            .function_variables
+            .get(self.source_location.function_idx as usize)
+            // We can potentially optimize here and use a binary search as variables are
+            // sorted by their depth. This likely requires some minimum `num_variables` breakpoint
+            // to beat the linear search though.
+            .map(|fv| (fv.variable_idx, fv.num_variables))
+            .unwrap_or((u32::MAX, 0));
```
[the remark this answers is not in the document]
Dav1dde: I think this is correct, will have to check this out.

symbolic-symcache/src/v9/raw.rs
```diff
@@ -216,6 +216,22 @@ pub struct Type {
...
+    }
+
+    pub fn as_pointer(&self) -> Option<&PointerType> {
+        match self.kind {
+            1 => Some(unsafe { &self.ty.pointer }),
+            _ => None,
+        }
+    }
+}
+
```
tobias-wilfert, line 234: Is there a specific reason to go with the 2 functions here rather than have a function that returns an enum?

examples/addr2line/src/main.rs
```diff
@@ -6,7 +6,7 @@ use clap::builder::ValueParser;
 use clap::{value_parser, Arg, ArgAction, ArgMatches, Command};
 
 use symbolic::common::{ByteView, Language, Name, NameMangling};
-use symbolic::debuginfo::{Function, Location, Object, ObjectDebugSession, Type, TypeRef};
+use symbolic::debuginfo::{Function, Object, ObjectDebugSession, Type, TypeRef, VariableLocation};
```
tobias-wilfert, line 9: Could we not also have dropped the `pub use self::variable::*;` instead of doing the renames?

symbolic-symcache/src/v9/lookup.rs
```diff
@@ -111,13 +152,78 @@ impl<'data> SourceLocation<'data, '_> {
             .get_function(self.source_location.function_idx)
             .unwrap_or_default()
     }
+
+    /// The variables available at this instruction.
+    pub fn variables(&self) -> Variables<'data, 'cache> {
```
tobias-wilfert, line 157: meta: Is your analyzer/setup able to find the 'references' for this (mine sadly isn't) (I know it is used but would be cool if my IDE did too)?

tobias-wilfert (approved):
Looks good to me, just small questions but nothing blocking.

symbolic-symcache/src/v9/raw.rs
Dav1dde: No, partially just my laziness (naming is hard) and the hope to be able to return a `debuginfo::Type` which didn't work out. Will introduce an enum.

examples/addr2line/src/main.rs
Dav1dde: That's possible, it was an explicit decision from before to mirror what other types in `base.rs` do, since the types end up being referenced by other types which are already imported top level (`Function`).

Gonna add a note to the v14 release issue to reconsider this.

symbolic-symcache/src/v9/lookup.rs
Dav1dde: Unfortunately not, probably because it is only referenced in a macro to support multiple symcache versions.

symbolic-symcache/src/v9/raw.rs
Dav1dde: Going to do that in a follow-up where I want to address the conversions.

symbolic-symcache/src/v9/lookup.rs
```diff
@@ -57,6 +65,37 @@ impl<'data> SymCache<'data> {
...
+                size: TypeSize::Bytes(primitive.size.into()),
+                encoding: match primitive.encoding {
+                    0 => Some(PrimitiveTypeEncoding::Boolean),
+                    1 => Some(PrimitiveTypeEncoding::Address),
+                    2 => Some(PrimitiveTypeEncoding::SignedInt),
+                    3 => Some(PrimitiveTypeEncoding::UnsignedInt),
+                    4 => Some(PrimitiveTypeEncoding::SignedChar),
+                    5 => Some(PrimitiveTypeEncoding::UnsignedChar),
+                    6 => Some(PrimitiveTypeEncoding::Float),
+                    7 => Some(PrimitiveTypeEncoding::ComplexFloat),
```
szokeasaurusrex, line 83: Were these numbers selected arbitrarily? Would it be worth it to define constants for them? 

They differ from the `DW_ATE_*` constants used elsewhere

symbolic-symcache/src/v9/lookup.rs
```diff
@@ -57,6 +65,37 @@ impl<'data> SymCache<'data> {
...
+                    _ => None,
+                },
+            }));
+        }
+
+        if let Some(pointer) = raw_ty.as_pointer() {
```
szokeasaurusrex, line 89: [nit] if I am reading this correctly, this can be replaced with `else if` without changing behavior

```suggestion
        else if let Some(pointer) = raw_ty.as_pointer() {
```

symbolic-symcache/src/v9/lookup.rs
```diff
@@ -57,6 +65,37 @@ impl<'data> SymCache<'data> {
...
+            return Some(Type::Pointer(PointerType {
+                size: TypeSize::Bytes(pointer.size.into()),
+                pointee: (pointer.pointee_idx != u32::MAX).then_some(TypeId(pointer.pointee_idx)),
+            }));
+        }
+
+        None
```
szokeasaurusrex, line 96: In what case can we reach this line? It seems to be unlikely; if I understand it correctly, `raw_ty` should either be a primitive or a pointer. If we don't expect to hit this, perhaps we should instead `panic!` and/or mark as `unreachable!`?

symbolic-symcache/src/v9/raw.rs
```diff
@@ -216,6 +216,22 @@ pub struct Type {
...
+        }
+    }
+
+    pub fn as_pointer(&self) -> Option<&PointerType> {
+        match self.kind {
+            1 => Some(unsafe { &self.ty.pointer }),
+            _ => None,
+        }
+    }
+}
```
szokeasaurusrex, line 233: Both of these functions seem unsafe to me. 

I suppose the safety assumption would be that `self.kind` accurately reflects the type of `self.ty`. However, with both of those fields being public, I don't think it is possible to guarantee this, so there's a safety risk.

I would consider making the fields private to mitigate this

szokeasaurusrex (commented):
Added a couple minor suggestions and questions

symbolic-symcache/src/v9/lookup.rs
Dav1dde: Not arbitrary, they are written to the symcache as `PrimitiveTypEncoding as u8`, this is the inverse. I cleaned this up a bit in #1039

symbolic-symcache/src/v9/lookup.rs
Dav1dde: Unfortunately that triggers quite an annoying clippy lint to write it it as `else { raw_ty.as_pointer().map() }` which formats much worse.

Also changed/addressed in #1039 using an enum.

symbolic-symcache/src/v9/raw.rs
Dav1dde: This is also reworked, but it is also safe. Even if the `kind` is something else, since all members are Pods, all bit patterns are sound and safe, though they might not make any sense.

#1039 introduces a macro which deals with all the unsafety and adds some justifications for them.

symbolic-symcache/src/v9/lookup.rs
Dav1dde: If we load a symcache with a mismatched version or a symcache which was corrupted. There is some checks in place to make sure this doesn't happen but there is no guarantee.

The symcache has the general assumption that everything is correct and if you load garbage you have sound code but undefined output, this just follows it.

PR #1038, opened 2026-08-05
feat(variables): add initial version of `variables.c` test fixture

JoshuaMoelans:
- build.sh to generate binary(with O0 to avoid optimizing out everything)
- `test_objects__elf_variables.snap` for validating
- updated `test_objects` with new `test_elf_variables`
- smol README with instructions how to run/rebuild binary/refresh the snapshot/add coverage/listing uncovered items

resolves https://linear.app/getsentry/issue/INGEST-1087/dwarf-add-new-test-fixture-for-variables

To be followed up with #1042 (or maybe we merge that one in here, not sure what flows more naturally?)


symbolic-debuginfo/tests/test_objects.rs
```diff
@@ -93,59 +93,65 @@ impl fmt::Debug for FunctionsDebug<'_, '_> {
...
+/// Helper to create neat snapshots for the variables of a function tree.
+///
+/// Unlike [`FunctionsDebug`] this leaves out line records and prints location ranges relative to
+/// the start of their function, so the snapshot only changes when variable debug info changes and
+/// not when the fixture is merely relinked or relocated.
+struct VariablesDebug<'data, 'session> {
+    session: &'session ObjectDebugSession<'data>,
+    functions: &'session [Function<'session>],
+    depth: usize,
+}
```
JoshuaMoelans: This is still very similar to `FunctionsDebug`, so maybe we just want to consolidate them into one struct (with some flags for what to print etc...). OTOH since we introduce the fixture specifically for Variables, maybe keeping it separate is ok too?

symbolic-testutils/fixtures/linux/variables/build.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+# Builds the ELF/DWARF variable fixture from variables.c.
+#
+# Run this via Docker so the output does not depend on your machine:
+#
+#   docker run --rm --platform linux/amd64 \
+#       -v "$PWD/..:/testutils" -w /testutils/variables gcc:14 ./build.sh
+#
+# See README.md for details.
```
JoshuaMoelans, line 9: Seemed like the least convoluted (and most reproducible) way to build the `variables` fixture, but open to suggestions on simplifying this.

\+ do we want to run this in CI too, or just make sure that when we change the `variables.c` we actually remember to build & commit the new binary?

symbolic-testutils/variables/build.sh
```diff
@@ -0,0 +1,12 @@
...
+#
+# Run this via Docker so the output does not depend on your machine:
+#
+#   docker run --rm --platform linux/amd64 \
+#       -v "$PWD/..:/testutils" -w /testutils/variables gcc:14 ./build.sh
+#
+# See README.md for details.
+set -eu
+
+gcc -g -gdwarf-5 -O0 -o ../fixtures/linux/variables variables.c
```
JoshuaMoelans: (think about how this changes the DWARF format too + maybe add second test for inlining/ 'USED' , or 'USED' macro or 'volatile'...)

symbolic-testutils/fixtures/linux/variables/variables.c
```diff
@@ -0,0 +1,104 @@
...
+ * type living on the abstract DIE, which symbolic does not follow for
+ * variables. Both `param` and `doubled` have plain frame-base locations at
+ * -O0, so `inlined` rendering with no variables at all in the snapshot is
+ * purely the missing origin lookup.
+ */
+static inline __attribute__((always_inline)) int inlined(int param)
+{
+    int doubled = param * 2;
+    return doubled + 1;
+}
```
JoshuaMoelans, line 80: (IMO we can keep the fixture itself relatively clean of this slightly noisy type of clanker-aided comments, the PR with comments/intermittent commits can tell the full story if anyone's interested)
```suggestion
/*
* Inlined variables are not supported yet; DW_AT_abstract_origin is not followed 
* so we don't resolve these for now
*/
static inline __attribute__((always_inline)) int inlined(int param)
{
    int doubled = param * 2;
    return doubled + 1;
}
```

symbolic-testutils/variables/build.sh
JoshuaMoelans: added inlining test in https://github.com/getsentry/symbolic/pull/1038/commits/9e24c3a72c618c927d3ed68524be28a609c593a0

JoshuaMoelans: and even more coverage, separately, in #1042

symbolic-debuginfo/tests/test_objects.rs
```diff
@@ -47,14 +47,27 @@ impl fmt::Debug for FilesDebug<'_> {
...
 
+/// Selects which parts of a function tree [`FunctionsDebug`] prints, and how.
+#[derive(Clone, Copy)]
+enum DebugMode {
+    /// The full function tree: absolute addresses, line records, and variables.
+    Full,
+    /// Only the variables of each function, with location ranges relative to the start of their
+    /// function, so a snapshot only changes when variable debug info changes, not when the
+    /// fixture is merely relinked or relocated. Toolchain changes might still churn the snapshot.
+    RelativeVariables,
```
Dav1dde: What problem is this trying to solve? Why does a toolchain change not impact the snapshot, variables may now be stored in different locations.

symbolic-testutils/variables/README.md
```diff
@@ -0,0 +1,69 @@
...
+`fixtures/linux/variables` (ELF/DWARF 5, x86-64), asserted by `test_elf_variables` in
+`symbolic-debuginfo/tests/test_objects.rs`.
+
+`variables.c` is our fixture which is meant to grow along with symbolic's variable support.
+
+To run the test without rebuilding the snapshot, use:
+
+```sh
+cargo test -p symbolic-debuginfo --test test_objects test_elf_variables
+```
```
Dav1dde: This will go stale, I would not reference the test here.

symbolic-testutils/variables/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Rebuilding
+
+If you make changes to `variables.c`, updating the snapshot takes two steps. First, rebuild the
+fixture — run this from *this* directory; it writes `../fixtures/linux/variables`:
```
Dav1dde: I'd follow what is done for other fixtures, store the build script next to the binary in its own directory.

symbolic-testutils/variables/README.md
```diff
@@ -0,0 +1,69 @@
...
+version, your local checkout path would be embedded in the debug info and the binary would differ
+from the committed one.
+
+Second, refresh the snapshot and check the diff is what you expected:
+
+```sh
+cargo insta test --accept -p symbolic-debuginfo --test test_objects
+```
+
+To keep remote up-to-date, commit the rebuilt binary together with the updated snapshot.
```
Dav1dde: This is a lot of fluff we don't really care about. Also I would not reference the exact test here, just a note that you need to re-run the tests, but even that is kinda implied, when you change the fixture, the tests also change (also enforced in CI).

symbolic-testutils/fixtures/linux/variables/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+Second, refresh the snapshot and check the diff is what you expected:
+
+```sh
+cargo insta test --accept -p symbolic-debuginfo --test test_objects
+```
+
+To keep remote up-to-date, commit the rebuilt binary together with the updated snapshot.
+
+## Adding coverage
```
Dav1dde: I am not sure what this is supposed to tell me other than "prefer to add a new function over changing existing ones". Where this is a fine suggestions, though I don't agree with the reasoning, there is nothing wrong with the fixture changing.

symbolic-testutils/fixtures/linux/variables/variables.c
Dav1dde: Yes please, humans read the text and they don't have an infinite context/attention window either.

Dav1dde (changes requested):

symbolic-debuginfo/tests/test_objects.rs
JoshuaMoelans: @Dav1dde do you mean in general, using the RelativeVariables over the Full DebugMode? Or just the phrasing of the comment? 

The last sentence is meant to indicate that toolchain changes do indeed churn the snapshot, but if it doesn't read cleanly I can reword it for sure.

symbolic-testutils/fixtures/linux/variables/README.md
JoshuaMoelans: Do you feel like we should just cut out the reasoning in the paragraph (I think keeping the first sentence as a suggestion by itself is probably enough). It wasn't meant as a "try never to change the existing parts of the fixture", more of a "if you want to keep changes to the fixture minimal, adding functions is self-contained" hint.

symbolic-debuginfo/tests/test_objects.rs
JoshuaMoelans: IIUC the RelativeVariables that I (+ clanker) introduced was just meant to simplify the variable extraction snapshot, but after talking with @szokeasaurusrex  about it too, it makes sense to just reuse the existing debug format since we never rewrite the snapshot by hand (and just regenerate it). This simplified the PR quite a bit https://github.com/getsentry/symbolic/pull/1038/commits/008b8d7dc292136f75162e6f7377bd06683e1092

loewenheim (approved):
LGTM, assuming the snapshot contains what you expect it to.

PR #1046, opened 2026-08-14
fix(debuginfo): Increase inlinee limits

klochek:


Dav1dde (approved):
Just rolling the dice (no strong feelings either way, current numbers probably fine): Maybe 512 and 250k (200k is pretty close to the actual number) and 600 might be quite close to the max depth.

PR #1050, opened 2026-08-20
fix(debuginfo): Make `Function`'s `Drop` iterative

loewenheim:
This requires some minor changes in tests because implementing `Drop` means you can't partially move a type and have to be more careful with declaration order.

ref: INGEST-1133


loewenheim:
Consensus in the team is that this isn't worth the complication.

PR #1051, opened 2026-08-20
fix(debuginfo): Limit `Function`'s `Debug`

loewenheim:
The `Debug` impl of `Function` is naturally recursive because the type is. Instead of rewriting the impl to be iterative I decided to simply institute a depth limit of 16. This means that after 16 levels of inlinees the `inlinees` field is displayed as `[..]`. I believe that should be plenty for all practical cases.

ref: INGEST-1133


jjbayer (approved):
A test case with `insta::assert_debug_snapshot` would be nice here.

loewenheim:
Consensus in the team is that this isn't worth the complication.

PR #1052, opened 2026-08-24
fix(demangle): Patch segfault in Swift demangler

loewenheim:
It's possible to trigger a null dereference with some mangled names in the Swift demangler. This patches the vulnerability for our vendored copy. We'll also report the problem upstream.

ref: INGEST-1134


Dav1dde (approved):

loewenheim:
Upstream PR: https://github.com/swiftlang/swift/pull/91675

PR #1053, opened 2026-08-25
fix(debuginfo): Increase nested inlinees limit

rodolfoBee:
Increasing the limit following previous issues with broken debug info.


symbolic-debuginfo/src/function_builder.rs
```diff
@@ -359,7 +359,7 @@ impl<'s> FunctionBuilderStack<'s> {
 }
 
 /// Bounds to ensure that the maximum number of inlinees does not get too large.
-const MAX_NESTED_INLINEES: usize = 250000;
+const MAX_NESTED_INLINEES: usize = 300000;
```
klochek: Let's push this to 400000, just in case.  Otherwise, good to go.

jjbayer (approved):

PR #1063, opened 2026-08-31
fix(debuginfo): Remove recursive function parsing

klochek:
Rather than futzing with bounds (that change as the code is changed at the stack frame size changes,) just rip the thing out and make an iterative solution for function parsing.  This also required de-recursing the writer in symcache, which was much easier (and so closely maps to what we had previously.)

For the dwarf processor, the basic idea is that parse_functions loops around, parsing a single tag each loop, maintaining a stack of in-progress functions.  The depth value of the tag drive whether or not to pop elements from the stack (thereby finishing them.)  Otherwise, we just push nested elements as we encounter them.  There is a nasty edge-case where a function that has invalid/empty ranges can still retain nested functions or inlinees (and the inlinees must be discarded, while the functions must not,) so we guard against when we detect a "dead code" function at the top of the stack, and handle the tags accordingly.


symbolic-debuginfo/src/dwarf/mod.rs
```diff
@@ -637,6 +637,165 @@ struct DwarfUnit<'d, 'a> {
     prefer_dwarf_names: bool,
 }
 
+type Builders<'a> = Vec<(Range, FunctionBuilder<'a>)>;
+
+struct RegularSubProgram<'a> {
+    index: usize,
+    depth: isize,
+    builders: Vec<(Range, FunctionBuilder<'a>)>,
```
jjbayer: Wanna use the `Builders` type alias here?

symbolic-debuginfo/src/dwarf/mod.rs
```diff
@@ -637,6 +637,165 @@ struct DwarfUnit<'d, 'a> {
...
+    call_line: u64,
+    language: Language,
+}
+
+struct DeadcodeSubProgram {
+    depth: isize,
+}
+
+enum InProgressSubProgram<'a> {
+    Deadcode(DeadcodeSubProgram),
```
jjbayer: nit

```suggestion
    DeadCode(DeadCodeSubProgram),
```

symbolic-debuginfo/src/dwarf/mod.rs
```diff
@@ -637,6 +637,165 @@ struct DwarfUnit<'d, 'a> {
...
+impl<'a> InProgressSubProgram<'a> {
+    fn language(&self) -> Language {
+        match self {
+            InProgressSubProgram::Deadcode(_) => Language::Unknown,
+            InProgressSubProgram::Regular(p) => p.language,
+            InProgressSubProgram::Inlined(p) => p.language,
+        }
+    }
+
+    fn own_builder_mut(&mut self) -> Option<&mut Builders<'a>> {
```
jjbayer: Why is this called `own_builder_mut` rather than `builders_mut`?

symbolic-debuginfo/src/dwarf/mod.rs
```diff
@@ -637,6 +637,165 @@ struct DwarfUnit<'d, 'a> {
...
+                    output.functions.push(builder.finish()?);
+                }
+
+                Ok(())
+            }
+
+            // Inlinees don't output anything directly, they just contribute to the running list
+            // of function builders.
+            InProgressSubProgram::Inlined(p) => {
+                let Some(builders) = function_stack[p.owning_function_idx].own_builder_mut() else {
```
jjbayer: Are the `[]` always safe here?

symbolic-symcache/src/writer.rs
```diff
@@ -456,6 +471,321 @@ impl<'a> SymCacheConverter<'a> {
         }
     }
 
+    /// Processes an individual [`Function`], adding its line information to the converter.
+    ///
+    /// `call_locations` is a non-empty sorted list of `(address, call_location index)` pairs.
```
loewenheim: This line needs to be on `InProgressFunction`.

symbolic-symcache/src/writer.rs
```diff
@@ -456,6 +471,321 @@ impl<'a> SymCacheConverter<'a> {
...
+    fn process_symbolic_functions(
+        &mut self,
+        tr: &dyn TypeResolver,
+        mut function_stack: Vec<InProgressFunction<'_>>,
+    ) {
+        while let Some(in_progress_function) = function_stack.pop() {
+            let function = in_progress_function.function;
+            let base_idx = in_progress_function.base_index;
+            let fn_depth = in_progress_function.depth;
+            let call_locations = in_progress_function.call_locations;
```
loewenheim:
```suggestion
            let InProgressFunction {function, base_index: base_idx, depth: fn_depth, call_locations} = in_progress_function;
```
Or deconstruct it in the `while let` right away.

loewenheim (changes requested):
I've only reviewed the symcache part in detail so far.

Aside from some nits, I would suggest splitting this up—the symcache and DWARF parts are effectively totally disjoint and don't need to be in the same PR. Also, the change to the `max_inline_depth` option seems extraneous.

symbolic-debuginfo/src/dwarf/mod.rs
```diff
@@ -1215,52 +1227,114 @@ impl<'d, 'a> DwarfUnit<'d, 'a> {
...
-            let size = range.end - range.begin;
-            let variables = variables
-                .iter()
-                .filter_map(|variable| self.variable_for_range(variable, *range))
-                .collect();
+            // It's possible the top function is dead-code; if so, we want to ignore anything
+            // nested inside that is NOT a subprogram.
+            let deadcode_top = function_stack
+                .last()
+                .is_some_and(|p| matches!(p, InProgressSubProgram::Deadcode(_)));
```
loewenheim:
```suggestion
            let deadcode_top = matches!(
                function_stack.last(),
                Some(InProgressSubProgram::Deadcode(_))
            );
```

symbolic-debuginfo/src/dwarf/mod.rs
```diff
@@ -1215,52 +1227,114 @@ impl<'d, 'a> DwarfUnit<'d, 'a> {
...
+            let next_depth = entries.next_depth();
+
+            // Use next_depth to finish any in-progress functions that are higher on the stack.
+            while let Some(last_func) = function_stack.last() {
+                if last_func.depth() < next_depth {
+                    break;
+                }
+
+                let last_func: InProgressSubProgram<'_> =
+                    function_stack.pop().expect("already checked");
```
loewenheim:
```suggestion
                let last_func = function_stack.pop().expect("already checked");
```

loewenheim (approved):
Looks good now, aside from two small nits.

PR #1065, opened 2026-09-02
fix(debuginfo): Remove depth limits from function parsing

klochek:
We will now handle function parsing in both symcache convertor and dwarf DIE parsing iteratively, so there is no longer a need for bounding these by function depth.

This also required introducing an iterative fmt::Debug implementation for Function, authored by Claude.

### DIFF
--- a/symbolic-debuginfo/src/base/mod.rs
+++ b/symbolic-debuginfo/src/base/mod.rs
@@ -742,6 +742,19 @@ pub struct Function<'data> {
     pub variables: Vec<variable::Variable<'data>>,
 }
 
+// We need a custom drop implementation to avoid recursing down the inlinees.
+impl<'data> Drop for Function<'data> {
+    fn drop(&mut self) {
+        let mut inlinees = vec![std::mem::take(&mut self.inlinees)];
+
+        while let Some(inlinee_set) = inlinees.pop() {
+            for mut f in inlinee_set {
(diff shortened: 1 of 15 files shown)


symbolic-debuginfo/src/base/mod.rs
loewenheim, line 1: This raises a question—do we want the iterative `Drop/Debug` after all, now that we've taken away the parsing limits? The `Debug` impl at least is very low value compared to the complication it introduces, IMO.

symbolic-symcache/src/writer.rs
loewenheim, line 1: The symcache change should be its own PR.

loewenheim (changes requested):

symbolic-debuginfo/src/base/mod.rs
klochek: Without an iterative drop impl, symbolic can still stack-overflow on carefully-crafted dwarf files.  The fmt::Debug impl is a bit more of a wash--it's only really used in tests, but it's pretty helpful to have something that will always work.

loewenheim: The `Drop` makes sense, but I think the `Debug` is far more trouble than it's worth. I'd remove it.

loewenheim (approved):

PR #1068, opened 2026-09-17
fix(demangler): Add recursion limit to swift `Demangler`

tobias-wilfert:
Ref: INGEST-1195


symbolic-demangle/build.rs
```diff
@@ -1,6 +1,8 @@
 fn main() {
     #[cfg(feature = "swift")]
     {
+        println!("cargo:rerun-if-changed=vendor/swift/include");
```
tobias-wilfert, line 4: Drive-by: Added since changing the .h files in this directory didn't trigger a recompile for me.

symbolic-demangle/vendor/swift/lib/Demangling/Demangler.cpp
```diff
@@ -800,6 +800,7 @@ NodePointer Demangler::demangleSymbol(StringRef MangledName,
 NodePointer Demangler::demangleType(StringRef MangledName,
         std::function<SymbolicReferenceResolver_t> Resolver) {
   DemangleInitRAII state(*this, MangledName, std::move(Resolver));
+  RecursionGuard guard(*this);
```
tobias-wilfert, line 803: Added preemptively.

symbolic-demangle/vendor/swift/lib/Demangling/Demangler.cpp
loewenheim, line 1: Are these really the only two places where you have to guard the recursion?

loewenheim (approved):

symbolic-demangle/vendor/swift/include/swift/Demangling/Demangler.h
```diff
@@ -421,6 +421,19 @@ class Demangler : public NodeFactory {
...
+  /// Same value as MaxDepth in NodePrinter (set by upstream).
+  static const int MaxRecursionDepth = 768;
+  int RecursionDepth = 0;
+
+  struct RecursionGuard {
+    Demangler &Dem;
+    RecursionGuard(Demangler &Dem) : Dem(Dem) {
+      ++Dem.RecursionDepth;
+      assert(Dem.RecursionDepth <= MaxRecursionDepth);
+    }
```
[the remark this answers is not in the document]
tobias-wilfert: What makes you think it gets past the FFI? Could there perhaps be a little catch on the CPP side that we added specifically for this in the last commit to this logic?

PR #1072, opened 2026-09-21
fix(unreal): Don't accept unbounded custom nodes in RuntimeProperties

klochek:


tobias-wilfert:
I think adding a test, if possible, would be nice. Also left a comment on the internal ticket.

symbolic-unreal/src/context.rs
```diff
@@ -326,6 +334,10 @@ fn load_data_bag(
             let name = String::from_utf8_lossy(child.tag().name().as_ref()).to_string();
             let value = child.value::<String>()?.unwrap_or_default();
             dest_data.insert(name, value);
+
+            if dest_data.len() > MAX_DATA_BAG_NODES {
```
loewenheim: I think this check can be moved to the start of the loop.

symbolic-unreal/src/context.rs
```diff
@@ -8,6 +8,12 @@ use similar_asserts::assert_eq;
 
 use crate::{error::Unreal4Error, xml::XMLReader};
 
+// Bounds for custom/unknown child nodes within the RuntimeProperties tag.
```
loewenheim, line 11:
```suggestion
/// Bounds for custom/unknown child nodes within the RuntimeProperties tag.
```

symbolic-unreal/src/context.rs
```diff
@@ -8,6 +8,12 @@ use similar_asserts::assert_eq;
...
+const MAX_CUSTOM_NODES: usize = 100_000;
+
+// Bounds for all child nodes loaded with `load_databag`.
```
loewenheim, line 14:
```suggestion
/// Bounds for all child nodes loaded with `load_databag`.
```

loewenheim (approved):

tobias-wilfert (approved):

PR #1073, opened 2026-09-21
fix(demangle): Bound output string size

klochek:


tobias-wilfert (changes requested):
Looks good, would be nice to also add a patch file like the ones we already have (for example  `symbolic-demangle/vendor/swift/1-arguments.patch`) and a changelog message.

loewenheim (approved):

tobias-wilfert (approved):

PR #1075, opened 2026-09-22
fix(symcache): Do not undecrate DWARF function names from PE files

### LINKED ISSUE #1071: GCC/MinGW/Native backend: DW_AT_linkage_name values are mangled
### Environment
Platform: Windows, GCC/MinGW, DWARF debug info, minidump
Issue ID: https://technically-games-llc.sentry.io/issues/TRANSCRIBE-2/
### Steps to Reproduce
1. Build a C++ application with MinGW-w64 GCC (x86_64) as `RelWithDebInfo`, linking the
   executable with `-Wl,--build-id`
2. Split the debug info:
   `objcopy --only-keep-debug TranScribe.exe TranScribe.exe.debug && strip TranScribe.exe`
   `sentry-cli debug-files check` reports both files as `Usable: yes` with the same debug ID
3. Upload both with
   `sentry-cli debug-files upload --include-sources --wait <dir>`.
4. Initialize sentry-native with the `native` backend and `MINIDUMP` crash reporting mode, then
   crash deliberately (null pointer dereference in a constructor).
### Expected Result
Frames symbolicated and demangled, as `addr2line -f -C` does locally against the same
uploaded debug file:
```std::__atomic_base::load(std::memory_order) const atomic_base.h:505 spdlog::logger::should_log(spdlog::level::level_enum) const logger.h:263 spdlog::logger::log(...) logger.h:116 void spdlog::logger::log<char [20]>(...) logger.h:90 void spdlog::logger::info<char [20]>(...) logger.h:243 TSMainWindow::TSMainWindow(QWidget*) main_window.cpp:54 qMain(int, char**) main.cpp:88``` 

### Actual Result
File names, line numbers and inline frames are all correct, but every function name is rendered
as a raw Itanium mangled name with the leading underscore missing:
```ZNKSt13__atomic_baseIiE4loadESt12memory_order atomic_base.h:505 ZNK6spdlog6logger10should_logENS_5level10level_enumE logger.h:263 ZN6spdlog6logger3logENS_10source_locENS_5level10level_enumE... logger.h:116 ZN6spdlog6logger3logIA20_cEEvNS_5level10level_enumERKT_ logger.h:90 ZN6spdlog6logger4infoIA20_cEEvRKT_ logger.h:243 ZN12TSMainWindowC2EP7QWidget main_window.cpp:54 Z5qMainiPPc main.cpp:88```

### Additional Context
The names in the uploaded debug file do have the `_Z` prefix, so the prefix is being removed
somewhere in Sentry's processing. For the same address:
``$ addr2line -f -e Transcribe.exe.debug 0x140006b88 _ZN12TSMainWindowC2EP7QWidget D:/Development/TranScribe2/src/TranscribeGUI/private/main_window.cpp:54``

### PR DESCRIPTION
**Pull request by Dav1dde**
Adjacent to #1074. 

This is overall a more fundamental fix than #1074, it addresses all DWARF function names parsed from the PE embedded DWARF data.

```rs
    // Without the change in this PR, the function name would resolve as `_tmainCRTStartup` (leading underscore stripped).
    assert_eq!(symbols[0].function().name(), "__tmainCRTStartup");
```

Fixes: #1071


symbolic-symcache/src/writer.rs
```diff
@@ -127,7 +132,18 @@ impl<'a> SymCacheConverter<'a> {
         self.set_arch(object.arch());
         self.set_debug_id(object.debug_id());
 
-        self.is_windows_object = matches!(object.file_format(), FileFormat::Pe | FileFormat::Pdb);
+        let file_format = object.file_format();
+
+        // Symbols/Functions from PDB files are generally already formatted already, but they may
```
loewenheim:
```suggestion
        // Symbols/Functions from PDB files are generally formatted already, but they may
```

loewenheim (approved):
There's a typo in the title.

Dav1dde:
Going to remove the previous `_Z` workaround, turns out this is not a good heuristic, there are cases such as `_ZwGetContextThread@8` where this incorrectly skips the un-decoration.

